PR #5, opened 2020-03-27
Relax licence restrictions to support custom licences

abn:
This change adds support of use of custom licenses for projects. If the
value specified for `licence` is not in the SPDX License List or
"Proprietary", we treat it as a custom licence withe value (as is) as
the name.

As a side-effect, the validation of licence is no longer performed when
executing `poetry check`.

Resolves: python-poetry/poetry#2020


abn:
This was okayed on discored.

PR #6, opened 2020-03-27
Add support of inline tables in include on pyproject.toml.

stephsamson:
Resolves: #8 

- [x] tests
- [x] documentation [here](https://github.com/python-poetry/poetry/pull/2268)

This PR supports inline tables in the include directive. This is similarly functional to the `package` directive.

### DIFF
--- a/poetry/core/factory.py
+++ b/poetry/core/factory.py
@@ -115,7 +115,18 @@ class Factory(object):
             package.build_config = build or {}
 
         if "include" in local_config:
-            package.include = local_config["include"]
+            package.include = []
+
+            for include in local_config["include"]:
+                if not isinstance(include, dict):
+                    include = {"path": include}
+
+                formats = include.get("format", [])
+                if formats and not isinstance(formats, list):
+                    formats = [formats]
+                include["format"] = formats
+
+                package.include.append(include)
 
         if "exclude" in local_config:
             package.exclude = local_config["exclude"]
--- a/poetry/core/json/schemas/poetry-schema.json
+++ b/poetry/core/json/schemas/poetry-schema.json
@@ -73,26 +73,43 @@
                 ],
                 "properties": {
                     "include": {
-                        "type": "string",
-                        "description": "What to include in the package."
+                        "$ref": "#/definitions/include-path"
                     },
                     "from": {
                         "type": "string",
                         "description": "Where the source directory of the package resides."
                     },
                     "format": {
-                        "oneOf": [
-                            {"type": "string"},
-                            {"type":  "array", "items": {"type":  "string"}}
-                        ],
-                        "description": "The format(s) for which the package must be included."
+                        "$ref": "#/definitions/package-formats"
                     }
                 }
             }
         },
         "include": {
             "type": "array",
-            "description": "A list of files and folders to include."
+            "description": "A list of files and folders to include.",
+            "items": {
+                "anyOf": [
+                    {
+                        "$ref": "#/definitions/include-path"
+                    },
+                    {
+                        "type": "object",
+                        "additionalProperties": false,
+                        "required": [
+                            "path"
+                        ],
+                        "properties": {
+                            "path": {
+                                "$ref": "#/definitions/include-path"
+                            },
+                            "format": {
+                                "$ref": "#/definitions/package-formats"
+                            }
+                        }
+                    }
+                ]
+            }
         },
         "exclude": {
             "type": "array",
@@ -188,6 +205,22 @@
             "items": {
                 "type": "string"
             }
+        },
+        "include-path": {
+            "type": "string",
+            "description": "Path to file or directory to include."
+        },
+        "package-format": {
+            "type": "string",
+            "enum": ["sdist", "wheel"],
+            "description": "A Python packaging format."
+        },
+        "package-formats": {
+            "oneOf": [
+                {"$ref": "#/definitions/package-format"},
+                {"type": "array", "items": {"$ref": "#/definitions/package-format"}}
+            ],
+            "description": "The format(s) for which the package must be included."
         },
         "dependencies": {
             "type": "object",
--- a/poetry/core/masonry/builders/builder.py
+++ b/poetry/core/masonry/builders/builder.py
@@ -6,6 +6,7 @@ import tempfile
 
 from collections import defaultdict
 from contextlib import contextmanager
+from typing import Optional
 from typing import Set
 from typing import Union
 
@@ -60,12 +61,27 @@ class Builder(object):
 
             packages.append(p)
 
+        includes = []
+        for include in self._package.include:
+            formats = include.get("format", [])
+
+            if (
+                formats
+                and self.format
+                and self.format not in formats
+                and not ignore_packages_formats
+            ):
+                continue
+
+            includes.append(include)
+
         self._module = Module(
             self._package.name,
             self._path.as_posix(),
             packages=packages,
-            includes=self._package.include,
+            includes=includes,
         )
+
         self._meta = Metadata.from_package(self._package)
 
     def build(self):
@@ -113,23 +129,49 @@ class Builder(object):
 
         return False
 
-    def find_files_to_add(self, exclude_build=True):  # type: (bool) -> list
+    def find_files_to_add(
+        self, exclude_build=True
+    ):  # type: (bool) -> Set[BuildIncludeFile]
         """
         Finds all files to add to the tarball
         """
-        to_add = []
+        to_add = set()
 
         for include in self._module.includes:
+            include.refresh()
+            formats = include.formats or ["sdist"]
+
             for file in include.elements:
                 if "__pycache__" in str(file):
                     continue
 
                 if file.is_dir():
+                    if self.format in formats:
+                        for current_file in file.glob("**/*"):
+                            include_file = BuildIncludeFile(
+                                path=current_file, source_root=self._path
+                            )
+
+                            if not current_file.is_dir() and not self.is_excluded(
+                                include_file.relative_to_source_root()
+                            ):
+                                to_add.add(include_file)
                     continue
 
-                file = file.relative_to(self._path)
+                if (
+                    isinstance(include, PackageInclude)
+                    and include.source
+                    and self.format == "wheel"
+                ):
+                    source_root = include.base
+                else:
+                    source_root = self._path
 
-                if self.is_excluded(file) and isinstance(include, PackageInclude):
+                include_file = BuildIncludeFile(path=file, source_root=source_root)
+
+                if self.is_excluded(
+                    include_file.relative_to_source_root()
+                ) and isinstance(include, PackageInclude):
                     continue
 
                 if file.suffix == ".pyc":
@@ -140,31 +182,19 @@ class Builder(object):
                     continue
 
                 logger.debug(" - Adding: {}".format(str(file)))
-                to_add.append(file)
+                to_add.add(include_file)
 
-        # Include project files
-        logger.debug(" - Adding: pyproject.toml")
-        to_add.append(Path("pyproject.toml"))
-
-        # If a license file exists, add it
-        for license_file in self._path.glob("LICENSE*"):
-            logger.debug(" - Adding: {}".format(license_file.relative_to(self._path)))
-            to_add.append(license_file.relative_to(self._path))
-
-        # If a README is specified we need to include it
-        # to avoid errors
-        if "readme" in self._poetry.local_config:
-            readme = self._path / self._poetry.local_config["readme"]
-            if readme.exists():
-                logger.debug(" - Adding: {}".format(readme.relative_to(self._path)))
-                to_add.append(readme.relative_to(self._path))
-
-        # If a build script is specified and explicitely required
+        # If a build script is specified and explicitly required
         # we add it to the list of files
         if self._package.build_script and not exclude_build:
-            to_add.append(Path(self._package.build_script))
+            to_add.add(
+                BuildIncludeFile(
+                    path=self._path / self._package.build_script,
+                    source_root=self._path,
+                )
+            )
 
-        return sorted(to_add)
+        return to_add
 
     def get_metadata_content(self):  # type: () -> bytes
         content = METADATA_BASE.format(
@@ -268,3 +298,40 @@ class Builder(object):
             yield name
 
             shutil.rmtree(name)
+
+
+class BuildIncludeFile:
+    def __init__(
+        self,
+        path,  # type: Path
+        source_root=None,  # type: Optional[Path]
+    ):
+        """
+        :param path: a full path to the file to be included
+        :param source_root: the root path to resolve to
+        """
+        self.path = Path(path)
+        self.source_root = None if not source_root else Path(source_root).resolve()
+        if not self.path.is_absolute() and self.source_root:
+            self.path = (self.source_root / self.path).resolve()
+        else:
+            self.path = self.path.resolve()
+
+    def __eq__(self, other):  # type: (Union[BuildIncludeFile, Path]) -> bool
+        if hasattr(other, "path"):
+            return self.path == other.path
+        return self.path == other
+
+    def __ne__(self, other):  # type: (Union[BuildIncludeFile, Path]) -> bool
+        return not self.__eq__(other)
+
+    def __hash__(self):
+        return hash(self.path)
+
+    def __repr__(self):  # type: () -> str
+        return str(self.path)
+
+    def relative_to_source_root(self):  # type(): -> Path
+        if self.source_root is not None:
+            return self.path.relative_to(self.source_root)
+        return self.path
--- a/poetry/core/masonry/builders/builder.py
+++ b/poetry/core/masonry/builders/builder.py
@@ -184,13 +184,11 @@ class Builder(object):
                 logger.debug(" - Adding: {}".format(str(file)))
                 to_add.add(include_file)
 
-        # If a build script is specified and explicitly required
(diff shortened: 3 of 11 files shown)


poetry/core/masonry/builders/builder.py
```diff
@@ -61,11 +61,28 @@ def __init__(
 
             packages.append(p)
 
+        includes = []
+        for include in self._package.include:
+            if isinstance(include, dict):
```
abn: Should we consider moving this check into the factory implementation? So we can normalise access to includes? Same with formats.

poetry/core/masonry/utils/module.py
```diff
@@ -74,7 +74,16 @@ def __init__(self, name, directory=".", packages=None, includes=None):
             )
 
         for include in includes:
-            self._includes.append(Include(self._path, include))
+            if isinstance(include, dict):
```
abn: Comment above applies here to.

tests/masonry/builders/test_complete.py
```diff
@@ -387,22 +387,22 @@ def test_package_src():
 
 
 def test_package_with_include(mocker):
-    module_path = fixtures_dir / "with-include"
+    module_path = fixtures_dir / "with_include"
```
abn: Is this rename required? Personally, I like that these use `-` since they are not package directories.

poetry/core/masonry/builders/builder.py
stephsamson: Yes, I think that's a good idea.

tests/masonry/builders/test_complete.py
stephsamson: Mm, you're right. I also prefer `-` but the rest of the tests use `_` so I just wanted to maintain consistency.

abn: We can tackle this change in a separate PR if it is not required for the fix in this PR.

abn:
@stephsamson this should also resolve python-poetry/poetry#2087 correct?

theunkn0wn1:
As per our discussion in discord,
The feature I would like to see out of this is to be able to embed resource files into the wheel (and subsequently the installed package) at an arbitrary location.

For instance, in my project i would like to embed the `pyproject.toml` into the installed sources so the library may access this resource at runtime. I wish to do this to access the version string contained in the `pyproject.toml`. Presently, even if the `tool.poetry.include` directive is used the file does not end up in the installation directory, thus is unusable to the installed application.
My application has the structure
```
/
/pyproject.toml
/src/...
/test/...
```
What I would like to do is copy, at build time, `/pyproject.toml` to `/src/pyproject.toml` so it may be embedded as a resource.

stephsamson:
> @stephsamson this should also resolve [python-poetry/poetry#2087](https://github.com/python-poetry/poetry/issues/2087) correct?

Yes!

poetry/core/json/schemas/poetry-schema.json
```diff
@@ -92,7 +92,42 @@
         },
         "include": {
             "type": "array",
-            "description": "A list of files and folders to include."
+            "description": "A list of files and folders to include.",
+            "items": {
+                "anyOf": [
+                    {
+                        "type": "string"
```
abn, line 99:
```suggestion
                        "type": "string",
                        "description": "Path to file or directory to include."
```

poetry/core/json/schemas/poetry-schema.json
```diff
@@ -92,7 +92,42 @@
...
+                    },
+                    {
+                        "type": "object",
+                        "description": "Information about where the file or folder resides.",
```
abn, line 103:
```suggestion
```

poetry/core/json/schemas/poetry-schema.json
```diff
@@ -92,7 +92,42 @@
...
+                        "additionalProperties": false,
+                        "required": [
+                            "include"
+                        ],
+                        "properties": {
+                            "include": {
+                                "type": "string",
+                                "description": "What to include in the package."
```
abn, line 111:
```suggestion
                                "description": "Path to file or directory to include"
```

poetry/core/json/schemas/poetry-schema.json
```diff
@@ -92,7 +92,42 @@
...
+                            },
+                            "format": {
```
abn, line 113: we should probably move this to defintions section with as something like `package-format` and restrict values to what we support - `sdist`, `wheel` etc. Might need to check of 'bdist' is supported as well. Also, might be good to have descriptions for strings.

poetry/core/masonry/builders/builder.py
```diff
@@ -127,6 +141,15 @@ def find_files_to_add(self, exclude_build=True):  # type: (bool) -> list
                     continue
 
                 if file.is_dir():
+                    if self.format == "sdist" or self.format in include.formats:
+                        for f in Path(file).glob("**/*"):
```
abn, line 145:
```suggestion
                        for f in file.glob("**/*"):
```

`file` is already a `Path` object here.

poetry/core/masonry/builders/builder.py
```diff
@@ -127,6 +141,15 @@ def find_files_to_add(self, exclude_build=True):  # type: (bool) -> list
...
+                            included = f.relative_to(self._path)
+                            if (
+                                included not in to_add
```
abn, line 148: Probably should make `to_add` a set here

poetry/core/masonry/builders/builder.py
```diff
@@ -127,6 +141,15 @@ def find_files_to_add(self, exclude_build=True):  # type: (bool) -> list
                     continue
 
                 if file.is_dir():
+                    if self.format == "sdist" or self.format in include.formats:
```
abn, line 144: How does this impact cases where include is specified only for wheel? We might need to add more test coverage for edge cases here.

poetry/core/json/schemas/poetry-schema.json
```diff
@@ -92,7 +92,42 @@
...
+                        "description": "Path to file or directory to include."
+                    },
+                    {
+                        "type": "object",
+                        "additionalProperties": false,
+                        "required": [
+                            "include"
+                        ],
+                        "properties": {
+                            "include": {
```
abn, line 109:
```suggestion
                        "required": [
                            "path"
                        ],
                        "properties": {
                            "path": {
```

poetry/core/masonry/builders/builder.py
stephsamson: The `self.format == "sdist"` check exists because of the cases when inline tables are not used. For example:

```toml
include = ["some.txt", "foo", "bar"]
```

Because they are not specified with a format, it is assumed that they are to be included for sdists.

Regarding specifying wheels only, you are right. Wheels are not affected by this change since the adding mechanism is elsewhere (in `Wheel._copy_module`) which means that if only wheels are specified in `include`, any dir includes are excluded.

poetry/core/json/schemas/poetry-schema.json
stephsamson: Not sure if `bdist` is necessary since it's just used for [extracting metadata](https://github.com/python-poetry/core/blob/master/poetry/core/_vendor/pkginfo/tests/test_bdist.py). Or do you mean that we should treat `wheel == bdist`?

poetry/core/masonry/builders/builder.py
```diff
@@ -115,58 +131,70 @@ def is_excluded(self, filepath):  # type: (Union[str, Path]) -> bool
...
-
-        # If a build script is specified and explicitely required
+                to_add.append(include_file)
+
+        # If a build script is specified and explicitly required
         # we add it to the list of files
         if self._package.build and not exclude_build:
-            to_add.append(Path(self._package.build))
+            to_add.append(
+                IncludeFile(path=Path(self._package.build), source_root=self._path)
```
abn, line 194: We can move, the `Path()`. `.relative_to()` and `.resolve()` logic into the class itself I think.

poetry/core/masonry/builders/sdist.py
```diff
@@ -301,6 +302,38 @@ def find_nearest_pkg(rel_path):
...
+        for license_file in self._path.glob("LICENSE*"):
+            logger.debug(" - Adding: {}".format(license_file.relative_to(self._path)))
+            to_add.append(
+                IncludeFile(
+                    path=license_file.relative_to(self._path), source_root=self._path
+                )
+            )
+
+        # If a README is specified we need to include it
+        # to avoid errors
```
abn, line 324:
```suggestion
        # If a README is specified we need to include it to avoid errors
```

poetry/core/masonry/utils/include.py
```diff
@@ -46,3 +46,24 @@ def refresh(self):  # type: () -> Include
         self._elements = sorted(list(self._base.glob(self._include)))
 
         return self
+
+
+class IncludeFile:
```
abn, line 51: This hould probably be in builder module, and could be renamed to something like `BuildIncludeFile` to be more descriptive of purpose.

poetry/core/masonry/utils/include.py
```diff
@@ -46,3 +46,24 @@ def refresh(self):  # type: () -> Include
...
+    def __init__(
+        self,
+        path,  # type: Path
+        source_root=None,  # type: Optional[Path]
+    ):
+        self.path = path
+        self.source_root = source_root or Path(".")
```
abn, line 58: This might cause some expected issues. Better to treat `None` explicitly.

abn, line 58:
```suggestion
        self.parent = source_root or Path(".")
```

poetry/core/masonry/utils/include.py
```diff
@@ -46,3 +46,24 @@ def refresh(self):  # type: () -> Include
...
+    def __repr__(self):  # type: () -> str
+        return str(self.path)
+
+    @property
+    def rel_path(self):  # type: () -> Path
+        return self.path
+
+    @property
+    def full_path(self):  # type: () -> Path
+        return self.source_root / self.path
```
abn, line 69:
```suggestion
    def relative_path(self):  # type: () -> Path
        if self.parent:
            return self.path.resolve().relative_to(self.parent.resovle())
        return self.path

    def resolve(self):  # type: () -> Path
        if self.parent:
            return (self.parent / self.path).resolve()
        return self.path.resolve()
```

Something like this might be a cleaner approach.

poetry/core/factory.py
```diff
@@ -112,7 +112,18 @@ def create_poetry(self, cwd=None):  # type: (Optional[Path]) -> Poetry
...
 
         if "include" in local_config:
-            package.include = local_config["include"]
+            package.include = []
+
+            for include in local_config["include"]:
+                if not isinstance(include, dict):
+                    include = {"path": include}
+
+                formats = include.get("format", [])
```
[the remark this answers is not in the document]
stephsamson: I believe the rationale for having it called `format` is because it can also be a string representing just one format.

poetry/core/masonry/builders/builder.py
```diff
@@ -115,58 +131,67 @@ def is_excluded(self, filepath):  # type: (Union[str, Path]) -> bool
...
+        self, exclude_build=True
+    ):  # type: (bool) -> List[IncludeFile]
         """
         Finds all files to add to the tarball
         """
         to_add = []
 
         for include in self._module.includes:
+            include.refresh()
+            formats = include.formats or ["sdist"]
```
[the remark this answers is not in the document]
stephsamson: Yes, this is the same check I was doing here basically: https://github.com/python-poetry/core/pull/6#discussion_r407199243

poetry/core/masonry/utils/include.py
stephsamson: I'm not sure I understand the example for `relative_path`.

`self.path`, which is usually the relative path, when resolved relative to its parent because it resolves itself to whatever the current directory is (in the example below, it is my `poetry-core` directory: `/Users/stephsamson/Code/opensource/poetry-core/`) instead of the actual parent directory (`'/Users/stephsamson/Code/opensource/poetry-core/tests/masonry/builders/fixtures/with_include_inline_table'`).

example:
```
ValueError:  '/Users/stephsamson/Code/opensource/poetry-core/LICENSE' does not start with '/Users/stephsamson/Code/opensource/poetry-core/tests/masonry/builders/fixtures/with-include'
```

poetry/core/masonry/builders/builder.py
```diff
@@ -37,7 +38,7 @@ class Builder(object):
 
     def __init__(
         self, poetry, ignore_packages_formats=False
-    ):  # type: ("Poetry", bool) -> None
+    ):  # type: ("Poetry", "Env", "IO", bool) -> None
```
abn, line 41: Looks like a regression during merge.

poetry/core/masonry/builders/builder.py
```diff
@@ -113,23 +129,53 @@ def is_excluded(self, filepath):  # type: (Union[str, Path]) -> bool
...
 
                 if file.is_dir():
+                    if self.format in formats:
+                        for current_file in file.glob("**/*"):
+                            include_file = BuildIncludeFile(
+                                path=current_file, source_root=self._path
+                            )
+
+                            if (
+                                include_file.relative_to_source_root() not in to_add
```
abn, line 156: Shouldn't this be
```suggestion
                                include_file not in to_add
```

Also, since this is a set we can skip this check.

poetry/core/masonry/builders/builder.py
```diff
@@ -268,3 +302,38 @@ def temporary_directory(cls, *args, **kwargs):
...
+    def __init__(
+        self,
+        path,  # type: Path
+        source_root=None,  # type: Optional[Path]
+    ):
+        """
+        :param path: a full path to the file to be included
+        :param source_root: the root path to resolve to
+        """
+        self.path = Path(path).resolve()
```
abn, line 317: Should resolve be done after we check for source root? Since if source root was `/tmp/build` and path is `./somefile.txt`, this would resolve to `/home/user/source/somefile.txt` if current directory is `/home/user/source`.

poetry/core/masonry/builders/builder.py
```diff
@@ -268,3 +302,38 @@ def temporary_directory(cls, *args, **kwargs):
...
+        self.source_root = None if not source_root else Path(source_root).resolve()
+        if not self.path.is_absolute() and self.source_root:
+            self.path = (self.source_root / self.path).resolve()
+
+    def __eq__(self, other):  # type: (Union[BuildIncludeFile, Path]) -> bool
+        if hasattr(other, "path"):
+            return self.path == other.path
+        return self.path == other
+
+    def __ne__(self, other):  # type: (Union[BuildIncludeFile, Path]) -> bool
```
abn, line 327: Does this need to be explicitly defined?

poetry/core/masonry/builders/builder.py
```diff
@@ -140,31 +186,19 @@ def find_files_to_add(self, exclude_build=True):  # type: (bool) -> list
...
-        # If a build script is specified and explicitely required
+                to_add.add(include_file)
+
+        # If a build script is specified and explicitly required
         # we add it to the list of files
         if self._package.build_script and not exclude_build:
-            to_add.append(Path(self._package.build_script))
+            to_add.add(
+                BuildIncludeFile(
+                    path=self._path / Path(self._package.build_script),
```
abn, line 196:
```suggestion
                    path=self._path / self._package.build_script,
```

poetry/core/masonry/builders/sdist.py
```diff
@@ -302,6 +303,37 @@ def find_nearest_pkg(rel_path):
...
+    def find_files_to_add(
+        self, exclude_build=False
+    ):  # type: (bool) -> List[BuildIncludeFile]
+        to_add = super(SdistBuilder, self).find_files_to_add(exclude_build)
+
+        # Include project files
+        logger.debug(" - Adding: pyproject.toml")
+        to_add.add(
+            BuildIncludeFile(
+                path=self._path / Path("pyproject.toml"), source_root=self._path
```
abn, line 315:
```suggestion
                path=self._path / "pyproject.toml", source_root=self._path
```

poetry/core/masonry/builders/sdist.py
```diff
@@ -302,6 +303,37 @@ def find_nearest_pkg(rel_path):
...
+            )
+        )
+
+        # If a license file exists, add it
```
abn, line 319: Shouldn't we adde the following in all builds?

poetry/core/masonry/builders/builder.py
stephsamson: Yes, otherwise the hashing function for the set does not work properly: there can be cases where a `Path` is checked if it is in a set of `BuildIncludeFile` objects.

poetry/core/masonry/builders/sdist.py
stephsamson: Since built distributions are made with source distributions anyway, all of these files will be included. If I add it to all builds, then we will have the `LICENSE` files twice and `README` files twice: once from the `sdist` and once again from the second time `find_files_to_add` will be called from `WheelBuilder._copy_module`.

tests/masonry/builders/test_complete.py
```diff
@@ -266,7 +266,7 @@ def test_complete_no_vcs():
         "my_package-1.2.3.dist-info/METADATA",
         "my_package-1.2.3.dist-info/RECORD",
     ]
-
+    # import pdb; pdb.set_trace()
```
abn, line 269:
```suggestion

```

PR #12, opened 2020-03-31
 Fix PEP-517 issues for projects using build scripts

abn:
In order to do this, we introduce a context manager in sdist builder. This can also be reused for the editable builder in the poetry code base.

Resolves: python-poetry/poetry#1516

Removed tests inclusion in sdist due to python-poetry/poetry#2283. This was breaking CI with preview poetry release.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -43,7 +43,7 @@ jobs:
       uses: actions/cache@v1
       with:
         path: .venv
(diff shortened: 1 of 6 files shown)


abn:
Had to remove test source in sdist as this was causing issues due to files generated during run-time.

aviramha:
Did it release in any version?

PR #17, opened 2020-04-08
fix subpackage check for setup.py generation

abn:
This is a forward port of python-poetry/poetry#1626.

Resolves: python-poetry/poetry#2283


abn:
Unrelated failures in CI.

poetry/core/masonry/builders/sdist.py
```diff
@@ -240,7 +240,15 @@ def find_nearest_pkg(rel_path):
             if from_top_level == ".":
                 continue
 
-            is_subpkg = any([filename.endswith(".py") for filename in filenames])
+            is_subpkg = any(
+                [filename.endswith(".py") for filename in filenames]
+            ) and not all(
+                [
```
stephsamson, line 246: Could this be extracted out to a different variable to make it easier to read? It took me a while to understand that this was checking that all of the Python files are not part of the excluded files.

abn: I agree; but I wanted to keep this at par with the fix we have already merged in poetry pr. That might be a seperate change.

stephsamson: Okay, makes sense. Then this PR looks good to go.

PR #19, opened 2020-04-09
Add publish-default setting to tools.poetry section

utek:
This PR adds `publish-default` setting to `[tool.poetry]` section, that allows override of the default behavior of `poetry publish`. Setting it to repository name causes default publish to said repository instead of pypi.org.

Required by: https://github.com/python-poetry/poetry/pull/2287

I'm not sure how to use those changes in the PR on poetry side. If anybody could point me to documentation it would be great.


abn:
We should be able to revisit this if we implement publish configuration settings as discussed in #27

PR #20, opened 2020-04-09
github: clean up default test workflow

abn:
This is a cleanup of the current github actions workflows to remove duplicate code.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -16,114 +16,64 @@ jobs:
         run: |
           pip install pre-commit
           pre-commit run --all-files
-  Linux:
+
+  Tests:
     needs: Linting
-    runs-on: ubuntu-latest
+    name: ${{ matrix.os }} / ${{ matrix.python-version }}
+    runs-on: ${{ matrix.os }}-latest
     strategy:
       matrix:
+        os: [Ubuntu, MacOS, Windows]
         python-version: [2.7, 3.5, 3.6, 3.7, 3.8, pypy2, pypy3]
-
+        exclude:
+          - os: MacOS
+            python-version: pypy2
+          - os: MacOS
+            python-version: pypy3
+          - os: Windows
+            python-version: pypy2
+          - os: Windows
+            python-version: pypy3
     steps:
-    - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -16,114 +16,44 @@ jobs:
...
-        poetry install
-    - name: Test
-      run: |
-        $env:Path += ";$env:Userprofile\.poetry\bin"
-        poetry run pytest -q tests
+      - name: Set up cache
+        uses: actions/cache@v1
+        with:
+          path: .venv
+          key: venv-${{ runner.os }}-py${{ matrix.python-version }}-${{ hashFiles('**/poetry.lock') }}
```
sdispater, line 53: We still need the full python version here since if it changes the virtualenv will be invalid and will need to be recreated.

.github/workflows/tests.yml
```diff
@@ -16,114 +16,44 @@ jobs:
...
-        poetry run pytest -q tests
-  Windows:
-    needs: Linting
-    runs-on: windows-latest
-    strategy:
-      matrix:
-        python-version: [2.7, 3.5, 3.6, 3.7, 3.8]
+      - name: Install and configure poetry
+        run: |
+          pip install --pre poetry
```
sdispater, line 46: I prefer the usage of the custom installer since it's the official recommended way to install Poetry, so if there is anything wrong with we will see it.

abn: @sdispater I agree we should test it, but I feel that it should be tested in the poetry project using a separate workflow on develop or master. And since `pip install` is also expected to work, I think this is also valuable. I'd like to keep the workflow itself trivial. This also avoids any path related issues, dealing with them reliably across OSes can be fragile. Additionally, this also avoids us having to download the unnecessary python version dependencies. But these are minor advantages.

If you feel strongly about retaining the use of `get-poetry` I can add that in.

abn:
@sdispater I think we can discuss the use of `pip` vs `get-poetry` outside of this, I have maintained the used of the installer for now. Also added a cache check step to prevent any issues with corrupt caches.

abn:
The same changes have been also added for poetry.

See: https://github.com/python-poetry/poetry/pull/2289

PR #21, opened 2020-04-12
Add support for markers inverse

sdispater:
This PR adds support for markers inverse. While this is not a direct improvement to `poetry-core`, it will help fix Poetry issues like this one: https://github.com/python-poetry/poetry/issues/1238

### DIFF
--- a/poetry/core/version/markers.py
+++ b/poetry/core/version/markers.py
@@ -419,6 +419,28 @@ class SingleMarker(BaseMarker):
             operator = "not in"
         elif self._operator == "not in":
             operator = "in"
+        elif self._operator == "~=":
+            # This one is more tricky to handle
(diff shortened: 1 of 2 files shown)


poetry/core/version/markers.py
```diff
@@ -157,6 +157,10 @@ def _format_marker(marker, first=True):
 
 
 class BaseMarker(object):
+    @property
+    def inverse(self):  # type: () -> BaseMarker
```
abn, line 161: @sdispater since this is returning a new object by performing an inversion should we a method instead of a property? Non-blocker, but might be a good idea.

sdispater: I changed it to a `invert()` method.

abn:
:rocket:

PR #22, opened 2020-04-12
Improve support for PEP-440 direct references

abn:
With this change we allow for PEP-508 string for file dependencies
to use PEP-440 direct reference using the file URI scheme (RFC-8089).

Note that this implementation will not allow non RFC-8089 file path
references. In order to allow for sdist to be portable in sane use
cases, we ensure that relative path dependencies do not use the
file URI scheme, but preserve path if relative or directories.

In addition to file resource, directory dependencies now support
the same scheme.

References:
- https://www.python.org/dev/peps/pep-0508/#backwards-compatibility
- https://www.python.org/dev/peps/pep-0440/#direct-references
- https://tools.ietf.org/html/rfc8089
- https://discuss.python.org/t/what-is-the-correct-interpretation-of-path-based-pep-508-uri-reference/2815/11

### DIFF
--- a/poetry/core/packages/utils/utils.py
+++ b/poetry/core/packages/utils/utils.py
@@ -1,6 +1,13 @@
 import os
 import posixpath
 import re
+import sys
+
+from typing import Union
+
+from poetry.core._vendor.six.moves.urllib.parse import unquote  # noqa
+from poetry.core._vendor.six.moves.urllib.parse import urlsplit  # noqa
+from poetry.core._vendor.six.moves.urllib.request import url2pathname  # noqa
 
 from poetry.core.packages.constraints.constraint import Constraint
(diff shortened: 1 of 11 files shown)


poetry/core/packages/utils/utils.py
```diff
@@ -1,6 +1,13 @@
 import os
 import posixpath
 import re
+import sys
+
+from typing import Union
+
+from poetry.core._vendor.six.moves.urllib.parse import unquote  # noqa
```
stephsamson, line 8: Just out of curiosity, why is there a `#noqa` flag here? Is one of the linters acting funny?

abn: Because this uses `._vendor`, linters do not like this.

PR #23, opened 2020-04-14
Allow the develop property for git dependencies

sdispater:
Resolves: https://github.com/python-poetry/poetry/issues/1816

This PR is just a part of a fix made in this PR: https://github.com/python-poetry/poetry/pull/1986

The `test_factory.py` file has been added because I needed it as part of the tests for this PR and I forgot to transfer it when splitting poetry-core from the main repository.


poetry/core/packages/package.py
```diff
@@ -299,6 +299,7 @@ def add_dependency(
                     rev=constraint.get("rev", None),
                     category=category,
                     optional=optional,
+                    develop=constraint.get("develop", True),
```
abn, line 302: Should this be `True` by default here? Not sure what the general expectation is.

sdispater: I wanted to set it to `False` by default but since the current behavior is similar to `True`, I don't want to break things if people rely on it, even though it's not documented anywhere and, as such, they should not rely on it.

PR #24, opened 2020-04-14
builder.sdist: remove setuptools import

abn:
It appears we do not really need setuptools for now. The use of setuptools was preventing pip from installing poetry correctly for PEP-517 builds when `--no-binary :all:` was being used. Additionally, now we can install poetry-core from sdist as expected. 

Resolves: python-poetry/poetry#454

### DIFF
--- a/poetry/core/masonry/builders/sdist.py
+++ b/poetry/core/masonry/builders/sdist.py
@@ -26,7 +26,7 @@ from .builder import Builder
 
 SETUP = """\
 # -*- coding: utf-8 -*-
-from setuptools import setup
+from distutils.core import setup
 
 {before}
 setup_kwargs = {{
--- a/tox.ini
+++ b/tox.ini
@@ -1,5 +1,6 @@
 [tox]
 skipsdist = true
+isolated_build = True
 envlist = py27, py35, py36, py37, py38, pypy, pypy3
 
 [testenv]


abn:
@sdispater was there a reason why we needed `setuptools`? I can see you made this change a while ago.

https://github.com/python-poetry/poetry/commit/f759876231ca6c0910ad823f0ba405e21b3f1c4b

sdispater:
@abn from what I recall, distutils is deprecated and the `setup()` of situtils does not support some keyword arguments that `setuptools` supports (like `extras_require`, for instance)

abn:
This might mean we need to specify setuptools as a dependency, since otherwise pep517 builds are failing unless the `build-system.requires` explicitly contains setuptools. At least until we can remove the need for the setup file. This might, however, mean we will end up re-implementing some logic from setuptools.

abn:
It looks like we might need to handle the issue some other way. The issue can only be reproduced if pip is called with `--no-binary :all:` for either `poetry-core` inteself or another package that uses it as the backend.

```console
    Using cached https://files.pythonhosted.org/packages/b7/70/5c287bd1e7c8c86a318755514e02c6d431335de541c43eef65d0fb5af955/poetry-core-1.0.0a5.tar.gz
    Installing build dependencies: started
    Installing build dependencies: finished with status 'done'
    Getting requirements to build wheel: started
    Getting requirements to build wheel: finished with status 'done'
      Preparing wheel metadata: started
      Preparing wheel metadata: finished with status 'done'
  Skipping bdist_wheel for poetry-core, due to binaries being disabled for it.
  Installing collected packages: poetry-core
    Running setup.py install for poetry-core: started
      Running setup.py install for poetry-core: finished with status 'error'
      ERROR: Complete output from command /tmp/.venv/bin/python -u -c 'import setuptools, tokenize;__file__='"'"'/tmp/pip-install-uupqj3gr/poetry-core/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' install --record /tmp/pip-record-5zbpsbn6/install-record.txt --single-version-externally-managed --prefix /tmp/pip-build-env-2ulvu7lx/overlay --compile --install-headers /tmp/.venv/include/site/python3.7/poetry-core:
      ERROR: Traceback (most recent call last):
        File "<string>", line 1, in <module>
      ModuleNotFoundError: No module named 'setuptools'
      ----------------------------------------
  ERROR: Command "/tmp/.venv/bin/python -u -c 'import setuptools, tokenize;__file__='"'"'/tmp/pip-install-uupqj3gr/poetry-core/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' install --record /tmp/pip-record-5zbpsbn6/install-record.txt --single-version-externally-managed --prefix /tmp/pip-build-env-2ulvu7lx/overlay --compile --install-headers /tmp/.venv/include/site/python3.7/poetry-core" failed with error code 1 in /tmp/pip-install-uupqj3gr/poetry-core/
```

abn:
Closing this one for now, since the issue seems to be only present for an older version of pip in the environment.

gweis:
> @abn from what I recall, distutils is deprecated and the `setup()` of situtils does not support some keyword arguments that `setuptools` supports (like `extras_require`, for instance)

Sorry for the hijack.
I am curious to find out where distils has been deprecated. https://docs.python.org/3/distutils/apiref.html# ... Seems to me it's still a big part of python standard lib. even setuptools is recommended.

abn:
@gweis i do not think `distutils` is deprecated. Direct use is simply discouraged. 

> The setuptools project adds new capabilities to the setup function and other APIs, makes the API consistent across different Python versions, and is hence recommended over using distutils directly.

Issue #14, opened 2020-04-03
Consider *not* to vendor dependencies


decathorpe:
Python projects are starting to rely on pyproject.toml / poetry based build systems now, and this is also affecting fedora packages, for example. In this context, it's important to have a working, "native" poetry package.

However, using vendored dependencies - like in this project (which looks like will be required by what will be poetry 1.1.0) - will make it almost impossible to package poetry, let alone package it correctly.

Please reconsider the decision to use bundled dependencies.

abn:
Embedded/Bundled dependencies have long been a problem (JARs anyone?). Unfortunately, packaging and distributing these kinds of projects are still not straightforward (this includes golang and java packages too).

To your specific concern, [doesn't pip do the same thing?](https://github.com/pypa/pip/tree/master/src/pip/_vendor) In cases like poetry and pip vendoring base dependencies solves a lot of usability issues to a large portion of the end-users. So, I am not sure if it makes sense to not vendor.

In the specific context of fedora packages, I'd think poetry would be handled similar to what is done in [python-pip.spec](https://src.fedoraproject.org/rpms/python-pip/blob/master/f/python-pip.spec#_130).

decathorpe:
Well, pip is really a special case in fedora, because - if I understand correctly - it's handled this way to solve bootstrap issues when upgrading to newer python versions, that doesn't apply to poetry.

And I really want to avoid to have to de-bundle things from poetry(-core).

encukou:
The issue with bundled stuff is that if one of those dependencies fixes a security vulnerability, a LTS distro needs to fix it everywhere – potentially in old versions of poetry, long after poetry has moved on and introduced incompatible changes.

`pip` is quite special because it's needed by Python itself. It's an example of something that's hard to maintain in a distro, not of something every build backend should do.

leycec:
**Gentoo packager here.** I'm packaging `poetry >= 1.1.0a1` and `poetry-core >= 1.0.0a5` for our humble source-based distro and can only concur with everything and everyone above – excluding [@abn in his apologetic "But `pip` does this bad thing, so `poetry` should too!" screed](https://github.com/python-poetry/core/issues/14#issuecomment-608439910). Appeal to authority is *no* justification for continued bad behaviour. `pip` is irrelevant; this is `poetry`.

[...]

PR #25, opened 2020-04-16
Cleanup vendored dependencies

abn:
This pull request contains 3 relevant changes to package vendoring.

1. Remove unused vendored dependencies.
2. Debundle compatibility only dependencies. These are only required in smaller scenarios and will be dropped by poetry 1.3 releases anyway.
3. Prototype sys.path injection I am not sure how reliable this is, feedback would be nice. Considering poetry use case in isolated, I do not expect this to have any issues. This would also mean that distribution package maintainers can debundle extremely easily by dropping the _vendor direcotry and removing the injection. We could go further by testing if the directory exists before pre-pending.

Relates-to: #14


poetry/core/__init__.py
```diff
@@ -1 +1,15 @@
...
+except ImportError:
+    # noinspection PyUnresolvedReferences
+    from pathlib2 import Path
+
 __version__ = "1.0.0a5"
+
+__vendor_site__ = (Path(__file__).parent / "_vendor").as_posix()
+
+if __vendor_site__ not in sys.path:
+    sys.path.insert(0, __vendor_site__)
```
sdispater, line 15: While I agree it would be better in theory, I wonder if this won't conflict with what we do in Poetry, and if the vendors of poetry-core will not take precedence.

pyproject.toml
```diff
@@ -21,12 +21,27 @@ classifiers = [
 packages = [
     {include = "poetry"},
 ]
+include = [
+    "poetry/core/_vendor/**/*"
```
sdispater, line 25: As far as I know this is not necessary since it will included by default, won't it?

poetry/core/__init__.py
abn: Good question, I want to do a bit more testing on this one actually. We could move it into `poetry/__init__.py`. Although, I am not a 100% on the ordering.

One thing I was wondering if we could provide a kind of proxy instead that can be smart enough to pick system one if the vendor site is not available. That would mean it would achieve what we want, while maintaining explicit imports.

pyproject.toml
abn: Hmm, you might be right since we are greedy when we do the package lookup I think. Although, I have dropped the `__init__.py` file, figured I'd be explicit here.

poetry/core/__init__.py
abn: For posterity, if the sys path option does  need to be replaced later. This can be an alternative.

```py
import sys
import importlib

from poetry.core.utils._compat import Path

__all__ = []
__vendored__ = {}

try:
    # poor developer's requirements parser
    with open(Path(__file__).parent / "vendor.txt") as f:
        for line in f.readlines():
            name, version = line.split("==")
        __vendored__[name] = version
        module_name = "poetry.core._vendor.{}".format(name)

        try:
            locals()[name] = importlib.import_module(module_name)
        except ImportError:
            locals()[name] = importlib.import_module(name)

        sys.modules[module_name] = locals()[name]
        __all__.append(name)
except (FileNotFoundError, ValueError, IndexError) as e:
    pass
```

pradyunsg: At that point, it might make sense to just switch to using pip's internal helper: `pip._vendor:vendored`.

PR #26, opened 2020-04-17
Add PEP517 check suite and support disabling setup.py for sdists

abn:
1. Adds scaffolding and a few integration tests for PEP517 tests.
2. Support opt-in disabling of setup.py in sdists.
3. Support "tool.poetry.build" section to support "generate-setup-file"

```toml
[tool.poetry.build]
generate-setup-file = true
```

This change will also mean that `tool.poetry.build = "build.py" usage` is now deprecated in favour of `tool.poetry.build.script`. The older cofiguration will continue to work as expected.

I have also added some additional converage for PEP-517 concerns in a new github actions workflow.

### Integration Test Suite Notes
To run pytest integration cases locally, you can do so using one of the following.

1. Use tox. This runs only the integration tests placed in `tests/integration`.
```sh
tox -e integration
``` 
2. Using pytest. Using the `--integration` option will enable all tests with the `integration` marker. You can restrict running only these tests by specifying `tests/integration` directory as an argument.
```sh
pytest --integration
```

If developing/adding tests, you can simply mark a test as integration, and it will only be executed when the integration suite is executed.


abn:
PEP-517 tests seem to pass okay locally on all versions, but CI does not like them. Need to investigate a bit further.

abn:
@sdispater any thought on the naming for the configuration section? The idea is to reuse this section for project specific configuration. Ideally would like to evolve this further later to provide project specific overrides etc.

poetry/core/masonry/builders/sdist.py
```diff
@@ -87,11 +87,12 @@ def build(self, target_dir=None):  # type: (Path) -> Path
                 else:
                     tar.addfile(tar_info)  # Symlinks & ?
 
-            setup = self.build_setup()
-            tar_info = tarfile.TarInfo(pjoin(tar_dir, "setup.py"))
-            tar_info.size = len(setup)
-            tar_info.mtime = time.time()
-            tar.addfile(tar_info, BytesIO(setup))
+            if not self._poetry.get_project_config("disable-setup-file", False):
```
stephsamson, line 90: How about having the config name to be `use-setup-file` so that it's easier to read here? Then users specify `use-setup-file = false` in their `pyproject.toml`.

abn: Might be better considering in the long run we want users to explicitly say they want to use the setup file. We could also consider something along the lines of `generate-`/`include-`/`provide-` iinstead too. Since, what we do is not generate a setup file to be inclided in the sdist.

finswimmer: +1 for `generate-setup-file`

sdispater:
The table to store the setting should be `tool.poetry.build.config` (or even a simpler `tool.poetry.build` where it would be possible to put build-related settings) instead of `tool.poetry.config`.

The reason for that is I don't want people to start thinking that they can put any setting in this section (we already have `poetry.toml` for that). So having a `build` section would be better. However this will break the projects of people currently using the `build` property to specify a build script.

I don't know how many users use this feature, and I don't mind breaking it this it was never documented and it needs improvements anyway but that's something to have in mind and we need to find a way to make it work.

abn:
@sdispater I fear quite a few people are using this feature as this is the only way today the extensions work, I am reluctant to recommend removing it as it is a useful feature.

Also, `tool.poetry.toml` is not really documented either as far as I can tell.

sdispater:
@abn When I said `poetry.toml`, I meant the `poetry.toml` file that's used for the `--local` option of the `config` command (see https://python-poetry.org/docs/configuration/#local-configuration).

If we are not comfortable in removing it, we will need to make it backward compatible, meaning we will need to support both (`build` as a string pointing to a build script and `build` as a table representing a build configuration).

So basically people would need to migrate from:

```toml
build = "my_script.py"
```

to

```toml
[tool.poetry.build.config]
script = "my_script.py"
generate-setup-file = false
```

abn:
@sdispater the change has been applied.

mattpeople:
Hello 2023 people!

Once again we have a great poetry point release breaking existing features.

This previous statement is no longer true:

> we will need to make it backward compatible, meaning we will need to support both (build as a string pointing to a build script and build as a table representing a build configuration).

The previous syntax is now gone and your builds are breaking until you use the new syntax (which i couldn't find/search in the docs anywhere? it only exists here?) of:

```ini
[tool.poetry.build.config]
script = "my_script.py"
generate-setup-file = true
```

phillipuniverse:
@mattpeople that config doesn't work, here's what I have:

```toml
  [tool.poetry.build.config]
  generate-setup-file = true
```

and I get this error:

```
The Poetry configuration is invalid:
  - [build] {'config': {'generate-setup-file': True}} is not valid under any of the given schemas
```

The one in the description of this PR does:

```toml
  [tool.poetry.build]
  generate-setup-file = true
```

PR #27, opened 2020-04-19
Add `private` property to configuration

utek:
Add `private` property to project properties. 

Required by https://github.com/python-poetry/poetry/pull/2322

### DIFF
--- a/poetry/core/factory.py
+++ b/poetry/core/factory.py
@@ -63,6 +63,8 @@ class Factory(object):
         package.keywords = local_config.get("keywords", [])
         package.classifiers = local_config.get("classifiers", [])
 
+        package.private = local_config.get("private", False)
+
         if "readme" in local_config:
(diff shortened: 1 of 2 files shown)


sdispater:
If we go the route of `toml.poetry.build` (or `toml.poetry.build.config`) for #26, we might as well have a `toml.poetry.publish` (or `toml.poetry.publish.config`) to configure the publishing aspects of Poetry that are inherent to the project.

However, this begs the question: should it be integrated in `poetry-core` since the publishing part is not part of it?

abn:
We have a bit of duplication in core and poetry. We should probably discuss how to consolidate it such that poetry code builds on top of core. But, I agree, we should introduce publish session within poetry and not core. Unless ofcourse we move the entire configuration schema to core.

alkasm:
#26 has been merged in but this has sat stale since. Hoping to see python-poetry/poetry#2322 merged so #1537 can be closed out. Any updates?

PR #30, opened 2020-04-25
Fix git dependencies

simonepri:
Resolves: https://github.com/python-poetry/poetry/issues/2326, https://github.com/huggingface/transformers/issues/3982


poetry/core/utils/helpers.py
```diff
@@ -81,6 +81,8 @@ def parse_requires(requires):  # type: (str) -> List[str]
             continue
 
         if current_marker:
+            if "@" in line:
```
simonepri, line 84: Is that sufficient to check whether it is a URL requirement?

poetry/core/utils/helpers.py
```diff
@@ -81,6 +81,8 @@ def parse_requires(requires):  # type: (str) -> List[str]
...
+                line += " "
             line = "{}; {}".format(line, current_marker)
```
abn, line 86:
```suggestion
        if current_marker:
            line = "{} ; {}".format(line, current_marker)
```

I wonder if this could introduce some edge case. I'd just add the whitespace in all cases.

tests/conftest.py
```diff
@@ -3,7 +3,7 @@
 import pytest
 
 from poetry.core.utils._compat import Path
-from tests.utils import tempfile
+from tests._helpers.utils import tempfile
```
abn, line 6: Probably just move utils.py to `tests/utils/__init__.py`

tests/packages/test_main.py
```diff
@@ -183,6 +183,22 @@ def test_dependency_from_pep_508_with_git_url():
...
+        ' ; extra == "foo;"'
+    )
+
+    dep = dependency_from_pep_508(name)
+
+    assert "poetry" == dep.name
+    assert dep.is_vcs()
+    assert "git" == dep.vcs
+    assert "https://github.com/python-poetry/poetry.git" == dep.source
+    assert "b;ar;" == dep.reference
```
abn, line 198: Is this a valid scenario?

simonepri: Yes, you can habe a branch with `;` in the name.
See https://github.com/python-poetry/core/blob/fbfcda73b2c5524c3eda29b1d9b70734a3a9dcf3/tests/packages/test_vcs_dependency.py#L55

tests/conftest.py
simonepri: Hiding test helpers inside the `__init__.py` does not seem ideal to me.
Is that common?

An alternative that I like more, is to just rename it to something else like `testutils`

poetry/core/utils/helpers.py
simonepri: If we always add the space we break the old `parse_requires` tests (because the space was not expected).

abn: I am okay with the test's expected output needing updates as long it won't break any existing use cases. As far as I can tell, this change shouldn't affect anything reading metadata we generate.

simonepri: No it shouldn't, but I preferred to stick with the same representation generated here:
https://github.com/python-poetry/core/blob/e1c6978c5a0c104d373d78bcfc10a722409de583/poetry/core/_vendor/packaging/requirements.py#L123-L141

poetry/core/packages/__init__.py
```diff
@@ -63,8 +63,8 @@ def dependency_from_pep_508(
     name = parts[0].strip()
     if len(parts) > 1:
         rest = parts[1]
-        if ";" in rest:
-            name += ";" + rest.split(";", 1)[1]
+        if " ;" in rest:
+            name += " ;" + rest.split(" ;", 1)[1]
```
abn, line 67:
```suggestion
        if ";" in rest:
            name += " ;" + rest.split(";", 1)[1].strip()
```

poetry/core/utils/helpers.py
abn: Understandable; but this use case is a bit different as the objective here is to convert a requirement txt file into a list of requires_dist. And I suspect using `@` to check if url is not really reliable. Considering that the extra whitespace is good hygene in all cases, I'd prefer that we applied the above suggestion.

poetry/core/packages/__init__.py
simonepri: This won't work if the url contains `;`.

poetry/core/utils/helpers.py
simonepri: Done

simonepri: Done

simonepri:
@abn Any update on this?

abn:
I will get to this one this week hopefully. I wanted to go through the relevant peps before getting back to you.

tholiao:
This PR fixes an issue with a widely used library for natural language processing; there are two major conference deadlines in a month so any expedition would be greatly, greatly appreciated!

simonepri:
@abn Sorry for pinging you again.
Do we have any blocker for this?

PR #32, opened 2020-04-30
Allow url dependencies in multiple constraints dependencies

sdispater:
Resolves: https://github.com/python-poetry/poetry/issues/2033

This PR is just a port of the following PR made on the main Poetry project: https://github.com/python-poetry/poetry/pull/2035

### DIFF
--- a/poetry/core/json/schemas/poetry-schema.json
+++ b/poetry/core/json/schemas/poetry-schema.json
@@ -496,6 +496,9 @@
                     },
                     {
                         "$ref": "#/definitions/path-dependency"
+                    },
+                    {
+                        "$ref": "#/definitions/url-dependency"
                     }
                 ]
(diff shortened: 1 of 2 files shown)


poetry/core/json/schemas/poetry-schema.json
```diff
@@ -479,7 +479,7 @@
         },
         "multiple-constraints-dependency": {
             "type": "array",
-            "minItems": 1,
+            "minItems": 2,
```
abn, line 482: Should we enforce this? It might break some existing projects. And it is a nice to have to be able to specify an array of 1 :)

sdispater: Thanks fro pointing this out :-) I changed it back to `1`.

PR #39, opened 2020-06-27
Run `pre-commit autoupdate`. Fix resulting flake8 errors

hukkinj1:


hukkinj1:
The CI failure here (https://github.com/python-poetry/poetry-core/pull/39/checks?check_run_id=814831965) seems like a Poetry bug to me at first glance.

If I'm reading it right, seems like Poetry wants to install `futures` because `isort` requires it but the `futures` version doesn't support Python versions >=3.0. `poetry-core` however does support python ^3.5. What Poetry doesn't seem to understand is that isort (and thus poetry-core) only needs `futures` when python_version < "3.2", so I believe there shouldn't be a conflict.

I could perhaps simply remove `isort` from dev-dependencies as it seems we only run it in `pre-commit` anyways?

hukkinj1:
I tried to remove the `isort` dev-dependency, but it did not fix the failing CI, because another dependency had the same conditional `futures` dependency, so the same issue persisted.

I guess the root cause should be fixed then.

For the record, here's Poetry output from the failing Python 2.7 CI run:
```
SolverProblemError

The current project's Python requirement (~2.7 || ^3.5) is not compatible with some of the required packages Python requirement:
  - futures requires Python >=2.6, <3, so it will not be satisfied for Python >=3.5,<4.0

Because no versions of isort match >4.3.21,<5.0.0
 and isort (4.3.21) depends on futures (*), isort (>=4.3.21,<5.0.0) requires futures (*).
Because futures (3.3.0) requires Python >=2.6, <3
 and no versions of futures match !=3.3.0, futures is forbidden.
Thus, isort is forbidden.
So, because poetry-core depends on isort (^4.3.21), version solving failed.
```

hukkinj1:
Rebased on master. The CI failure seems fixed now with poetry 1.1.0a3

hukkinj1:
I did a rebase and fixed conflicts, after which there are no more major changes left here anymore. Seems like the changes have been committed by another contributor after this PR was opened.

finswimmer:
Hello @hukkinj1,

thanks a lot for your contribution!

> I did a rebase and fixed conflicts, after which there are no more major changes left here anymore.

Yes, I think this is due to this PR https://github.com/python-poetry/poetry-core/pull/63

For the changes left in your PR please see my comment here about it: https://github.com/python-poetry/poetry-core/pull/63#discussion_r476141267

I will close your PR in favor of the above mentioned.

Thanks again :+1: 

fin swimmer

PR #45, opened 2020-07-17
Add build script without setup.py support

sdispater:
Since we can now disable the generation of the `setup.py` file, we need a way to make the build scripts feature work without it.

This PR takes care of it by assuming that if the generation is disabled and a build script file is referenced then the build script is self sufficient and will take care of compiling and copying necessary files to the project. An example of such a build script has been added to the test fixtures.


poetry/core/masonry/builders/wheel.py
```diff
@@ -91,44 +96,59 @@ def build(self):
 
     def _build(self, wheel):
         if self._package.build_script:
-            with SdistBuilder(poetry=self._poetry).setup_py() as setup:
-                # We need to place ourselves in the temporary
-                # directory in order to build the package
+            if not self._poetry.package.build_config.get("generate-setup-file", True):
```
abn, line 99: We should probably make this a property of the package (`package.build_generate_setup_file`) or have something like `package.build.generate_setup_file`; we can handle this later if required but otherwise we will end up with multiple places to update default value etc.

poetry/core/masonry/builders/wheel.py
```diff
@@ -91,44 +96,59 @@ def build(self):
...
+                # Since we have a build script but no setup.py generation is required,
+                # we assume that the build script will build and copy the files
+                # directly.
```
abn, line 102: Would this be a problematic assumption? Aren't we encouraging different behaviour based on an unrelated configuration? ie. this could break existing build scripts if the maintainer chooses to switch of setup file generation (or when we switch to it being disabled by default).

sdispater: The `build` script feature of Poetry was never part of the public API and an undocumented feature up until now since I knew from the start that we would no longer rely on `setup.py` at some point and that the current implementation was a hack more than anything. So breaking this does not feel too problematic to me. Also I don't think a lot of people relies on it at the moment.

Delegating the bulk of the work to the `build` script was always the plan so that people can build their extensions any way they see fit, without being tied to a specific build tool like `setuptools`.

PR #46, opened 2020-07-18
"include" in pyproject.toml now relates to "package_data" in setup.py.

atollk:
Before this change, only package-includes were considered when building the package_data field in setup.py. "include = [...]" fields were ignored which made using PEP 561-specified "py.typed" impossible.

Resolves: https://github.com/python-poetry/poetry/issues/1338


stephsamson:
@atollk This looks good to me. Can you rebase it against master to fix conflicts? Thank you for your contribution!

tests/masonry/builders/test_complete.py
```diff
@@ -507,7 +507,7 @@ def test_package_with_include(mocker):
             "src_package",
             "tests",
         ]
-        assert ns["package_data"] == {"": ["*"]}
+        assert ns["package_data"] == {"": ["*"], "extra_dir": ["vcs_excluded.txt"]}
```
abn, line 510: I am unsure if this behaviour is actually desired. Since `extra_dir` is not necessarily a package. According to the [`setuptools` documenation](https://setuptools.readthedocs.io/en/latest/setuptools.html), this needs to be a package name.

tests/masonry/builders/test_sdist.py
```diff
@@ -565,6 +565,23 @@ def test_sdist_package_pep_561_stub_only():
...
+    builder.build()
+
+    sdist = root / "dist" / "pep-561-0.1.tar.gz"
+
+    assert sdist.exists()
+
+    with tarfile.open(str(sdist), "r") as tar:
+        assert "pep-561-0.1/setup.py" in tar.getnames()
+        setup_file = tar.extractfile("pep-561-0.1/setup.py").read().decode("utf-8")
+    assert "package_data = \\\n{'': ['*'], 'pep_561': ['py.typed']}" in setup_file
```
abn, line 582: Doesn't `'': ['*']` already include `'pep_561': ['py.typed']`? Seems redundant; especially since we are moving towards disabling `setup.py` files by default.

abn (changes requested):

staticdev:
@atollk I am almost sure this fixes my problem. Just to make sure...
I add `include = ["src/package/locale/**/*.mo"]` and it creates on wheel/bdist the mo files in the root of the package (src/package/locale) instead of inside the package (package/locale).

staticdev:
@atollk would we be able to have this on the next release?

atollk:
@staticdev I haven't pursued this topic further since @abn 's opinion was that the change should not be added at all.

PR #53, opened 2020-07-27
Improve package equality check and repr

abn:
We need to be a bit more stricter with how we compare packages as prior to this we would check two packages,  `foo-1.0.0` with `source_reference=x` and  `foo-1.0.0` with `source_reference=y` as the same. Which can lead to some unexpected issues at poetry runtime and incorrect validation in poetry's test suites as well.


abn:
@sdispater any concerns on this one?

PR #55, opened 2020-07-28
semver: fix PEP 440 compatible release parsing

abn:
Relates-to: python-poetry/poetry#2734


sdispater:
It should fix https://github.com/python-poetry/poetry/issues/2737 as well.

abn:
> It fixes the root cause, but not the fact that it failed silently.

@ewjoachim yes that is correct. Unfortunately, we need a better way to handle those logs. The issue is that the failure was handled as expected in the sense that current logic explicitly handles the issue by skipping the dependency. Not ideal. Either we need to raise an error gracefully, or simply propagate the relevant log correctly.

PetterS:
> It fixes the root cause, but not the fact that it failed silently.

Completely agree that we need to aim for much more robustness.

ewjoachim:
@PetterS the "failing silently" part was addressed in #2738 & #2739 by @abn and @sdispater ;)

PR #57, opened 2020-08-10
Respect REUSE spec when including licenses

MrJohz:
* The [REUSE spec](https://reuse.software/spec/) states that a LICENSES folder should exist with copies of all of the licenses used in this project.
* Previously, poetry tried to include all files like LICENSE*, but this didn't respect directories, meaning that poetry would fail if a LICENSES folder existed. (This is the python-poetry/poetry#2565 bug.)
* Now poetry is a bit more specific about the "correct" license files to add
  * LICENSE or COPYING
  * LICENSE.* or COPYING.* (i.e. with any extension)
  * LICENSES/** (i.e. all files within the LICENSES directory)
* In addition, if a suggested licensing file doesn't exist, or isn't a file, poetry will now simply ignore it, rather than throwing an error.

Resolves: python-poetry/poetry#2565

- [ ] Updated **documentation** for changed code. (I don't think this is necessary, as there isn't any documentation for the original behaviour?  I can add some if that would be preferred.)

This is a copy of python-poetry/poetry#2573, which I will now close given that the relevant code has been moved to poetry-core.


milliams:
As nothing more than a user who has been hit by python-poetry/poetry#2565, this looks like it would solve my issue which is currently preventing me from releasing.

Also, indeed I can't see anywhere in the documentation that would need updating.

poetry/core/masonry/builders/wheel.py
```diff
@@ -168,9 +168,20 @@ def _write_metadata(self, wheel):
...
                 self._write_entry_points(f)
 
+        license_files_to_add = []
         for base in ("COPYING", "LICENSE"):
-            for path in sorted(self._path.glob(base + "*")):
-                self._add_file(wheel, path, "%s/%s" % (self.dist_info, path.name))
+            license_files_to_add.append(self._path / base)
+            license_files_to_add.extend(self._path.glob(base + ".*"))
+
+        license_files_to_add.extend((self._path / "LICENSES").glob("**" + os.sep + "*"))
```
abn, line 176:
```suggestion
        license_files_to_add.extend(self._path.glob("LICENSES/**/*"))
```
pathlib is smart enough to handle OS diferences.

poetry/core/masonry/builders/wheel.py
```diff
@@ -168,9 +168,20 @@ def _write_metadata(self, wheel):
             with self._write_to_zip(wheel, self.dist_info + "/entry_points.txt") as f:
                 self._write_entry_points(f)
 
+        license_files_to_add = []
         for base in ("COPYING", "LICENSE"):
-            for path in sorted(self._path.glob(base + "*")):
-                self._add_file(wheel, path, "%s/%s" % (self.dist_info, path.name))
+            license_files_to_add.append(self._path / base)
+            license_files_to_add.extend(self._path.glob(base + ".*"))
```
abn, line 174:
```suggestion
            license_files_to_add.append(base)
            license_files_to_add.extend(self._path.glob(base + ".*"))
```

poetry/core/masonry/builders/wheel.py
```diff
@@ -168,9 +168,20 @@ def _write_metadata(self, wheel):
...
+
+        license_files_to_add.extend((self._path / "LICENSES").glob("**" + os.sep + "*"))
+
+        for path in license_files_to_add:
```
abn, line 178:
```suggestion
        for path in set(license_files_to_add):
```

poetry/core/masonry/builders/wheel.py
```diff
@@ -168,9 +168,20 @@ def _write_metadata(self, wheel):
...
+            if not path.is_file():
+                continue
+
+            self._add_file(
+                wheel, path, "%s/%s" % (self.dist_info, path.relative_to(self._path))
+            )
```
abn, line 184:
```suggestion
            if path.is_file():
                self._add_file(
                    wheel, path, "%s/%s" % (self.dist_info, path.as_posix())
                )
```

poetry/core/masonry/builders/wheel.py
MrJohz: Useful to know, thanks!

MrJohz: It turns out that this isn't the case when using the `.glob` method.  I'm not sure why it isn't normalised, but you can see in [this](https://github.com/python-poetry/poetry-core/runs/993745827) run when this change was added that the only failure was in the Windows builds.

I've reverted this change as a result.

poetry/core/masonry/builders/wheel.py
MrJohz: This line (and the `path.relative_to(...)` line later) are necessary because the license files must be found relative to `self._path`, and because otherwise, when we add the files to the zip file, they'll come out containing absolute parts (e.g. `project/tmp/testdir-7843923/...`).

We could probably just use a string here (that was my first attempt), but juggling a mixed array of strings _and_ `Path` objects made the code more complicated, whereas converting everything into an absolute `Path` object relative to the `self._path` variable means that when we later add the files to the zip, we need to do a lot less work to make sure they're all in the right format.

MrJohz:
@abn - I worked through your suggestions, but they didn't all work out so successfully because of various weird reasons - most notably, the `.glob` method in Windows _doesn't_ normalise path separators.

I replied to your comments inline - could you advise whether the new changes are sufficient, or whether I should make other changes?

Thanks!

poetry/core/masonry/builders/wheel.py
abn: I'll check shortly but this might be related to https://bugs.python.org/issue31202

poetry/core/masonry/builders/wheel.py
abn: Hmm, that is fine. I just wanted to avoid the possibility of ending up with a ValueError. But considering we control what goes into the list, this should be fine. No strong feelings on this one :).

MrJohz: Yeah, I tried `Path(base)` briefly, but I think it's helpful to have either all the paths as absolute, or none of them.  But I think it works reasonably like this, and the fact that tests fail when I go for different options gives me more confidence that this is a fairly stable approach that shouldn't run into `ValueError` problems.

poetry/core/masonry/builders/wheel.py
abn: Looks like it is. In this case; can you try this suggestion? This should also confirm that the issue is as identified above or if pathlib does not handle os diferences as advertised.

```suggestion
        license_files_to_add.extend(self._path.joinpath("LICENSES").glob("**/*"))
```

MrJohz: It could be related - it's odd that it only affects things when the "incorrect" pathsep is used though!

Even if it were fixed, Poetry would still need to support the older versions where the fix hasn't been implemented yet, so I think that means we're stuck with `sep` for now.

abn (approved):

PR #61, opened 2020-08-13
Make sure filenames containing commas are quoted when writing RECORD file for wheel

finswimmer:
This PR uses the `csv` module, to make sure filenames containing commas are quoted, when writing the RECORD file for the wheel packages.

Resolves: https://github.com/python-poetry/poetry/issues/2052

### DIFF
--- a/poetry/core/masonry/builders/wheel.py
+++ b/poetry/core/masonry/builders/wheel.py
@@ -13,11 +13,13 @@ import tempfile
 import zipfile
 
 from base64 import urlsafe_b64encode
+from io import BytesIO
 from io import StringIO
 
 from packaging.tags import sys_tags
 from poetry.core import __version__
 from poetry.core.semver import parse_constraint
+from poetry.core.utils._compat import PY2
 from poetry.core.utils._compat import decode
 
 from ..utils.helpers import escape_name
@@ -180,6 +182,9 @@ class WheelBuilder(Builder):
         # Write a record of the files in the wheel
         with self._write_to_zip(wheel, self.dist_info + "/RECORD") as f:
             record = StringIO()
+            if PY2:
+                record = BytesIO()
+
             csv_writer = csv.writer(
                 record, delimiter=str(","), quotechar=str('"'), lineterminator="\n"
             )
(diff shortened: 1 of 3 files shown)


poetry/core/masonry/builders/wheel.py
```diff
@@ -188,7 +193,12 @@ def _write_record(self, wheel):
 
             # RECORD itself is recorded with no hash or size
             csv_writer.writerow((self.dist_info + "/RECORD", "", ""))
-            f.write(record.getvalue())
+
+            record_values = record.getvalue()
```
abn, line 197: You should be able to rely on the compat methods that are already present to handle this instead of introducing compat logic here.

poetry/core/masonry/builders/wheel.py
```diff
@@ -181,7 +181,7 @@ def _write_record(self, wheel):
         with self._write_to_zip(wheel, self.dist_info + "/RECORD") as f:
             record = StringIO()
             csv_writer = csv.writer(
-                record, delimiter=",", quotechar='"', lineterminator="\n"
+                record, delimiter=str(","), quotechar=str('"'), lineterminator="\n"
```
abn, line 189:
```suggestion
                record, delimiter=csv.excel.delimiter, quotechar=csv.excel.quotechar, lineterminator="\n"
```

abn: Alternatively you can set `dialect='excel'` which is the default I think?

finswimmer:
Thanks for your review @abn. I've included your suggestions.

poetry/core/masonry/builders/wheel.py
```diff
@@ -179,10 +181,23 @@ def _write_metadata(self, wheel):
     def _write_record(self, wheel):
         # Write a record of the files in the wheel
         with self._write_to_zip(wheel, self.dist_info + "/RECORD") as f:
+            record = StringIO()
+            if PY2:
```
abn, line 185: Is this required since we are getting the compat method to deal with this later on?

finswimmer: Yep, the csv writer doesn't like to work well with the StringIO.

poetry/core/masonry/builders/wheel.py
```diff
@@ -179,10 +181,23 @@ def _write_metadata(self, wheel):
...
+                record = BytesIO()
```
abn, line 186:
```suggestion
            record = BytesIO() if PY2 else StringIO()
```
Minor suggestion to be a bit more concise here.

finswimmer: Ok, as you like :) But I changed the order, to make clearer what's the default and we can just kick the end of the line once python2 isn't supported anymore

PR #63, opened 2020-08-15
Update vendors, fixup dependencies, cleaup qa issues, add pre-commit action workflow

abn:
This change contains various minor code quality related fixups. This also includes an update to vendored packages.


poetry/core/masonry/builders/builder.py
```diff
@@ -36,7 +41,7 @@ class Builder(object):
 
     def __init__(
         self, poetry, ignore_packages_formats=False
-    ):  # type: ("Poetry", bool) -> None
+    ):  # type: (Poetry, bool) -> None
```
finswimmer, line 44: I think we should keep the quotes here and in the other places. `Poetry` isn't imported during runtime. While this works in the "python2 way" of type hinting, the quotes are needed in the "python3 way". We might want to switch to this style, once python2 support is dropped. Keeping the quotes makes it easy to use automatic converter for this step.

poetry/core/vcs/git.py
```diff
@@ -127,8 +127,8 @@ def url(self):  # type: () -> str
             "/" + self.pathname.lstrip(":/"),
         )
 
-    def format(self):
-        return "{}".format(self.url, "#{}".format(self.rev) if self.rev else "",)
+    def format(self):  # type: () -> str
+        return self.url
```
finswimmer, line 131: Not sure here. Shouldn't it be `return "{}".format(self.url, "#{}{}".format(self.rev) if self.rev else "",)`?

pyproject.toml
```diff
@@ -43,16 +43,12 @@ functools32 = {version = "^3.2.3-2", python = "~2.7"}
...
-pytest-cov = "^2.8.1"
-pytest-mock = "^2.0.0"
+pytest-cov = "^2.8"
+pytest-mock = "^2.0"
 tox = "^3.0"
-cleo = "^0.8.1"
-tomlkit = "^0.5.8"
-vendoring = {version = "^0.2.2", python = "~3.8"}
-isort = "^4.3.21"
+vendoring = {version = "^0.3", python = "~3.8"}
```
finswimmer, line 49: Do you remove `tomlkit` here, because it's one of the vendored packages?

finswimmer (changes requested):
Some minor comments :)

poetry/core/vcs/git.py
abn: Well I just removed the redundant parameter for now; since adding that will start failing tests etc. So, decided to retain the current behaviour (ie. not add the `#` bit). We shold fix this elsewhere since this change has no functional change.

pyproject.toml
abn: Correct. These shouldnt be installed as a dev dependency, but rahter `poetry.core._vendors` added to `PYTHONPATH` during development. Else you can end up with some tests passing but distrubuted package failing.

poetry/core/masonry/builders/builder.py
abn: Good catch.

abn:
@finswimmer this is ready again :)

poetry/core/vcs/git.py
finswimmer: Ok, makes sense.

finswimmer (approved):

PR #64, opened 2020-08-15
dev: add makefile with vendor/sync target

abn:
This is to help maintainers manage vendored packages a bit easier.


stephsamson (approved):
Nice! 👍

Issue #2, opened 2020-03-16
No Readme.md file available


stealthman22:
Hello.
Please without a read me, we will have no idea what this project is about.
It Would be nice to add one.

boukeversteegh:
Hesitant to use this package as our build backend without any documentation. Thanks!

FredHappyface:
So to clarify is this project solely for building poetry or can it be used by third parties? 

Is it planned for instance to outline what functionality is covered here and what is covered by poetry 'full'?

Cheers

abn:
@FredHappyface I have added some information in #66. This package is intended to be a build backend fully compliant with PEP 517. Some shared functionality like configuration/`pyproject.toml`/`poetry.lock` parsing will also be available here. Poetry itself will use this package for everythign that is implemented here.

Right now, we do not intend formalise a public API other than what is required for PEP 517. But we do want to make things more accessible for downstream developers so they don't have to reinvent the wheel when building poetry plugins etc.

PR #66, opened 2020-08-15
Update project urls and readme content

abn:
Resolves: #2

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -8,8 +8,8 @@ license = "MIT"
 
 readme = "README.md"
 
-homepage = "https://github.com/python-poetry/core"
-repository = "https://github.com/python-poetry/core"
+homepage = "https://github.com/python-poetry/poetry-core"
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -26,6 +26,9 @@ exclude = [
     "**/*.pyi",
 ]
 
+[tool.poetry.urls]
+"Bug Tracker" = "https://github.com/python-poetry/poetry/issues"
+
```
finswimmer, line 31: I think we should also add a paragraph to the `README.md`, that issues belonging to `poetry-core` must be reported in the issue tracker of `poetry`.

abn: That's coming later in the contributing md file.

finswimmer: Ok, I see that we already send people to the poetry issue tracker, when they try to open an issue in poetry-core. So I think that's fine.

PR #71, opened 2020-08-27
Experimental: Add support for building pinned artifacts

abn:
This change allows poetry projects to be configured via pyproject.toml
to optionally enable version pinning and also to include nested
dependencies for application wheels.

**NOTE:** This is only a proposal, intended to serve as a basis for internal discussion.


franziskakuehn:
is this going to be accepted?

is this a fat wheel approach?

radoering:
> is this going to be accepted?

This is quite outdated, so probably not in the near future. Without taking a closer look, I assume that python-poetry/poetry#9428 is a more recent variant of this, which produces more reliable markers but is only implemented in poetry, not in poetry-core; in other words: not in the build-backend.

> is this a fat wheel approach?

No, it is about pinning dependencies instead of allowing version ranges.

PR #77, opened 2020-09-02
execute SdistBuilder and WheelBuilder separately

finswimmer:
The PR removes the `CompleteBuilder` when building both `sdist` and `wheel`. Instead the `SdistBuilder` and `WheelBuilder` are executed sequential.

Resolves: https://github.com/python-poetry/poetry/issues/1332
Resolves: https://github.com/python-poetry/poetry/issues/2046

- [x] Updated **tests** for changed code.


poetry/core/masonry/builder.py
```diff
@@ -14,6 +16,5 @@ def build(self, fmt):
         if fmt not in self._FORMATS:
             raise ValueError("Invalid format: {}".format(fmt))
 
-        builder = self._FORMATS[fmt](self._poetry)
-
-        return builder.build()
+        for builder in self._FORMATS[fmt]:
```
abn, line 19:
```suggestion
        builders = [self._FORMAT[fmt]] if fmt else self._FORMATS.values()
        for builder in builders:
```

Alternatively `listify` if not a list.

Something like that might be better here I reckon. The reason being using a list for for each format does not really make sense since `all` is an exception and we can derive it. 

We might also want to keep backwards compatible `CompleteBuilder` with a deprecation warning at least for one release cycle.

abn (changes requested):

poetry/core/masonry/builder.py
finswimmer: Yes, I wasn't happy about the lists as well. But from the solutions I have tried, it was the most beautiful to me :) 

You suggestion would need a change in the poetry repo as well. Because there `"all"` is passed to method and not nothing. 

https://github.com/python-poetry/poetry/blob/17b165bf8c8d989d672417fd4610721df59ae1d0/poetry/console/commands/build.py#L24-L36

I can do so if you like.

> We might also want to keep backwards compatible CompleteBuilder with a deprecation warning at least for one release cycle.

To whom might this be interesting?

abn: The issue is that we do not know who is actually using this already. That said, since we have not yet released a bet for core yet, I think it is reasonable to break it. I am not for it a 100% but in this case we can simply hope no one used the `CompleteBuilder` explicitly

abn (approved):

PR #78, opened 2020-09-03
Refactor sources and extras handling on packages and dependencies

sdispater:
Resolves: https://github.com/python-poetry/poetry/issues/2545
Resolves: https://github.com/python-poetry/poetry/issues/2494
Resolves: https://github.com/python-poetry/poetry/issues/2300
Resolves: https://github.com/python-poetry/poetry/issues/2080
Resolves: https://github.com/python-poetry/poetry/issues/1609

This is a preliminary work to solve the referenced issues on Poetry's end.

This PR refactors and improves the way packages and dependencies are compared to each other, by taking into account their source and their features/extras.

This the first step of a complete refactor of packages and sources that will be done in the 1.2 release of Poetry.

### DIFF
--- a/poetry/core/factory.py
+++ b/poetry/core/factory.py
@@ -1,9 +1,14 @@
 from __future__ import absolute_import
 from __future__ import unicode_literals
 
+import logging
+
+from typing import Any
 from typing import Dict
 from typing import List
 from typing import Optional
+from typing import Union
+from warnings import warn
 
 from .json import validate_object
 from .packages.dependency import Dependency
@@ -12,6 +17,9 @@ from .poetry import Poetry
 from .pyproject import PyProjectTOML
 from .spdx import license_by_id
 from .utils._compat import Path
+
+
+logger = logging.getLogger(__name__)
 
 
 class Factory(object):
@@ -71,21 +79,38 @@ class Factory(object):
 
                 if isinstance(constraint, list):
                     for _constraint in constraint:
-                        package.add_dependency(name, _constraint)
+                        package.add_dependency(
+                            self.create_dependency(
+                                name, _constraint, root_dir=package.root_dir
+                            )
+                        )
 
                     continue
 
-                package.add_dependency(name, constraint)
+                package.add_dependency(
+                    self.create_dependency(name, constraint, root_dir=package.root_dir)
+                )
 
         if "dev-dependencies" in local_config:
             for name, constraint in local_config["dev-dependencies"].items():
                 if isinstance(constraint, list):
                     for _constraint in constraint:
-                        package.add_dependency(name, _constraint, category="dev")
+                        package.add_dependency(
+                            self.create_dependency(
+                                name,
+                                _constraint,
+                                category="dev",
+                                root_dir=package.root_dir,
+                            )
+                        )
 
                     continue
 
-                package.add_dependency(name, constraint, category="dev")
+                package.add_dependency(
+                    self.create_dependency(
+                        name, constraint, category="dev", root_dir=package.root_dir
+                    )
+                )
 
         extras = local_config.get("extras", {})
         for extra_name, requirements in extras.items():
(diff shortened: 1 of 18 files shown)


poetry/core/packages/package_specification.py
```diff
@@ -0,0 +1,118 @@
+from typing import FrozenSet
+from typing import List
+from typing import Optional
+
+from poetry.core.utils.helpers import canonicalize_name
+
+
+class PackageSpecification(object):
```
abn, line 8: :heart:

poetry/core/packages/package_specification.py
```diff
@@ -0,0 +1,118 @@
...
+    def __init__(
+        self,
+        name,
+        source_type=None,
+        source_url=None,
+        source_reference=None,
+        source_resolved_reference=None,
```
abn, line 15: It would be good for these to move into an object (and later to a dataclass once we drop 2.7). It becomes easier to maintain.

abn (changes requested):
Overall I like the direction we are heading in. Some comments.

poetry/core/packages/package_specification.py
```diff
@@ -0,0 +1,118 @@
...
+
+        return name
+
+    @property
+    def source_type(self):  # type: () -> Optional[str]
+        return self._source_type
+
+    @property
+    def source_url(self):  # type: () -> Optional[str]
+        return self._source_url
```
finswimmer, line 53: At least [here](https://github.com/python-poetry/poetry/blob/17b165bf8c8d989d672417fd4610721df59ae1d0/poetry/repositories/installed_repository.py#L146-L156) we also need setters for these properties. Don't know about the others for now :)

finswimmer (changes requested):

poetry/core/packages/package_specification.py
sdispater: After the release of the `1.2` version of Poetry I plan on refactoring this bit by using sources objects that will be attached to dependency objects.

This will remove the need to have multiple dependency classes.

poetry/core/factory.py
```diff
@@ -134,6 +159,154 @@ def create_poetry(self, cwd=None):  # type: (Optional[Path]) -> Poetry
 
         return Poetry(poetry_file, local_config, package)
 
+    def create_dependency(
```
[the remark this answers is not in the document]
sdispater: I will change this one to a class method. I'll leave the `create_poetry()` one as is for now to avoid unnecessary clutter to this PR.

We can always do this in a separate PR.

poetry/core/factory.py
```diff
@@ -134,6 +159,154 @@ def create_poetry(self, cwd=None):  # type: (Optional[Path]) -> Poetry
...
+                        extras=constraint.get("extras", []),
+                    )
+                else:
+                    dependency = DirectoryDependency(
+                        name,
+                        path,
+                        category=category,
+                        optional=optional,
+                        base=root_dir,
+                        develop=constraint.get("develop", True),
```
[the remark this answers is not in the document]
sdispater: I made the change so that directory dependencies are not in `develop` mode by default.

poetry/core/packages/dependency.py
```diff
@@ -49,21 +62,25 @@ def __init__(
...
         self._transitive_python_versions = None
         self._transitive_python_constraint = None
         self._transitive_marker = None
 
-        self._extras = []
+        if extras is None:
+            self._extras = frozenset()
+        else:
+            self._extras = frozenset(extras)
+
```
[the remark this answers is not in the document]
sdispater: Done :-)

poetry/core/packages/package_specification.py
sdispater: Eventually, I want them to be set at objects' initialization and not be settable (see my comment above about sources objects).

I don't want to give the impression that settings them is part of a public API that people can use.

abn: Woops wrong button. :(

@sdispater we should really discuss what an "api" is for both core and poetry. Define it, document it and refactor to make it possible at a target release.

poetry/core/factory.py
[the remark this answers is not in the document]
abn: Sounds reasonable.

poetry/core/factory.py
```diff
@@ -134,6 +159,154 @@ def create_poetry(self, cwd=None):  # type: (Optional[Path]) -> Poetry
...
+                if platform:
+                    marker = marker.intersect(
+                        parse_marker(
+                            create_nested_marker(
+                                "sys_platform", parse_generic_constraint(platform)
+                            )
+                        )
+                    )
+            else:
+                marker = parse_marker(markers)
```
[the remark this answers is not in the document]
sdispater: `markers` will always have priority over `python` and `platform` so they are indeed mutually exclusive.

Regarding the propagation of markers, I prefer we keep the current implementation, for now at least, since it's the most accurate way to have platform agnostic lock files.

We are not ready to have self sufficient lock files and it would most likely need to rethink the current design.

poetry/core/packages/package_specification.py
sdispater: @abn I think we will be able to discuss this API after the 1.1 release. However, what I can say is that the versioning of Poetry follows semantic versioning only for the CLI at the moment, even though we try to have a sensible backwards compatibility policy.

poetry/core/factory.py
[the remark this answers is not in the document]
abn: I am still not fully convinced about that rationale as this is not necessarily about being platform agnostic, but rather making sure any markers specified for root dependencies are propagated down. I do not believe this will chnage the status quo per say. Ofcourse I might be missing some pieces here. 

That said I am happy to take this particular discussion elsewhere.

As for being mutually exclusing I have a feeling this does have UX consequences. Since this is the existing behaviour we should simply doc it in another PR to avoid the confusion.

abn (approved):

abn:
@sdispater LGTM one comment for offline discussion maybe is https://github.com/python-poetry/poetry-core/pull/78/files#r489371619

poetry/core/packages/package_specification.py
```diff
@@ -0,0 +1,118 @@
+from typing import FrozenSet
```
abn, line 1: @sdispater one minor request can we rename this to `poetry.core.packages.specification` please?

abn (changes requested):

poetry/core/packages/package_specification.py
sdispater: @abn You only want to the module name to be changed, right? Not the name of the class? `poetry.core.packages.specification` would provide the `PackageSpecification`.

abn: Correct. Feels a bit cleaner.

sdispater: Done :-)

abn (approved):

poetry/core/packages/package_specification.py
abn: Considering that we will need to change code anyway with the `PackageSpecification` class I think explicitly setting `_source_url` for now should be fine.

PR #79, opened 2020-09-06
Fix version union constraint not properly generated

tomzx:
Migrated from https://github.com/python-poetry/poetry/pull/1767.

When we have a union of constraints, those were not properly generated.
For example, `^1.2.3,!=1.3.5` would be converted to `^1.2.3,!=1.3.5` (same
string) under PEP508, which is not valid. This PR will instead produce
`>1.2.3,<2.0.0,!=1.3.5`.

The approach may not be the best/cleanest, so I'm open to hear suggestions of improvements.

Thanks!

Note: Should fix https://github.com/python-poetry/poetry/issues/1522.

# Pull Request Check List

### DIFF
--- a/poetry/core/packages/dependency.py
+++ b/poetry/core/packages/dependency.py
@@ -159,7 +159,10 @@ class Dependency(object):
             if self.constraint.excludes_single_version():
                 requirement += " ({})".format(str(self.constraint))
             else:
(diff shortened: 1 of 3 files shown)


finswimmer (approved):
Good job :+1: 

Thanks a lot.

PR #80, opened 2020-09-12
Added trusted repository option

maayanbar13:
Upstream fix for [https://github.com/python-poetry/poetry/pull/2912](https://github.com/python-poetry/poetry/pull/2912).

Resolves upstream issue [https://github.com/python-poetry/poetry/issues/1556](https://github.com/python-poetry/poetry/issues/1556).

# Pull Request Check List
* [ ]  ~Updated **documentation** for changed code.~
* [ ] ~Added **tests** for changed code.~

Both not applicable, docs and tests are added in upstream PR.


poetry/core/json/schemas/poetry-schema.json
```diff
@@ -559,6 +559,10 @@
                 "secondary": {
                     "type": "boolean",
                     "description": "Declare this repository as secondary, i.e. it will only be looked up last for packages."
+                },
+                "trusted": {
+                    "type": "boolean",
+                    "description": "Declare this repository as trusted, i.e. option --trusted-host will be used when installing packages with pip."
```
abn, line 565: Lets drop mention of pip here since we longer use pip to download wheels and this needs to be passed into the executor.

abn (changes requested):

poetry/core/json/schemas/poetry-schema.json
```diff
@@ -559,6 +559,10 @@
                 "secondary": {
                     "type": "boolean",
                     "description": "Declare this repository as secondary, i.e. it will only be looked up last for packages."
+                },
+                "trusted": {
+                    "type": "boolean",
+                    "description": "Declare this repository as trusted, i.e. option --trusted-host will be used when installing packages."
```
abn, line 565: Sorry, I should have been a bit more clear in my last comment. Since the `--trusted-host` option is more to do with `pip`, and is more for the cli. We should be explicit about what trusted means here. I am thinking we should either call it `tls-verify` or have it as a table `tls = { verify = false }`. Latter allows for further context options. 

This also makes me wonder if this belongs in the `pyproject.toml` or in poetry config similar to auth details. The rationale being, having it in the `toml` means it could become a consideration. Usability vs security I guess. :)

maayanbar13: While i understand the sentiment, i believe that all those who use poetry also use `pip` separately for some cases, all of them will know the `trusted-host` option by heart, and its a better idea to keep this naming consistent, especially when this is the exact argument passed to `pip` when this is turned on.     
Still, I've added the description from the `pip` reference to make it clearer.

abn: Just to be clear this will no longer be passed to pip. Because pip does not talk to the indices, the packages are discovered and downloaded by poetry itself. I reckon a documentation entry hinting that this is similar to `trusted-host` in `pip` should suffice.

maayanbar13: Updated docs to reflect that its only similar to `pip` rather then actually passed to pip.

tomgrin10:
@abn Is there a reason this is not getting merged?

absassi:
Why is the option added to the schema of `pyproject.toml`? How to validate TLS connection is already part of the user configuration (`certificates.<repo>.cert`), it does not make sense to define one option in each place. Also, some users might prefer to validate their connections, while other don't want it for all sorts of reasons.

At least, I think this deserves more discussion before merging (otherwise it might have to be changed again, causing yet more breaking changes regarding certificates, like when `certificates` config was created, breaking the previous configuration options).

maayanbar13:
> Why is the option added to the schema of `pyproject.toml`? How to validate TLS connection is already part of the user configuration (`certificates.<repo>.cert`), it does not make sense to define one option in each place.     

I don't agree that this option defines  _how_ to validate the TLS connection but rather _if_.
It woudn't make a lot of sense to define that you don't want to use TLS inside a `.cert` file...
(And I think it would make sense to disallow using `HTTP` repositories unless this option is enabled.)

> Also, some users might prefer to validate their connections, while other don't want it for all sorts of reasons.

While I can see why this might be an interesting feature, the main usage of this option is in scenarios where you use a self-hosted private repository and can't be bothered setting up certificates (usually in some sort of enterprise environment).    
In that case all the users will want to have the same option (TLS disabled).    

> At least, I think this deserves more discussion before merging (otherwise it might have to be changed again, causing yet more breaking changes regarding certificates, like when `certificates` config was created, breaking the previous configuration options).

I don't agree that this is a breaking change: this option field is optional, and could default to the current behavior.    
If, down the line, the decision will be made to make this option user-specific, it could be implemented in a way that overrides the current option instead of replacing it. (I.E both options available, with the user-specific one having priority.)

absassi:
> I don't agree that this option defines how to validate the TLS connection but rather if.
> It woudn't make a lot of sense to define that you don't want to use TLS inside a .cert file...
> (And I think it would make sense to disallow using HTTP repositories unless this option is enabled.)

@maayanbar13 sorry if I wasn't clear, I don't mean to define it in the certificate file, but in a config option like `certificates.<repo>.trusted = true`. This way, all options related to TLS validation (i.e., the option of _if_ and the option of _how_) are grouped together.

It is also makes more sense to me there than in pyproject.toml for privately hosted repos, because it is something that is related to the repository, not the project.

Obviously, having both options available would be fine too.

maayanbar13:
> Obviously, having both options available would be fine too.

Sure, that makes sense. 
I've implemented said feature in the [downstream PR](https://github.com/python-poetry/poetry/pull/2912), and its pending the original author to merge it.  

That feature is unrelated to the `poetry-core` repo, so I think PR should be merged.  
Waiting on @abn 's response :)

maayanbar13:
@abn @absassi Is there any reason this is not getting merged?

mathbou:
Is there any reason this PR is still pending after all this time?

therewillbeblood:
@maayanbar13 can you please solve merge conflicts and push again ?

maayanbar13:
@therewillbeblood Sure :)

therewillbeblood:
@abn it seems you are assigne d as a reviewer, would you have an ETA to share to us when you could do the reivew ?

abn:
First off; apologies for not getting to this earlier.

Second, I think (along the lines of what @absassi was suggesting) that this should be a poetry level option for repository that ends up being configured as `poetry config certificates.foo.verify False`. Not something that lands in `pyproject.toml` but rather user specfic config for hte repository. The reason for this is partly consistency, and partly security. Wheter or not a host's certificate is verified is upto the end user and not the project, you do not want the insecure behaviour to be default across everyone.

For the above reason, I am closing this issue. But will review any downstream PR for the new `verify` option alongside current [certificate configurations](https://python-poetry.org/docs/repositories/#custom-certificate-authority-and-mutual-tls-authentication).

PR #81, opened 2020-09-15
Fix: Exclude file from wheel when listed in gitignore and src layout is used

finswimmer:
In case a `src`-layout is used, excluding files via `.gitignore` currently fails for build wheels.

This is due to the fact that `vcs.get_ignored_files()` gives the relative path to the project folder, but the current file is given relative to the package folder.

This PR fixes this by make the comparison relative to the project folder.

Resolves: https://github.com/python-poetry/poetry/issues/2910


poetry/core/masonry/builders/wheel.py
```diff
@@ -181,7 +181,7 @@ def _write_metadata(self, wheel):
                 relative_path = "%s/%s" % (self.dist_info, path.relative_to(self._path))
                 self._add_file(wheel, path, relative_path)
             else:
-                logger.debug("Skipping: %s", path.as_posix())
+                logger.debug("Skipping: %s" % path.as_posix())
```
abn, line 184: For logging it sould be `,`, this should defer evaluation to only when it is required.

poetry/core/masonry/builders/builder.py
```diff
@@ -161,22 +161,20 @@ def find_files_to_add(
...
-
-                include_file = BuildIncludeFile(path=file, source_root=source_root)
+                include_file = BuildIncludeFile(path=file, source_root=self._path)
 
                 if self.is_excluded(
                     include_file.relative_to_source_root()
                 ) and isinstance(include, PackageInclude):
                     continue
 
+                if (
```
abn, line 171: Would a better fix here be to fix `is_excluded` to support `BuildIncludeFile` and detect that it is not in the project root? I just feel uncomfortable that we are updating the include file instance property. We should really make that immutable.

abn (changes requested):

poetry/core/masonry/builders/builder.py
finswimmer: Passing `BuildIncludeFile` to `is_excluded` doesn't work here, because in other parts of the code we need to pass a `Path`. But the solutions seems to be simpler. The `Builder` knows the path to the project root, so we can just drop it in here. For readability reasons I've added the project root as a new attribute to `BuildIncludeFile` and added a method `relative_to_project_root`, which we can use here.

poetry/core/masonry/builders/wheel.py
finswimmer: Hm, I found this code accidentally, because my console printed out `Skipping: %s`. I've  replaced this now with `.fortmat(...)`

poetry/core/masonry/builders/wheel.py
```diff
@@ -181,7 +181,7 @@ def _write_metadata(self, wheel):
                 relative_path = "%s/%s" % (self.dist_info, path.relative_to(self._path))
                 self._add_file(wheel, path, relative_path)
             else:
-                logger.debug("Skipping: %s", path.as_posix())
+                logger.debug("Skipping: {}".format(path.as_posix()))
```
abn, line 184: https://docs.python.org/3/howto/logging.html#logging-variable-data 

I am not sure why the logger printed out the `%s` literally for you. Something else going on? Either way this change should not be required.

poetry/core/masonry/builders/builder.py
```diff
@@ -170,10 +172,12 @@ def find_files_to_add(
...
                     source_root = self._path
 
-                include_file = BuildIncludeFile(path=file, source_root=source_root)
+                include_file = BuildIncludeFile(
+                    path=file, project_root=self._original_path, source_root=source_root
+                )
 
                 if self.is_excluded(
-                    include_file.relative_to_source_root()
+                    include_file.relative_to_project_root()
```
abn, line 180: I like this appraoch.

poetry/core/masonry/builders/builder.py
```diff
@@ -191,7 +195,9 @@ def find_files_to_add(
         if self._package.build_script and not exclude_build:
             to_add.add(
                 BuildIncludeFile(
-                    path=self._package.build_script, source_root=self._path
+                    path=self._package.build_script,
+                    project_root=self._original_path,
+                    source_root=self._path,
```
abn, line 200: Is original path and path ever different?

poetry/core/masonry/builders/builder.py
```diff
@@ -332,6 +341,9 @@ def __hash__(self):
     def __repr__(self):  # type: () -> str
         return str(self.path)
 
+    def relative_to_project_root(self):  # type(): -> Path
+        return self.path.relative_to(self.project_root)
```
abn, line 345: Should we care about absolute path includes? These might throw an exception here since we do not explicitly disallow them yet?

finswimmer: Absolute path includes are not supported yet, correct? So if someone has an absolute path in the `include` they get problems anyway :)

poetry/core/masonry/builders/builder.py
finswimmer: At least it was the case for the `CompleteBuilder` which we have removed some days/weeks ago. There the `sdist` was unpacked somewhere else and for checking what's excluded by git, we needed to know where the original path was. I can remove it, but not sure if we get problems anywhere else :thinking:

poetry/core/masonry/builders/wheel.py
finswimmer: I couldn't figure out what's going on here. The more strange thing is, that the debug messages with "Adding: " are prepended by a `-` automatically but for "Skipping" not. As soon as I change "Adding:" to something different, the `-` disappears as well :thinking: I think we shouldn't stress that point in this PR.

poetry/core/masonry/builders/builder.py
finswimmer: Tests passes when I remove `self._original_path` and replace it in the code with `self._path`

poetry/core/masonry/builders/wheel.py
abn: Can you raise this in another issue please? This mostlikely is some wierd config somewhere.

abn (approved):

PR #89, opened 2020-09-26
vendoring: split sync and update make targets

abn:
The sync target previously regenerated the poetry.lock for vendored
packages, causing larger than necessary change set for patches. This
change introduces the update target to specifically handle lock file
updates, lock target to regenerate the lock file and repurpose the sync
target to only sync as expected.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,24 +3,27 @@ ROOT_DIR                        := $(patsubst %/,%,$(dir $(MAKEFILE_PATH)))
 
 VENDOR_SRC                      := $(ROOT_DIR)/vendors
 VENDOR_DIR                      := $(ROOT_DIR)/poetry/core/_vendor
+VENDOR_TXT                      := $(VENDOR_DIR)/vendor.txt
 POETRY_BIN                      ?= $(shell which poetry)
 
-
 .PHONY: vendor/lock
-vendor/lock:
-    # regenerate lock file
-	@pushd $(VENDOR_SRC) && $(POETRY_BIN) lock
-
-	# regenerate vendor.txt file (exported from lockfile)
-	@pushd $(VENDOR_SRC) && $(POETRY_BIN) export --without-hashes \
-			| egrep -v "(importlib|zipp)" \
-			| sort > $(VENDOR_DIR)/vendor.txt
-
+vendor/lock: $(VENDOR_LOCK)
+	# regenerate lock file
+	@pushd $(VENDOR_SRC) && $(POETRY_BIN) lock --no-update
 
 .PHONY: vendor/sync
-vendor/sync: | vendor/lock
+vendor/sync:
+	# regenerate vendor.txt file (exported from lockfile)
+	@pushd $(VENDOR_SRC) && $(POETRY_BIN) export --without-hashes 2> /dev/null \
+			| egrep -v "(importlib|zipp)" \
+			| sort > $(VENDOR_TXT)
+
 	# vendor packages
 	@vendoring sync
 
 	# strip out *.pyi stubs
 	@find "$(VENDOR_DIR)" -type f -name "*.pyi" -exec rm {} \;
+
+.PHONY: vendor/update
+vendor/update: | vendor/lock vendor/sync
+	@:
--- a/poetry.lock
+++ b/poetry.lock
@@ -61,11 +61,11 @@ python-versions = "*"
 
 [[package]]
 name = "cfgv"
-version = "3.0.0"
+version = "3.2.0"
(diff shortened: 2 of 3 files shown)


sinoroc:
I modified `vendors/pyproject.toml` as following:

```
$ git diff vendors/pyproject.toml 
diff --git a/vendors/pyproject.toml b/vendors/pyproject.toml
index 69ccbd0..75d101c 100644
--- a/vendors/pyproject.toml
+++ b/vendors/pyproject.toml
@@ -22,6 +22,6 @@ classifiers = [
 python = "^3.6"
 
 jsonschema = "^3.2.0"
-lark-parser = "^0.9.0"
+lark-parser = { git = "https://github.com/MegaIng/lark.git", branch = "custom_import_sources" }
 packaging = "^20.1"
 tomlkit = ">=0.7.0,<1.0.0"
```

Then I ran `make vendor/sync`:

```
$ make SHELL=bash vendor/sync 
# regenerate vendor.txt file (exported from lockfile)
~/workspace/poetry-core/vendors ~/workspace/poetry-core
# vendor packages
Load configuration... Done!
Clean existing libraries... Done!
Add vendored libraries...  
  Running pip install -t poetry/core/_vendor -r poetry/core/_vendor/vendor.txt --no-compile --no-deps
    ERROR: Invalid requirement: 'Warning: The lock file is not up to date with the latest changes in pyproject.toml. You may be getting outdated dependencies. Run update to update them.' (from line 9 of poetry/core/_vendor/vendor.txt)
  Command exited with non-zero exit code: 1
# strip out *.pyi stubs
```

Indeed the following has been generated in `poetry/core/_vendor/vendor.txt`
```
$ git diff poetry/core/_vendor/vendor.txt 
diff --git a/poetry/core/_vendor/vendor.txt b/poetry/core/_vendor/vendor.txt
index a1c24ba..b6a594b 100644
--- a/poetry/core/_vendor/vendor.txt
+++ b/poetry/core/_vendor/vendor.txt
@@ -1,8 +1,9 @@
 attrs==19.3.0
 jsonschema==3.2.0
-lark-parser==0.9.0
+lark-parser @ git+https://github.com/MegaIng/lark.git@custom_import_sources
 packaging==20.4
 pyparsing==2.4.7
 pyrsistent==0.16.0
 six==1.15.0
 tomlkit==0.7.0
+Warning: The lock file is not up to date with the latest changes in pyproject.toml. You may be getting outdated dependencies. Run update to update them.
```

Makefile
```diff
@@ -3,24 +3,27 @@ ROOT_DIR                        := $(patsubst %/,%,$(dir $(MAKEFILE_PATH)))
...
+VENDOR_TXT                      := $(VENDOR_DIR)/vendor.txt
 POETRY_BIN                      ?= $(shell which poetry)
 
-
 .PHONY: vendor/lock
-vendor/lock:
-    # regenerate lock file
+vendor/lock: $(VENDOR_LOCK)
+	# regenerate lock file
 	@pushd $(VENDOR_SRC) && $(POETRY_BIN) lock
```
sinoroc, line 12: Would `$(POETRY_BIN) lock --no-update` make sense here? (if and when it becomes available, of course)
https://github.com/python-poetry/poetry/pull/3034

Makefile
```diff
@@ -3,24 +3,27 @@ ROOT_DIR                        := $(patsubst %/,%,$(dir $(MAKEFILE_PATH)))
...
 
+.PHONY: vendor/sync
+vendor/sync:
 	# regenerate vendor.txt file (exported from lockfile)
 	@pushd $(VENDOR_SRC) && $(POETRY_BIN) export --without-hashes \
```
sinoroc, line 17: Maybe we should pay attention to not catch too much output here, so maybe redirect `stderr` or somehow figure out how to take advantage of the `--output` flag, or maybe simply the `-q` flag.

abn: We can use the `-o` flag here

Makefile
abn: I guess the question is if this is done after a change is made to the pypt or we are simply reloacking I guess.

sinoroc: Not sure what I need, what makes sense. I was a bit lost in this corner. Anyway it didn't seem to make my use case easy. My use case is: I want to change only 1 vendored dependency. And the things I had tried would trigger changes for more than 1 vendored dependency. I should give it another go some day so that I can give better comments here.

Makefile
sinoroc: I guess that would work, we would still need to exclude the lines that we do not need out of the file.

stephsamson (approved):

PR #101, opened 2020-10-14
remove dev-dependencies resolution from the build

roxchkplusony:
Relates-to: python-poetry/poetry#2174

### DIFF
--- a/poetry/core/factory.py
+++ b/poetry/core/factory.py
@@ -27,7 +27,9 @@ class Factory(object):
     Factory class to create various elements needed by Poetry.
     """
 
-    def create_poetry(self, cwd=None, with_dev=True):  # type: (Optional[Path]. bool) -> Poetry
+    def create_poetry(
+        self, cwd=None, with_dev=True
+    ):  # type: (Optional[Path]. bool) -> Poetry
         poetry_file = self.locate(cwd)
         local_config = PyProjectTOML(path=poetry_file).poetry_config
 
--- a/poetry/core/factory.py
+++ b/poetry/core/factory.py
@@ -27,7 +27,7 @@ class Factory(object):
     Factory class to create various elements needed by Poetry.
     """
 
-    def create_poetry(self, cwd=None):  # type: (Optional[Path]) -> Poetry
+    def create_poetry(self, cwd=None, with_dev=True):  # type: (Optional[Path]. bool) -> Poetry
         poetry_file = self.locate(cwd)
         local_config = PyProjectTOML(path=poetry_file).poetry_config
 
@@ -91,7 +91,7 @@ class Factory(object):
                     self.create_dependency(name, constraint, root_dir=package.root_dir)
                 )
 
-        if "dev-dependencies" in local_config:
+        if with_dev and "dev-dependencies" in local_config:
             for name, constraint in local_config["dev-dependencies"].items():
                 if isinstance(constraint, list):
                     for _constraint in constraint:
--- a/poetry/core/masonry/api.py
+++ b/poetry/core/masonry/api.py
@@ -31,7 +31,7 @@ get_requires_for_build_sdist = get_requires_for_build_wheel
 
 
 def prepare_metadata_for_build_wheel(metadata_directory, config_settings=None):
-    poetry = Factory().create_poetry(Path(".").resolve())
+    poetry = Factory().create_poetry(Path(".").resolve(), with_dev=False)
     builder = WheelBuilder(poetry)
 
     dist_info = Path(metadata_directory, builder.dist_info)
@@ -52,14 +52,14 @@ def prepare_metadata_for_build_wheel(metadata_directory, config_settings=None):
 
 def build_wheel(wheel_directory, config_settings=None, metadata_directory=None):
     """Builds a wheel, places it in wheel_directory"""
-    poetry = Factory().create_poetry(Path(".").resolve())
+    poetry = Factory().create_poetry(Path(".").resolve(), with_dev=False)
 
     return unicode(WheelBuilder.make_in(poetry, Path(wheel_directory)))
 
 
 def build_sdist(sdist_directory, config_settings=None):
     """Builds an sdist, places it in sdist_directory"""
-    poetry = Factory().create_poetry(Path(".").resolve())
+    poetry = Factory().create_poetry(Path(".").resolve(), with_dev=False)
 
     path = SdistBuilder(poetry).build(Path(sdist_directory))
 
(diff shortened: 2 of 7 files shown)


tests/fixtures/project_with_invalid_dev_deps/pyproject.toml
```diff
@@ -0,0 +1,13 @@
+[tool.poetry]
```
abn, line 1: I would move this to the builders fixtures as this needs to be tested with api not with factory. Maybe name it `with_invalid_path_dep_dev` and add package to the source too so we can evaluate the built wheel.

Additionally, I would also recommend that we add a fixture for `with_invalid_path_dep_main` that can be used to ensure an error is raised if encountered.

poetry/core/masonry/api.py
```diff
@@ -31,7 +31,7 @@ def get_requires_for_build_wheel(config_settings=None):
 
 
 def prepare_metadata_for_build_wheel(metadata_directory, config_settings=None):
-    poetry = Factory().create_poetry(Path(".").resolve())
+    poetry = Factory().create_poetry(Path(".").resolve(), with_dev=False)
```
abn, line 34: I do not think these tests need to change.

poetry/core/factory.py
```diff
@@ -27,7 +27,7 @@ class Factory(object):
     Factory class to create various elements needed by Poetry.
     """
 
-    def create_poetry(self, cwd=None):  # type: (Optional[Path]) -> Poetry
+    def create_poetry(self, cwd=None, with_dev=True):  # type: (Optional[Path]) -> Poetry
```
abn, line 30:
```suggestion
    def create_poetry(self, cwd=None, with_dev=True):  # type: (Optional[Path]. bool) -> Poetry
```

tests/test_factory.py
```diff
@@ -199,3 +199,16 @@ def test_create_poetry_fails_on_invalid_configuration():
   - 'description' is a required property
 """
     assert expected == str(e.value)
+
+
+def test_create_poetry_with_invalid_dev_dependencies():
```
abn, line 204: The test cases here should be focussed on the flag `with_dev` directly, ie. re-use exostomg fixtured to check if development dependencies were loaded and not loaded if `True` and `False` respecively.

poetry/core/masonry/api.py
```diff
@@ -52,14 +52,14 @@ def prepare_metadata_for_build_wheel(metadata_directory, config_settings=None):
 
 def build_wheel(wheel_directory, config_settings=None, metadata_directory=None):
     """Builds a wheel, places it in wheel_directory"""
-    poetry = Factory().create_poetry(Path(".").resolve())
+    poetry = Factory().create_poetry(Path(".").resolve(), with_dev=False)
```
abn, line 55: The verification of this shouldbe done by duplcating the existing test with the new fixtures added here. File content verification is not required, but just need to verify that a) wheel buidls when invalid dev dependency encountered and b) wheel builds will fail if these are encountered in the main category.

https://github.com/python-poetry/poetry-core/blob/0db07bc81b989c5c4933cf019fec7a4f93fbd744/tests/masonry/test_api.py#L46-L54

It might also be great to add an integration test case if easy enough. See the integration test suite at https://github.com/python-poetry/poetry-core/blob/master/tests/integration/test_pep517.py.

abn (changes requested):

tests/fixtures/project_with_invalid_dev_deps/pyproject.toml
roxchkplusony: Do you believe strongly that I should ONLY test in test_api and not in test_factory at all? I'll remove those tests if you don't want them, but I think they're doing something slightly different, something a little less e2e and applicative but more focused on the module itself.

tests/test_factory.py
roxchkplusony: Tests added for that.

poetry/core/masonry/api.py
roxchkplusony: I think for edge cases not meant to affect the build otherwise, it should stay in the lower level and not in the integration test. Let me know if you feel strongly about that though.

poetry/core/masonry/api.py
roxchkplusony: Because this isn't test code I'm a little confused what you mean. Could you elaborate?

abn: Ignore me.

tests/fixtures/project_with_invalid_dev_deps/pyproject.toml
abn: I believe you are testing the wrong things at the wrong level. The change is to ensure that when called via the PEP 517 api we ignore the dev dependnecies. The change in the factory code is to enable/disable inclusion of development dependencies and hence that is what should be tested, not if the "invalid" dependencies are not raising exceptions. This is what I have indicated in my other comment.

The suggestion to move the fixture is more for consistency as that is where the builder related fixtures are at. For verifying the flag, you can simply use an existing fixture here with both main and dev dependencies and parameterised test validating both flag states.

poetry/core/masonry/api.py
abn: This is still a case, not an edge case per say. We are verifying that packages with invalid or non-existent dev dependnecies can still be built. This is not an edge case. 

Integration testing (`pip` or `pep517` call) is option for this, but api (masonry) level testing is mandatory.

tests/fixtures/project_with_invalid_dev_deps/pyproject.toml
roxchkplusony: @abn I wrote new tests in test_api that I believe does what you are asking for. I also updated the test_factory tests which I believe also does what you are asking for. Did you get a chance to look at the new tests? Is your opinion still unchanged?

abn: Looks about right at first glance will review tomorrow. Thanks for the changes.

abn (changes requested):
You seem to have inadvertantly moved `tests/json/__init__.py`. Additionally, it could be great if we could parametrise the tests reduce some duplication (optional).

Otherwise looks good.

roxchkplusony:
Added back the __init__.py file I had inadvertently deleted. I don't see a great deduplication opportunity with the two test files, but let me know if your optional becomes mandatory. Otherwise, I hope I can get an approve.

abn (approved):

orlevii:
Any plans to release this fix soon? 🙏

PR #104, opened 2020-10-24
fix incorrect type hints and add missing ones

finswimmer:
This PR adds type hints to all method arguments and return values (except tests).

I guess there is still improvement. So feel free to comment :)


sdispater (dismissed):
Looks good at first glance 👍 

I will go over them in more detail when I make the move to Python 3.6+ compatibility only.

sdispater (approved):

PR #107, opened 2020-11-03
Update VERSION_PATTERN

Yajing-Zhao:
The VERSION_PATTERN should be compatible with semantic version guideline. I replace the old pattern string with the official recommendation in this fix.

For example, 1.0.0-0.3.7 should be a valid version according to the official website of [semver](https://semver.org/), but it breaks with the current VERSION_PATTERN. This proposed fix comes from the official website so it should be the most reliable one.

Could someone help review this pull request? @sdispater Thanks!

### DIFF
--- a/poetry/core/version/version.py
+++ b/poetry/core/version/version.py
@@ -12,41 +12,9 @@ _Version = namedtuple("_Version", ["epoch", "release", "dev", "pre", "post", "lo
 
 
 VERSION_PATTERN = re.compile(
-    r"""
-    ^
-    v?
-    (?:
-        (?:(?P<epoch>[0-9]+)!)?                           # epoch
-        (?P<release>[0-9]+(?:\.[0-9]+)*)                  # release segment
-        (?P<pre>                                          # pre-release
-            [-_.]?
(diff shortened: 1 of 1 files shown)


abn:
@Yajing-Zhao the pattern is compliant to PEP 440 as required by [Python Package Core Metadata Specification](https://packaging.python.org/specifications/core-metadata/#id7). Hence, not intended to be 100% semver.

Yajing-Zhao:
> @Yajing-Zhao the pattern is compliant to PEP 440 as required by [Python Package Core Metadata Specification](https://packaging.python.org/specifications/core-metadata/#version). Hence, not intended to be 100% semver.

Thanks, but in this [documentation](https://python-poetry.org/docs/cli/#version), "The new version should ideally be a valid semver string" is potentially confusing, as a valid semver string does not necessarily pass the test. Maybe change the documentation?

abn:
I guess the doc could use more clarity as we do encourage the use of semver compatible PEP 440 versions. But we do not mention that not all valid semver strings are PEP 440 compliant versions.

abn:
@Yajing-Zhao I am going to close this one. Feel free to create a new PR on poetry for the documentation changes. We appreciate you raising this by the way.

PR #108, opened 2020-11-16
Fix for including modules from different locations

jaharkes:
When we define multiple packages from different source locations,
Poetry currently only uses the last specified from= location.
This patch adds explicit paths to package_dir for additional packages.

This fixes python-poetry/poetry#1811, fixes python-poetry/poetry#2354,
and possibly even python-poetry/poetry#2450.

- [N/A] Updated **documentation** for changed code.

### DIFF
--- a/tests/masonry/builders/test_sdist.py
+++ b/tests/masonry/builders/test_sdist.py
@@ -644,8 +644,4 @@ def test_split_source():
     setup_ast.body = [n for n in setup_ast.body if isinstance(n, ast.Assign)]
     ns = {}
     exec(compile(setup_ast, filename="setup.py", mode="exec"), ns)
-    assert "moduleA" in ns["package_dir"] or "moduleB" in ns["package_dir"]
-    if "moduleA" in ns["package_dir"]:
-        assert ns["package_dir"]["moduleA"] == "libA/moduleA"
-    if "moduleB" in ns["package_dir"]:
-        assert ns["package_dir"]["moduleB"] == "libB/moduleB"
+    assert "module_a" in ns["package_dir"] or "module_b" in ns["package_dir"]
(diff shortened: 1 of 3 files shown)


finswimmer:
Thanks a lot for your contribution @jaharkes :+1: Could you please add a test for this fix?

jaharkes:
Thanks for looking at it, it looks like my test is hitting some case-sensitivity issue on windows. I guess I should have picked better names than 'moduleA'  and 'moduleB'.

tests/masonry/builders/test_sdist.py
```diff
@@ -587,3 +587,19 @@ def test_sdist_disable_setup_py():
...
+    builder = SdistBuilder(poetry)
+
+    # Check setup.py
+    setup = builder.build_setup()
+    setup_ast = ast.parse(setup)
+
+    setup_ast.body = [n for n in setup_ast.body if isinstance(n, ast.Assign)]
+    ns = {}
+    exec(compile(setup_ast, filename="setup.py", mode="exec"), ns)
+    assert "module_a" in ns["package_dir"] or "module_b" in ns["package_dir"]
```
layday, line 605: (Thanks for putting this together - I was having the same issue with two packages lying in different folders.)

Why is there an `or` here?  Shouldn't it always be that `ns["package_dir"].keys() == {"", "module_b"}`?  If not, that'd mean that the build is not reproducible given the same input, which isn't good.

jaharkes: Yes, I believe you are correct because the first entry in the list is module_a which will always be added as "".

layday: Well, you could just test for equality: `ns["package_dir"] == {"": "lib_a", "module_b": "lib_b/module_b"}`.

jaharkes:
@triage  rebased against the current master and successfully ran the tox tests locally.
It has been almost a year since I opened this PR, is there anything I can do to help move this along?

neersighted (approved):
This change looks quite straightforward, and indeed addresses the bug in question with good test coverage. I'm going to let the checks run and give it a merge if they're good.

Thanks for your review @layday -- I think you addressed all the concerns I had with the original version.

PR #109, opened 2020-11-17
Support nested extras specifications

alexifm:
Resolves: https://github.com/python-poetry/poetry/issues/3369

This introduces a way to use nested specifications of extras to avoid redundancy in packages listed in extras.


alexifm:
I raised this question for the main Poetry package.  I intend to introduce a PR over there for a separate component but for the build side, I wanted to submit this as a draft to see if something like this would make sense and be desired.

PR #116, opened 2020-11-30
Add build-system.requires entries to dev-dependencies

filmor:
Resolves: [python-poetry#3428](https://github.com/python-poetry/poetry/issues/3428)


sinoroc:
I am really not sure this is something that is really needed, not as long as custom build steps are not supported.

It's true there is unofficial, undocumented, partial support, but id doesn't count, does it?

If such a feature were to be added, I personally think it should be _opt-in_, it should be deactivated by default and could be activated by a user setting somewhere, somehow.

filmor:
This PR is very similar to https://github.com/python-poetry/poetry-core/pull/58, I didn't see that before.

PR #117, opened 2020-12-07
changed code to use default branch if none is specified.

Armaxxx:
Resolves: https://github.com/python-poetry/poetry/issues/3366

### DIFF
--- a/poetry/core/packages/vcs_dependency.py
+++ b/poetry/core/packages/vcs_dependency.py
@@ -99,7 +99,15 @@ class VCSDependency(Dependency):
 
         if self.extras:
             requirement += "[{}]".format(",".join(self.extras))
+        if self.branch == "":
+            if parsed_url.protocol is not None:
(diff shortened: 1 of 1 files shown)


sinoroc:
Fixes: https://github.com/python-poetry/poetry/issues/3366 ?

Armaxxx:
So, the tests are failing because they expect "master" to be the default branch, however by removing the branch from the url the default "main" branch will be pulled, however it is named.

finswimmer:
Hello @Armaxxx,

I just took a quick look at this. I think at least this method https://github.com/python-poetry/poetry-core/blob/abe9fe59fd7a0b181463461273c16af8a258a969/poetry/core/packages/vcs_dependency.py#L97 must be adopted, to not add a branch if there is no one defined.

fin swimmer

Armaxxx:
You're right, I completely missed it, adapted.

stephsamson:
Closing in favor of #143. Thank you @Armaxxx.

PR #118, opened 2020-12-12
Add support to declare multiple README files

wagnerluis1982:
Resolves: python-poetry/poetry#873

- [x] Updated **documentation** for changed code (python-poetry/poetry#4743).

---

This code reuses the `"readme"` entry in a way to allow the user to declare in the old way or the new way. The two following declarations become valid:

**Single file**
```toml
readme = "README.rst"
```

**Multiple files**
```toml
readme = [
    "README.rst",
    "HISTORY.rst"
]
```

If the user declares files in different formats, the strict validation will issue.


sinoroc:
> Where should I add documentation? Should I open another MR in the main project for that?

I think so, yes.

wagnerluis1982:
I rebased my branch to latest `master` and solved conflicts. Also added documentation to python-poetry/poetry#4743

poetry/core/factory.py
```diff
@@ -94,7 +97,12 @@ def configure_package(
         package.classifiers = config.get("classifiers", [])
 
         if "readme" in config:
-            package.readme = root / config["readme"]
+            if isinstance(config["readme"], (list, tuple)):
```
neersighted, line 100:
```suggestion
            if isinstance(config["readme"], Sequence):
```

I would much prefer to see `Sequence` from `collections.abc` here.

poetry/core/factory.py
```diff
@@ -419,6 +427,15 @@ def validate(cls, config: dict, strict: bool = False) -> Dict[str, List[str]]:
                                 )
                             )
 
+            # Checking readme file types (must match)
+            if "readme" in config and isinstance(config["readme"], (list, tuple)):
```
neersighted, line 431: Same as before.

```suggestion
            if "readme" in config and isinstance(config["readme"], Sequence):
```

poetry/core/factory.py
```diff
@@ -419,6 +427,15 @@ def validate(cls, config: dict, strict: bool = False) -> Dict[str, List[str]]:
...
+                readme_types = {content_type(readme) for readme in config["readme"]}
+                if len(readme_types) > 1:
+                    result["errors"] += [
+                        "Declared README files must be of same type: found %s"
```
neersighted, line 435: We don't use `%`-formatting anymore. Use an f-string here, and use string.join to comma-separate your list of types.

poetry/core/json/schemas/poetry-schema.json
```diff
@@ -55,8 +55,19 @@
       "$ref": "#/definitions/maintainers"
     },
     "readme": {
-      "type": "string",
-      "description": "The path to the README file"
+      "anyOf": [
+        {
+          "type": "string",
+          "description": "The path to the README file"
```
neersighted, line 61: Descriptions here should be sentences. For some reason the existing description was not -- add some periods and we should be good.

poetry/core/masonry/metadata.py
```diff
@@ -52,8 +52,11 @@ def from_package(cls, package: "Package") -> "Metadata":
         meta.version = normalize_version(package.version.text)
         meta.summary = package.description
         if package.readme:
-            with package.readme.open(encoding="utf-8") as f:
-                meta.description = f.read()
+            description = []
+            for readme in package.readme:
+                with readme.open(encoding="utf-8") as f:
+                    description += [f.read()]
```
neersighted, line 58:
```suggestion
                    description.append(f.read())
```

The allocation of another list is unnecessary/wasteful here -- I think `.append()` is cleaner to read.

poetry/core/packages/package.py
```diff
@@ -88,7 +88,8 @@ def __init__(
         self.documentation_url = None
         self.keywords = []
         self._license = None
-        self.readme = None
+        self.readme = ()
+        self.readme_type = None
```
neersighted, line 92: Since the description is a sum of READMEs, I'd probably call this the `description_type` instead.

poetry/core/packages/package.py
```diff
@@ -88,7 +88,8 @@ def __init__(
         self.documentation_url = None
         self.keywords = []
         self._license = None
-        self.readme = None
+        self.readme = ()
```
neersighted, line 91: Given the fact that this is now a tuple, we should pluralize it to make clear what is going on.

poetry/core/utils/helpers.py
```diff
@@ -104,3 +104,13 @@ def merge_dicts(d1: dict, d2: dict) -> None:
             merge_dicts(d1[k], d2[k])
         else:
             d1[k] = d2[k]
+
+
+def content_type(path):
```
neersighted, line 109: Since this is totally specific to the README/description functionality, let's keep it in the original file. A `@staticmethod` `_readme_content_type` will do the trick.

Also, please add type hints to this method.

tests/masonry/builders/fixtures/with_readme_files/pyproject.toml
```diff
@@ -0,0 +1,19 @@
+[tool.poetry]
```
neersighted, line 1: I'd rather share/re-use these fixtures (and have one single, complete version) rather than duplicate two partial versions.

tests/masonry/builders/test_builder.py
```diff
@@ -260,3 +260,17 @@ def test_builder_convert_script_files(fixture, result):
...
+
+def test_metadata_with_readme_files():
+    test_path = Path(__file__).parent / "fixtures" / "with_readme_files"
+    builder = Builder(Factory().create_poetry(test_path))
+
+    metadata = Parser().parsestr(builder.get_metadata_content())
+
+    readme1 = test_path / "README-1.rst"
+    readme2 = test_path / "README-2.rst"
+    assert metadata.get_payload() == "%s\n%s\n" % (
```
neersighted, line 273: Let's avoid `%`-formatting here -- instead, I would suggest a simple string.join.

tests/test_factory.py
```diff
@@ -185,6 +185,13 @@ def test_validate_fails():
     assert Factory.validate(content) == {"errors": [expected], "warnings": []}
 
 
+def test_validate_multiple_readme_files():
```
neersighted, line 188: I would also like to see a test here for mixed README formats -- in order to exercise your error check above.

neersighted (changes requested):

poetry/core/factory.py
wagnerluis1982: I can't do that! Because `str` is also a `Sequence`. I elaborate below.

To not break compatibility, I need to handle `type(readme) == 'str'` to wrap the single readme file in a single `list`/`tuple`.

With the changes, both `readme` declarations are possible now:
```toml
[tool.poetry]
# str syntax (keep compatibility)
readme = "README.md"
# list syntax (extension)
readme = [
    "README.md"
]
```

poetry/core/factory.py
wagnerluis1982: Same as before

poetry/core/factory.py
wagnerluis1982: A little suggestion: I noticed TOML parser always returns a `list`-like object, so I could make the code more readable by simply do
```python
if isinstance(config["readme"], list):
```

neersighted: I think I'd go the other direction, and check for `str` since Python 3 does not have multiple string types.

wagnerluis1982 (commented):
Thanks for the awesome suggestions.

I added my fixes to the code: please note that I rebased the branch to reduce the number of commits.

Comments about my fixes:
- I chose `isinstance(config["readme"], list)` as the TOML parser only delivers `list`-like objects, no need to check for `tuple`.
- About using a `Sequence`, please check my separate comment
- Pluralized `Package.readme` (became `readmes`) and `readme_type` => `description_type`
- Added a test for validation fail when readme types mismatch
- Refactoring according to the comments
- Fixed a few glitches I found on the way

poetry/core/factory.py
wagnerluis1982: What do you mean? To invert the `if` to check if is a `str` first?

wagnerluis1982: I see what you meant now, it was late night, I needed a sleep :smile:

neersighted (dismissed):

neersighted (approved):

PR #119, opened 2020-12-15
Simplify temporary_directory and use safe_rmtree

thejcannon:
Resolves: https://github.com/python-poetry/poetry/issues/3493

Unit test fails without changes and passes with changes 🎉

### DIFF
--- a/tests/utils/test_helpers.py
+++ b/tests/utils/test_helpers.py
@@ -1,5 +1,6 @@
 import os
 import pathlib
+
 from stat import S_IREAD
 
 import pytest
--- a/tests/utils/test_helpers.py
+++ b/tests/utils/test_helpers.py
@@ -1,5 +1,4 @@
 import os
-import pathlib
 
 from stat import S_IREAD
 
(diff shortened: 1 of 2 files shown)


tests/utils/test_helpers.py
```diff
@@ -62,3 +67,11 @@ def test_parse_requires():
 )
 def test_utils_helpers_canonical_names(raw):
     assert canonicalize_name(raw) == "a-b-c"
+
+
+def test_utils_helpers_temporary_directory_readonly_file():
```
finswimmer, line 72: Here a check whether the folder is removed successfully is missing, is it?

finswimmer (changes requested):
Thanks a lot :+1: 

Just a minor thing.

tests/utils/test_helpers.py
thejcannon: The bug itself would throw `PermissionError` on the removal attempt. So in that regard the test shows we no longer run into that issue.
That being said checking that it _actually_ was removed is probably also a good idea 😄

thejcannon: Alrighty, added some asserts. 👍

finswimmer (approved):
Thanks a lot and happy new year :rocket:

Bettdatt:
Yes

PR #122, opened 2021-01-07
Always close the wheel tempfile after writing to it

kurtmckee:
Resolves: python-poetry/poetry#3545

As mentioned in python-poetry/poetry#3545, PyPy3 on Windows does not auto-close the tempfile that is created when building wheels. It appears that CPython auto-closes the file as the zip file context manager exits, but PyPy3 does not, leading to a PermissionError.

There are two commits in this patch:

1. The first demonstrates the issue. In fact, by maintaining a reference to the file object, it stays open in CPython and generates the exact same PermissionError as in PyPy3.
2. The second closes the issue by moving the bare `os.fdopen()` call into a context manager.

Please let me know if there is anything else that needs to be done, or if you want something accomplished in a different way. Resolving this issue will allow me to run feedparser unit tests on Windows with PyPy3.

### DIFF
--- a/poetry/core/masonry/builders/wheel.py
+++ b/poetry/core/masonry/builders/wheel.py
@@ -78,18 +78,19 @@ class WheelBuilder(Builder):
         new_mode = normalize_file_permissions(st_mode)
         os.chmod(temp_path, new_mode)
 
-        with zipfile.ZipFile(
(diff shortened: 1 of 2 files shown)


kurtmckee:
It looks like there are linting tests that are failing. I'll get this fixed immediately.

Could the linters be run with tox as well? This would let me resolve quality problems prior to pushing. :smile:

sdispater (approved):

PR #124, opened 2021-01-13
Fix #3285

audunska:
Resolves: python-poetry#3285

Fix python-poetry#3285 by removing explicitly included files from the set in `poetry.core.masonry.builders.builder.Builder.find_excluded_files`. If files are both included and excluded, they are usually excluded because of vcs, so it makes sense to override by explicitly including.

Added a test for this case.


sdispater (approved):
Looks good to me 👍

PR #126, opened 2021-01-21
Conceptual fix for issue #3589

brechtm:
This makes just enough changes to the code to make #3589 work. This is mostly intended to start discussion and collect feedback.

There are (at least) two outstanding issues:

1. with a `from = "src"` packages setup, this adds the files in _src_ twice. The resulting tar.gz does seem to be fine, however.
2. Performance is abysmal when using `include = [{ path = "**/*", format = "sdist" }]` in a sizeable project. I believe this is because poetry checks for excluded files on a file-per-file bases and doesn't take shortcuts when entire directories are excluded.

Summary of commits:
- Log all files being added
- Pass unresolved paths to is_excluded()
- 'exclude' should be able to override 'include' matches
- Some remarks
- Exclude .git directories and files in them by default

Partly resolves: python-poetry/poetry#3589


poetry/core/masonry/builders/builder.py
```diff
@@ -181,9 +184,7 @@ def find_files_to_add(
                     path=file, project_root=self._path, source_root=source_root
                 )
 
-                if self.is_excluded(
-                    include_file.relative_to_project_root()
-                ) and isinstance(include, PackageInclude):
+                if self.is_excluded(file.relative_to(source_root)):
```
brechtm, line 187: This changes include/exclude behavior, breaking three tests. Currently, include always trumps exclude (see also python-poetry/poetry#1336), which is too limiting for some use cases.

One solution, without introducing order-dependent include/exclude rules as in a a _MANIFEST.in_, is to give globbed paths a lower priority than fully-qualified paths. I feel this offers more flexibility while being intuitive at the same time. This would work for the examples given in python-poetry/poetry#1336, such as:
```ini
exclude = [
    "assets/sample/**/*"
]
include = [
    "assets/sample/subdir/one.yaml",
    "assets/sample/subdir/two.yaml"
]
```
But the following would now also work:
```ini
include = [
    "**/*"
]
exclude = [
    ".travis.yml"
]
```

PR #131, opened 2021-02-03
Modernize code base

sdispater:
This PR modernizes the code by doing the following:

- Remove the PYTHON2.7/3.5 compatibility layers
- Use proper type annotations


poetry/core/packages/package.py
```diff
@@ -9,28 +8,20 @@
...
+    from poetry.core.semver.helpers import VersionTypes  # noqa
+    from poetry.core.semver.version import Version  # noqa
+    from poetry.core.spdx.license import License  # noqa
     from poetry.core.version.markers import BaseMarker  # noqa
 
-    from .directory_dependency import DirectoryDependency
-    from .file_dependency import FileDependency
-    from .url_dependency import URLDependency
-    from .vcs_dependency import VCSDependency
+    from .types import DependencyTypes
```
finswimmer, line 24: I still think we need to keep `from .dependency import Dependency` in this file outside the `TYPE_CHECKING` only section. Otherwise this this will break any poetry version <1.2 installed via pip(x).

finswimmer (changes requested):

poetry/core/packages/package.py
sdispater: We will need to fix it on Poetry's side then, I think, rather than keeping it here.

sdispater: @finswimmer Also, the next bugfix release of Poetry should change the `poetry-core` constraint to `< 1.2` to ensure that we do not inadvertently pull a newer version of `poetry-core`.

sdispater: @finswimmer I made and merged the changes mentioned above on Poetry's side so we should be good to go for this PR.

sdispater:
@finswimmer Let me know if you still see something that needs changing.

sdispater:
I am gonna go ahead and merge this.

The import issue has already been fixed on Poetry's side (https://github.com/python-poetry/poetry/pull/3696) and the fix will be available in the next `1.1.5` release.

poetry/core/pyproject/__init__.py
```diff
@@ -1,6 +0,0 @@
-from poetry.core.pyproject.exceptions import PyProjectException
```
ns-admetrics, line 1: This removal breaks https://github.com/python-poetry/poetry/blob/6eaa2984d9fb51a837c4ab11643f4dd9d7828d4d/poetry/console/commands/config.py#L14 (current master).

nickray, line 1: This removal breaks https://github.com/python-poetry/poetry/blob/6eaa2984d9fb51a837c4ab11643f4dd9d7828d4d/poetry/console/commands/config.py#L14 (current master).

poetry/core/semver/__init__.py
```diff
@@ -1,151 +0,0 @@
...
-
-from .empty_constraint import EmptyConstraint
-from .exceptions import ParseConstraintError
-from .patterns import BASIC_CONSTRAINT
-from .patterns import CARET_CONSTRAINT
-from .patterns import TILDE_CONSTRAINT
-from .patterns import TILDE_PEP440_CONSTRAINT
-from .patterns import X_CONSTRAINT
-from .version import Version
-from .version_constraint import VersionConstraint
```
nickray, line 13: This removal breaks https://github.com/python-poetry/poetry/blob/6eaa2984d9fb51a837c4ab11643f4dd9d7828d4d/poetry/repositories/repository.py#L5 (current master).

nickray, line 13: This removal breaks https://github.com/python-poetry/poetry/blob/6eaa2984d9fb51a837c4ab11643f4dd9d7828d4d/poetry/repositories/repository.py#L5 (current master).

nickray:
Besides the type annotations and Python version housecleaning, multiple changes in this PR modify the public API of `poetry-core`. Technically, these are semver-breaking changes. Is this intended? Is there a corresponding PR to `poetry` dealing with the fallout already?

sdispater:
@nickray Thanks for sharing you concerns.

`poetry-core` is a PEP-517 build backend and is not indented to be used as a library – except by Poetry itself – for now. The only public API that matters at the moment are the PEP-517 hooks provided by `poetry.core.masonry.api`.

There is currently a work in progress branch (https://github.com/python-poetry/poetry/compare/modern-core) on Poetry's side to update the imports and code to account for these changes in `poetry-core`.

PR #136, opened 2021-02-11
Fix unicode error

julienmalard:
This fixes [#221](https://github.com/python-poetry/poetry/issues/221) by exchanging the regex package for re (the former having fixed the long-standing bug in re (see [here](https://github.com/lark-parser/lark/issues/590) for details). Not sure if this will work for poetry (i.e., whether adding regex as a dependency is acceptable)...but thought I would propose this change and wait for feedback to take it from there.
Thanks!

Resolves: https://github.com/python-poetry/poetry/issues/221

Unfortunately couldn't figure out where to add the tests. Let me know and I will gladly do so!


stephsamson:
Thank you for the PR! Can you rebase against main and resolve the conflicts?

julienmalard:
Thanks @stephsamson ! I rebased, added failing tests, and pushed again (all local tests passing with latest commit). Please do let me know what else would be needed.

julienmalard:
P.S. I just reformatted. It currently seems to fail because regex is missing as a dependency; do you happen to know where I should specify it? (Would adding it to `[tool.poetry.dependencies]` work?)

Secrus:
Superseded by #276

PR #137, opened 2021-03-01
Glob to find stale dist-infos

nickray:
This is a preparation for https://github.com/python-poetry/poetry/pull/3742.

It seems the sibling method is not documented/tested, so not clear what to do on that front.
Alternative might be to pass through `name` in `dist_info_name` unescaped if it's the `*`-glob?


abn:
Hi @nickray thank you for the change. Looking at the desired change, I am not sure if this needs to be here per-say. This can be, and imho should be, done poetry side without core changes.

abn:
Closing. See https://github.com/python-poetry/poetry/pull/3742#issuecomment-817270281.

PR #139, opened 2021-03-18
mypy: Fix Missing return statement/value errors

jcmdln:
I see that this repository doesn't yet include `mypy` as a dependency like `python-poetry/poetry` does, though using it I was able to find quite a few errors. By resolving these errors, I hope to make poetry-core _some amount_ more correct. For the purposes of this PR, I narrowed the scope to the following:

```sh
$ poetry run mypy poetry --exclude=poetry/core/_vendor | grep "Missing return statement\|Return value expected"
poetry/core/version/version.py:135: error: Missing return statement
poetry/core/version/version.py:149: error: Missing return statement
poetry/core/version/version.py:183: error: Missing return statement
poetry/core/semver/version.py:303: error: Return value expected
poetry/core/semver/version.py:307: error: Return value expected
poetry/core/semver/version.py:328: error: Return value expected
poetry/core/semver/version.py:334: error: Return value expected
```

Each error was rather innocuous, though easy enough to correct. Extra consideration was taken, possibly incorrectly, to ignore the `poetry/core/_vendor` directory as this looks like external code as the folder name suggests. I included this directory in a `mypy.ini` file as seen in the proposed commits, though left granular configuration alone for now pending broader discussion.

The GitHub Actions should confirm this as well, provided I haven't made a mistake, though the tests pass locally:

```sh
$ poetry run tox -e py39
...
== 529 passed, 2 skipped, 8 deselected in 2.96 seconds ==
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,7 +33,8 @@ MANIFEST.in
 /docs/site/*
 /tests/fixtures/simple_project/setup.py
 /tests/fixtures/project_with_extras/setup.py
-.mypy_cache
+.mypy_cache/
+__pycache__/
 
 .venv
 /releases/*
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,0 +1,2 @@
+[mypy]
+exclude = poetry/core/_vendor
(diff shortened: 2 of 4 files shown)


danieleades:
@jcmdln is this PR still relevant, or can it be closed?

jcmdln:
I still use (and prefer) poetry but I haven't been following the development closely, so I'm not sure that this PR is still relevant. For now I'll close this PR.

PR #140, opened 2021-03-19
Improve PEP-440 support for versions

abn:
This is a broad change aligning `poetry.core.semver.version.Version` implementation to be closer to PEP-440 that to semver. This adds the foundation for improved dev and local tag support in poetry managed version as well as fix ordering bugs that existed due to the differences between semantic versioning specification and PEP-440.

Further, previously used inline copy of `packaging.version.Version` implementation has now been removed in favour of
`poetry.core.version.pep440.PEP440Version` implementation.

**Note:** This change contains minor breaking changes that will require minimal changes downstream in `poetry`. (see python-poetry/poetry#3831)

**References:**
- https://www.python.org/dev/peps/pep-0440/#semantic-versioning

Relates-to: python-poetry/poetry#2543
Relates-to: python-poetry/poetry#1151
Relates-to: python-poetry/poetry#892


stephsamson (approved):

stephsamson (approved):

poetry/core/version/grammars/parser.py
```diff
@@ -1,26 +1,20 @@
...
-    def parse(self, string: str) -> "Tree":
-        from lark import Lark
-
-        if self._parser is None:
-            self._parser = Lark.open(
-                os.path.join(os.path.dirname(__file__), f"{self._grammar}.lark"),
-                parser="lalr",
-            )
-
-        return self._parser.parse(string)
```
sdispater, line 26: We need this class to lazily load grammers, otherwise the performance of Poetry will be severely degraded.

sdispater: Can you make the changes to add it back?

sdispater: Also, I don't think this is the right place to instantiate the parsers.

abn: Sure. I don't have a strong opinion on this. I was keeping the parser here since the original one was here. 

Happy to move to a factory model. That said, these will not be loaded until these are imported anyway, and they are imported when they are needed only. So in a way they are already lazy loaded and shouldn't have any change from the previous case.

shadycuz:
@abn Is there a way to make `poetry version` go back to outputing `alpha` instead of `a`?

This change broke my release process.

neersighted:
PEP 440 mandates alpha is normalized to `a` -- this is also not the right venue to discuss it. Please use GitHub Discussions or Discord if you have questions.

PR #141, opened 2021-03-21
Output valid file URLs when building requirements

AndrewGuenther:
Resolves: python-poetry/poetry#3818

This code isn't currently covered by a test, but I can add one. I also didn't see any relevant docs to update but can if someone could point me to them.

Similar to the issue fixed by python-poetry/poetry#3121, file/directory dependencies in requirements.txt format (and transitively, in wheels) must be in URL format. Running `pip install .` for any poetry project containing a directory dependency will currently fail due to this issue.

### DIFF
--- a/poetry/core/packages/directory_dependency.py
+++ b/poetry/core/packages/directory_dependency.py
@@ -122,8 +122,8 @@ class DirectoryDependency(Dependency):
         if self.extras:
             requirement += "[{}]".format(",".join(self.extras))
 
-        path_url = path_to_url(self._path)
-        requirement += " @ {}".format(path_url)
+        path = path_to_url(self.path) if self.path.is_absolute() else self.path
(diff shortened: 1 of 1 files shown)


poetry/core/packages/directory_dependency.py
```diff
@@ -121,7 +122,8 @@ def base_pep_508_name(self) -> str:
         if self.extras:
             requirement += "[{}]".format(",".join(self.extras))
 
-        requirement += " @ {}".format(self._path.as_posix())
+        path_url = path_to_url(self._path)
```
abn, line 125: For consistency's sake I would say we do exactly this > https://github.com/python-poetry/poetry-core/blob/master/poetry/core/packages/file_dependency.py#L110

abn (changes requested):

abn (approved):

PR #142, opened 2021-03-22
Use deterministic time for generated sdist files

abn:
When generating setup.py and PKG-INFO files, ensure that generated
files use a deterministic timestamp to enhance reproducibility of
source distributions.

Resolves: python-poetry/poetry#1102

### DIFF
--- a/poetry/core/masonry/builders/sdist.py
+++ b/poetry/core/masonry/builders/sdist.py
@@ -2,7 +2,6 @@ import logging
 import os
 import re
 import tarfile
-import time
 
 from collections import defaultdict
 from contextlib import contextmanager
@@ -97,14 +96,16 @@ class SdistBuilder(Builder):
                 setup = self.build_setup()
                 tar_info = tarfile.TarInfo(pjoin(tar_dir, "setup.py"))
                 tar_info.size = len(setup)
-                tar_info.mtime = time.time()
+                tar_info.mtime = 0
+                tar_info = self.clean_tarinfo(tar_info)
                 tar.addfile(tar_info, BytesIO(setup))
 
             pkg_info = self.build_pkg_info()
 
             tar_info = tarfile.TarInfo(pjoin(tar_dir, "PKG-INFO"))
             tar_info.size = len(pkg_info)
-            tar_info.mtime = time.time()
+            tar_info.mtime = 0
+            tar_info = self.clean_tarinfo(tar_info)
             tar.addfile(tar_info, BytesIO(pkg_info))
         finally:
             tar.close()
--- a/tests/masonry/builders/test_sdist.py
+++ b/tests/masonry/builders/test_sdist.py
@@ -1,5 +1,6 @@
 import ast
 import gzip
+import hashlib
 import shutil
 import tarfile
 
@@ -250,6 +251,24 @@ def test_package():
         assert "my-package-1.2.3/LICENSE" in tar.getnames()
 
 
+def test_sdist_reproducibility():
+    poetry = Factory().create_poetry(project("complete"))
+
+    hashes = set()
+
+    for _ in range(2):
+        builder = SdistBuilder(poetry)
+        builder.build()
+
+        sdist = fixtures_dir / "complete" / "dist" / "my-package-1.2.3.tar.gz"
+
+        assert sdist.exists()
+
+        hashes.add(hashlib.sha256(sdist.read_bytes()).hexdigest())
+
+    assert len(hashes) == 1
+
+
 def test_setup_py_context():
     poetry = Factory().create_poetry(project("complete"))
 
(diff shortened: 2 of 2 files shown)


Natureshadow:
I propose that this code should use `SOURCE_DATE_EPOCH` from environment if set.

cf. https://reproducible-builds.org/docs/source-date-epoch/

poetry/core/masonry/builders/sdist.py
```diff
@@ -97,14 +96,16 @@ def build(self, target_dir: Optional[Path] = None) -> Path:
                 setup = self.build_setup()
                 tar_info = tarfile.TarInfo(pjoin(tar_dir, "setup.py"))
                 tar_info.size = len(setup)
-                tar_info.mtime = time.time()
+                tar_info.mtime = 0
+                tar_info = self.clean_tarinfo(tar_info)
```
Natureshadow, line 100:
```suggestion
                if "SOURCE_DATE_EPOCH" in os.environ:
                    tar_info.mtime = int(os.environ.get("SOURCE_DATE_EPOCH"))
                    tar_info = self.clean_tarinfo(tar_info)
                else:
                    tar_info.mtime = time.time()
```

poetry/core/masonry/builders/sdist.py
```diff
@@ -97,14 +96,16 @@ def build(self, target_dir: Optional[Path] = None) -> Path:
...
                 tar.addfile(tar_info, BytesIO(setup))
 
             pkg_info = self.build_pkg_info()
 
             tar_info = tarfile.TarInfo(pjoin(tar_dir, "PKG-INFO"))
             tar_info.size = len(pkg_info)
-            tar_info.mtime = time.time()
+            tar_info.mtime = 0
+            tar_info = self.clean_tarinfo(tar_info)
```
Natureshadow, line 108:
```suggestion
            if "SOURCE_DATE_EPOCH" in os.environ:
                tar_info.mtime = int(os.environ.get("SOURCE_DATE_EPOCH"))
                tar_info = self.clean_tarinfo(tar_info)
            else:
                tar_info.mtime = time.time()
```

tests/masonry/builders/test_sdist.py
```diff
@@ -1,5 +1,6 @@
 import ast
 import gzip
+import hashlib
```
Natureshadow, line 3:
```suggestion
import hashlib
import os
```

tests/masonry/builders/test_sdist.py
```diff
@@ -250,6 +251,24 @@ def test_package():
         assert "my-package-1.2.3/LICENSE" in tar.getnames()
 
 
+def test_sdist_reproducibility():
+    poetry = Factory().create_poetry(project("complete"))
+
+    hashes = set()
```
Natureshadow, line 257:
```suggestion
    hashes = set()
    os.environ["SOURCE_DATE_EPOCH"] = "0"
```

Natureshadow (changes requested):
As suggested earlier, I would propose to use `SOURCE_DATE_EPOCH`.

Defaulting to `0` if it is not set might be ok, I am not sure of that. I somehow tend to leaving the current timestamp as default.

Natureshadow:
Oh, quoting from https://reproducible-builds.org/specs/source-date-epoch/ , it seems that honouring this variable is a **MUST`` to comply with the reproducible builds standard. I was not aware of this. As such, I make my proposal a bit harder ;).

It also talks about "timestamp clamping", which says that all timestamps that do not use the value from this variable must be earlier, so `0` looks good. It does, however, look wrong to me to use such a pseudo-random magic value. I would therefore stick with `time.time()` if the source epoch is not set.

Natureshadow:
> Oh, quoting from https://reproducible-builds.org/specs/source-date-epoch/ , it seems that honouring this variable is a **MUST** to comply with the reproducible builds standard. I was not aware of this. As such, I make my proposal a bit harder ;).

Had a chat with the people form reproducible-builds.org.

It seems there were a few misunderstnadings on my side:

* `SOURCE_DATE_EPOCH` is mainly for binary builds, to align it with the source timestamp
* `0` is perfectly fine for reproducibility
* Using a timestamp that somehow reflects the last time the source was modified would be good. One idea is looking at git's `HEAD` if available, another would be to copy the mtime of `pyproject.toml`.

So if the maintainers want to keep it simple, just sticking with 0 seems fine.

stephsamson (approved):

Natureshadow:
@stephsamson It seems you reviewed this PR. Did you take my comments into account?

stephsamson:
@Natureshadow for the purposes of this PR, I think that keeping `tar_info.mtime` to `0` is fine as you mentioned [here yourself](https://github.com/python-poetry/poetry-core/pull/142#issuecomment-804209538) or did I misunderstand? Also, `SOURCE_DATE_EPOCH` as you said is for binary builds so I find that that is out of scope for `sdist` and subsequently this PR and python-poetry#1102.

Natureshadow:
@stephsamson Yes, I did say that using `0` is fine as well. I just wanted to know your opinion, as you left my change proposals open.

stephsamson:
@Natureshadow I think that your suggestion, specifically for checking `mtime` of pyproject.toml to use as a timestamp for binary distributions, is a good candidate for a separate issue/refactoring.

PR #146, opened 2021-03-27
schema: add support for source.links boolean

abn:
Thsi change is required to add support for archive link source urls as project sources. This will enable the use of scenarios currently facilitated in `pip` by the use of `--find-links` flag.


stephsamson (commented):
Should we add a section in the Poetry documentation about links?

abn:
Yup; that will come when the feature is enabled in poetry. Right now it will just be allowed in the schema.

stephsamson (approved):

PR #147, opened 2021-03-27
masonry: auto-detect default stub packages

abn:
This change ensures that [stub-only packages](https://www.python.org/dev/peps/pep-0561/#stub-only-packages) are auto-detected when `packages` is not specified in the project's `pyproject.toml`. This avoids a common user error when creating a stub-only project managed by poetry. 

One draw back of this approach is that we might not be able to support packages that actually want `-stubs` in their names. Not sure if this is a good practice anyway.


Secrus:
I am going to close this. It's quite old by now and stub packages are less popular since more and more projects adopt native types.

PR #153, opened 2021-04-01
fix(packages/dependency): add space after filename for file dependencies with markers

ossareh:
Resolves: https://github.com/python-poetry/poetry/issues/3872

local vendored files need a space after the file name and before the ";" which demarks the start of
markers

### DIFF
--- a/tests/packages/test_file_dependency.py
+++ b/tests/packages/test_file_dependency.py
@@ -67,7 +67,7 @@ def test_file_dependency_pep_508_local_file_relative_path(mocker):
     _test_file_dependency_pep_508(mocker, "demo", path, requirement)
 
 
-def test_to_pep_508_with_marker(mocker):
+def test_absolute_file_dependency_to_pep_508_with_marker(mocker):
     wheel = "demo-0.1.0-py2.py3-none-any.whl"
 
     abs_path = DIST_PATH / wheel
@@ -81,6 +81,10 @@ def test_to_pep_508_with_marker(mocker):
         requirement,
         marker=SingleMarker("sys.platform", "linux"),
     )
+
+
+def test_relative_file_dependency_to_pep_508_with_marker(mocker):
+    wheel = "demo-0.1.0-py2.py3-none-any.whl"
 
     rel_path = Path("..") / "fixtures" / "distributions" / wheel
     requirement = '{} @ {} ; sys_platform == "linux"'.format(
--- a/tests/packages/test_file_dependency.py
+++ b/tests/packages/test_file_dependency.py
@@ -4,6 +4,7 @@ import pytest
 
 from poetry.core.packages.dependency import Dependency
 from poetry.core.packages.file_dependency import FileDependency
+from poetry.core.version.markers import SingleMarker
 
 
 DIST_PATH = Path(__file__).parent.parent / "fixtures" / "distributions"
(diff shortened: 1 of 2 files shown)


ossareh:
Looking into those windows errors...

ossareh:
Update: this solves the case of the missing space prior to the ";", but immediately runs up against the fact that a relative file is not a valid URL:

```
...snip...
  File "/home/ossareh/dev/src/github.com/erisyon/vega/.tox/py39/lib/python3.9/site-packages/pip/_vendor/pkg_resources/__init__.py", line 3033, in _compute_dependencies
    reqs.extend(parse_requirements(req))
  File "/home/ossareh/dev/src/github.com/erisyon/vega/.tox/py39/lib/python3.9/site-packages/pip/_vendor/pkg_resources/__init__.py", line 3094, in parse_requirements
    yield Requirement(line)
  File "/home/ossareh/dev/src/github.com/erisyon/vega/.tox/py39/lib/python3.9/site-packages/pip/_vendor/pkg_resources/__init__.py", line 3103, in __init__
    raise RequirementParseError(str(e))
pip._vendor.pkg_resources.RequirementParseError: Invalid URL: blob/pyvcam-2.1.0-cp39-cp39-linux_x86_64.whl
```

It's possible that my "fix" is actually fixing the wrong thing; if I was including this as an absolute path with a `file://` scheme the `is_url()` portion of [my change](https://github.com/python-poetry/poetry-core/pull/153/files#diff-4ee16b9cad93048318361c44a6fa750eb1a7d4c891f6f99302cfa9e61b102cdbR265) may actually just do the right thing.

In my case I can't provide an absolute path due to varied setup on hosts that'll receive this code.

I'm going to think about this some more, if you decide the close this PR without merging I understand :) - if you want changes made, I'm happy to make them.

ossareh:
> It's possible that my "fix" is actually fixing the wrong thing; if I was including this as an absolute path with a file:// scheme the is_url() portion of my change may actually just do the right thing.

after more research I see this is incorrect.

This change makes the output of this stage of the process 508 compliant, where as prior to this it was not. So in that regard this is indeed fixing **a** problem.

The fact that pip doesn't have a solution for understanding relative paths is a well discussed situation: https://github.com/pypa/pip/issues/6658

tests/packages/test_file_dependency.py
```diff
@@ -62,3 +65,31 @@ def test_file_dependency_pep_508_local_file_relative_path(mocker):
 
     requirement = "{} @ {}".format("demo", path)
     _test_file_dependency_pep_508(mocker, "demo", path, requirement)
+
+
+def test_to_pep_508_with_marker(mocker):
```
finswimmer, line 70: I think we can use a parameterize test here, to test absolute and relative path.

finswimmer (changes requested):
Thanks a lot. Looks good to me :+1: 

Just a small suggestion to improve the test.

ossareh:
@finswimmer thanks for the feedback 👍

I went ahead and restructured that unit test into two. I considered parameterizing the test and I could see two main options:

 1. a loop which calls the underlying test with an absolute path and a relative path
 2. calling a new `_test_to_508_with_marker` with the paths and expectation which then calls `_test_file_dependency_pep_508`
 3. splitting the unit test into two for each case

I felt the 3rd option strikes the best balance with clarity, it's a step necessary for the 2nd option, and I'm happy to add that additional layer if you think the deduplication is desirable.

finswimmer (approved):

PR #154, opened 2021-04-02
Resolve invalid python version constraint

pbzweihander:
Resolves: python-poetry#

~~[`backports.functools_lru_cache`](https://github.com/jaraco/backports.functools_lru_cache) has following dependency in `setup.cfg` (https://github.com/jaraco/backports.functools_lru_cache/blob/74120920aa77047323b7ca25533f57bdae4ead1c/setup.cfg#L48):~~

```
	pytest-mypy; python_implementation != "PyPy" and python_version < "3.10" and python_version > "3"
```

~~It has a python version constraint of `<3.10 >3` and it seems invalid. And it breaks poetry. Related issue: https://github.com/python-poetry/poetry/issues/534~~

The constraint mentioned above was not actually invalid. That case was resolved in #155.

This PR makes poetry can resolve dependencies with the broken constraints, such as `python_version >= "3.6" and python_version <= "3.4"`

### DIFF
--- a/tests/packages/test_main.py
+++ b/tests/packages/test_main.py
@@ -272,3 +272,12 @@ def test_dependency_from_pep_508_should_not_produce_empty_constraints_for_correc
         str(dep.marker)
         == 'platform_python_implementation != "PyPy" and python_version <= "3.10" and python_version > "3"'
     )
+
+
+def test_dependency_from_pep_508_with_invalid_python_version_constraint():
+    name = 'some-dependency ; python_version >= "3.6" and python_version <= "3.4"'
+    dep = Dependency.create_from_pep_508(name)
+
+    assert dep.name == "some-dependency"
+    assert str(dep.constraint) == "*"
+    assert dep.python_versions == "*"
--- a/poetry/core/packages/dependency.py
+++ b/poetry/core/packages/dependency.py
@@ -276,6 +276,7 @@ class Dependency(PackageSpecification):
     def _create_nested_marker(
         self, name: str, constraint: Union["BaseConstraint", "VersionTypes"]
     ) -> str:
+        from poetry.core.semver.empty_constraint import EmptyConstraint
         from poetry.core.semver.version import Version
         from poetry.core.semver.version_union import VersionUnion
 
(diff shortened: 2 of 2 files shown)


pawamoy:
> It has a python version constraint of <3.10 >3 and it seems invalid.

I don't think it's invalid. See https://github.com/python-poetry/poetry/issues/3862#issuecomment-812437280 and the following comment.

tests/packages/test_main.py
```diff
@@ -254,3 +254,12 @@ def test_dependency_from_pep_508_with_python_full_version_pep440_compatible_rele
...
     assert dep.python_versions == "~=3.4 || <3"
+
+
+def test_dependency_from_pep_508_with_invalid_python_version_constraint():
+    name = 'pytest-mypy ; python_version < "3.10" and python_version > "3"'
+    dep = Dependency.create_from_pep_508(name)
+
+    assert dep.name == "pytest-mypy"
+    assert str(dep.constraint) == "*"
+    assert dep.python_versions == "<3.10 >=4"
```
pawamoy, line 265: Based on https://www.python.org/dev/peps/pep-0508/#versions and https://www.python.org/dev/peps/pep-0508/#environment-markers, I'd say Poetry should not consider `> "3"` to be equivalent to `>= "4". These versions should be compared using PEP440, and with PEP440 `3.6` is greater than `3`.

```python
>>> import packaging
>>> from packaging import version
>>> version.parse("3.6") > version.parse("3")
True
>>> version.parse("3.6") >= version.parse("4")
False
```

pawamoy:
Anyway, your changes are still important for other invalid markers :+1:

pbzweihander:
I have updated the tests for 'real' invalid constraint such as `python_version >= "3.6" and python_version <= "3.4"`

pbzweihander:
Oops, it seems like the test case with `python_version >= "3.6" and python_version <= "3.4"` does not produce `EmptyConstraint`...

pbzweihander:
I cannot come up with a python version constraint that produces `EmptyConstraint`. I'll just close this PR this moment. The actual issue was resolved in #155

PR #155, opened 2021-04-02
Fix an error when handling single digit Python markers

sdispater:
Resolves: python-poetry/poetry#3862

Supersedes #154 

The code responsible for this error (introduced 2 years ago https://github.com/python-poetry/poetry/commit/46ad807209b97aaf88fe17fa90b227cb7a398a5f) just needed to be removed.

This code was here to translate PEP 508 markers conditions to Poetry's internal comparison system. However, it seems that `packaging` interprets `>3` as `>3.0.0` just like Poetry does so it seems that this is no longer needed.

### DIFF
--- a/poetry/core/packages/dependency.py
+++ b/poetry/core/packages/dependency.py
@@ -409,7 +409,6 @@ class Dependency(PackageSpecification):
         path is specified, this is used as the base directory if the identified dependency is
         of file or directory type.
         """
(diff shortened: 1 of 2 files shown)


pawamoy:
I guess this would also fix https://github.com/python-poetry/poetry/issues/2372?

abn (approved):

PR #157, opened 2021-04-02
pep440: support upper bound post/local release comparisons

abn:
This change ensures that post and local releases are taken into 
consideration when checking if version range allows a post release 
local build release at upper and lower bounds.

The following conditions now hold for upper bound checks.

- `<=3.0.0` allows `3.0.0+local.1`, `3.0.0-1`
- `<=3.0.0+local.1` disallows `3.0.0+local.2`, allows `3.0.0-1` 
- `<=3.0.0-1` allows `3.0.0+local.1`, `3.0.0`

Lower bound checks require no modification and works due to the 
implicit version comparison of `poetry.core.pep440.PEP440Version`.

Note: Majority of the changes here are test coverage.

### DIFF
--- a/poetry/core/semver/version_range.py
+++ b/poetry/core/semver/version_range.py
@@ -74,10 +74,20 @@ class VersionRange(VersionRangeConstraint):
                 return False
 
         if self.full_max is not None:
-            if other > self.full_max:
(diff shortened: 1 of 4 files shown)


sdispater (approved):
Looks good to me 👍

PR #158, opened 2021-04-03
pep440: support post/local for semver allows

abn:
This change ensures that post and local releases are taken into
consideration when checking if semver version instance allows
post and local build releases.

The following conditions now hold `poetry.core.semver.Version.allows`.

- `3.0.0` allows `3.0.0+local.1`, `3.0.0-1`
- `3.0.0+local.1` disallows `3.0.0+local.2`, allows `3.0.0-1`
- `3.0.0-1` disallows `3.0.0`, `3.0.0+local.1`, allows `3.0.0-1+local.1`

This is a follow up from #157 .

Note this does not change equality check behaviour.

### DIFF
--- a/poetry/core/semver/version.py
+++ b/poetry/core/semver/version.py
@@ -80,10 +80,29 @@ class Version(PEP440Version, VersionRangeConstraint):
         return False
 
     def allows(self, version: "Version") -> bool:
-        return self == version
+        if version is None:
+            return False
+
+        _this, _other = self, version
+
+        # allow weak equality to allow `3.0.0+local.1` for `3.0.0`
+        if not _this.is_local() and _other.is_local():
+            _other = _other.without_local()
+        elif _this.is_local() and not _other.is_local():
+            _this = _this.without_local()
+
+        # allow weak equality to allow `3.0.0-1` for `3.0.0`
+        if not _this.is_postrelease() and _other.is_postrelease():
+            _other = _other.without_postrelease()
+        elif _this.without_postrelease() and not _other.without_postrelease():
+            _this = _this.without_postrelease()
+
+        return _this == _other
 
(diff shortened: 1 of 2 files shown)


tests/semver/test_version.py
```diff
@@ -113,18 +113,38 @@ def test_allows():
...
+    assert not v.allows(Version.parse("1.2.3+build.2"))
+    assert v.allows(Version.parse("1.2.3-1+build.1"))
 
 
 def test_allows_all():
     v = Version.parse("1.2.3")
 
     assert v.allows_all(v)
     assert not v.allows_all(Version.parse("0.0.3"))
-    assert not v.allows_all(
```
abn, line 124: @sdispater from what I can tell this was incorrect here. I am guessing this is the right fix.

sdispater: It is correct though. Version 1.2.3 does not allow all versions in the range 1.1.4 to 1.2.4 and a single version does not allow any version.

poetry/core/semver/version.py
```diff
@@ -80,10 +80,24 @@ def is_empty(self) -> bool:
...
+        if not _this.is_postrelease() and _other.is_postrelease():
+            _other = _other.without_postrelease()
+        elif _this.without_postrelease() and not _other.without_postrelease():
+            _this = _this.without_postrelease()
+
+        return _this == _other
 
     def allows_all(self, other: "VersionTypes") -> bool:
-        return other.is_empty() or other == self
+        return other.is_empty() or other.allows(self)
```
sdispater, line 100: We must leave it as is since a single version cannot allow all versions from a range unless this range is the same as the single version.

sdispater (changes requested):

sdispater (approved):

PR #159, opened 2021-04-07
Simplify/improve Package.clone()

0xDEC0DE:
Downstream clients (e.g., poetry) appear to assume that the result of a Package.clone() will be a full-fidelity copy of the original.

Use `copy.deepcopy` to clone the object, and ensure that this is the case.

Resolves: python-poetry/poetry#2422
Replaces: python-poetry/poetry#2611

- [✅] Added **tests** for changed code.
- [✅] Updated **documentation** for changed code.


sdispater (approved):
Thanks! Looks good to me 👍

PR #164, opened 2021-04-09
deps: relax compatibility requirement constraints

abn:
The compatibility dependency here prevents us from upgrading the poetry dependency.


sdispater (approved):

brechtm:
It looks like this change isn't included in the current released version (1.0.7). Could a new release be made since the _importlib-metadata_ version constraint is causing conflicts with other (minimum) version constraints in my Poetry-managed development venv.

PR #168, opened 2021-04-10
ci: update test scripts to use install-poetry.py and cleanup release workflow

abn:
This change also fixes python 3.10 failures.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -6,125 +6,48 @@ on:
       - '*.*.*'
 
 jobs:
-
-  Linux:
-    runs-on: ubuntu-latest
-
-    steps:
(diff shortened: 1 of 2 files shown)


abn:
THe new 3.10 failures are due to missing `setuptools` in the project virtual environment. Since new versions of poetry drops `pip` and `setuptools` from project environments, we have an issue with adding it as a development dependency (which is the right fix) in that adding via `poetry@master` will lead to all existing versions to fail as they mark `setuptools` and unsafe, leading to resolution errors.

@sdispater I wonder if this makes a good case for `pyproject.toml` allowing a configuration for `min-poetry` similar to what `tox` does.

stephsamson (approved):

PR #173, opened 2021-04-30
Include tests in sdist

kwshi:
Resolves: python-poetry/poetry#3952.

- [x] Added **tests** for changed code. (N/A)
- [x] Updated **documentation** for changed code. (N/A)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,6 +21,7 @@ classifiers = [
 packages = [
     {include = "poetry"},
 ]
+include = ["tests"]
 exclude = [
     "**/*.pyc",
     "**/*.pyi",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,7 +21,9 @@ classifiers = [
 packages = [
     {include = "poetry"},
 ]
-include = [{ path = "tests", format = "sdist" }]
+include = [
+    { path = "tests", format = "sdist" },
+]
 exclude = [
     "**/*.pyc",
     "**/*.pyi",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,7 +21,7 @@ classifiers = [
 packages = [
     {include = "poetry"},
 ]
-include = ["tests"]
+include = [{ path = "tests", format = "sdist" }]
 exclude = [
     "**/*.pyc",
     "**/*.pyi",


pyproject.toml
```diff
@@ -21,6 +21,7 @@ classifiers = [
 packages = [
     {include = "poetry"},
 ]
+include = ["tests"]
```
sdispater, line 24: This will include tests in the wheel as well.

It would be better to add the `tests` package to the list of packages:

```toml
packages = [
    {include = "poetry"},
    {include = "tests", format = "sdist"}
]
```

sdispater (changes requested):

pyproject.toml
abn: @sdispater think what we want is `include` not `packages`.

sdispater: @abn Yes, even though `packages` would still work in this case, maybe it's better to put it in `include`.

kwshi: 1. I have implemented the change to specify `format = "sdist"` in 5525c3c, because as @abn mentioned on python-poetry/poetry#4007, it's better to be explicit about intentions.
2. However, I'm confused about the actual behavior of this option:
  - Based on what I can tell, regardless of whether or not the `format = "sdist"` is specified, the tests do not appear in the wheel anyway--I confirmed this by running `poetry build -f wheel` on the project once _with_ and once _without_ the `format = "sdist"` and running `unzip -l dist/poetry_core*.whl` to examine the contents of the wheel; in both cases there were the same number of files (208), none of them including the `tests/` path.  Perhaps this is not the expected behavior?
  - Additionally, it appears that this `include` format is not mentioned in the [official Poetry documentation on `include`](https://python-poetry.org/docs/pyproject/#include-and-exclude).  Should I be looking elsewhere, or is this a shortage in the documentation?
3. I agree with @abn that `include` is preferable and not `packages`, because `packages` would cause the resulting wheel to expose a global Python module named `tests`, which at best exposes a module never meant to be actually imported by end users and at worst causes artificial installation conflicts between different libraries (e.g. `tomlkit` and `cachy`'s sdists produce installations that both expose a `tests` module, making them impossible to install together without patching, even though they otherwise have no reason to conflict).

pyproject.toml
```diff
@@ -21,6 +21,7 @@ classifiers = [
 packages = [
     {include = "poetry"},
 ]
+include = [{ path = "tests", format = "sdist" }]
```
abn, line 24: Sorry for the nitpick, but can you please add the entry on a new line similar to packages please.

abn (changes requested):

pyproject.toml
kwshi: No problem!  Fixed in 5f0fe08.

pyproject.toml
abn: As for (2), this is expected. There is a combination of conditions here leading to the situation.

1. The project specifies that it must only include the package `poetry`. This applies to wheel.
2. When no formats are specified, the wheel builder falls back onto (1) to determine whether to include something as a package (since `tests` is a package).

If you specify wheel as a format for the include, seems that forces the inclusion irrespective of it being a package.  More reason why to be explicit. :)

The documentation does not include `format` yet as it seems to be only on `master` at the moment (the doc fix).

abn (approved):

PR #176, opened 2021-05-06
update locked dependencies

abn:


finswimmer:
Seems to outdated for a long while now :)

PR #186, opened 2021-08-16
Fix the parsing of wildcards dependency constraints

sdispater:
Resolves: python-poetry/poetry#4176

### DIFF
--- a/poetry/core/semver/helpers.py
+++ b/poetry/core/semver/helpers.py
@@ -130,6 +130,13 @@ def parse_single_constraint(constraint: str) -> VersionTypes:
         op = m.group(1)
         version = m.group(2)
 
+        # Technically invalid constraints like `>= 3.*` will appear
+        # here as `3.`.
+        # Pip currently supports these and to avoid breaking existing
+        # users workflows we need to support them as well. To do so,
+        # we just remove the inconsequential part.
+        version = version.rstrip(".")
+
         if version == "dev":
             version = "0.0-dev"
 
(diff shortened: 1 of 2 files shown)


abn (approved):

bentheiii:
I think this fix is incorrect, as `name < 3.*` should accept version`3.6.0`, I think `name < 3.*` should be normalized to `name < 4`

finswimmer:
Hello @bentheiii,

`name < 3.*` could be read as "any version smaller 3.anything". So it's `< 3`.

> I think this fix is incorrect.

Please don't call it a "fix" (even it's in the title ;) ). It's a (temporary) workaround, to mimic pip's wrong behavior of allowing this kind of invalid PEP 440 constraint. Otherwise to many packages cannot be installed anymore. (See also: https://discuss.python.org/t/pep-440-wildcard-usage-in-comparison-clause/10045)

bentheiii:
It is indeed a pretty unclear whether `name (?) X.*` should resolve to "version chosen must match the constraint for **any** `X.*`" or "version chosen must match the constraint for **all** `X.*`". But since `name > 3.*` resolves to `name > 3`, poetry obviously goes with the former. Therefore, `name < 3.*` has to be less than any version `3.*` , it has to only be less than an arbitrarily high `*`, normalizing to `name < 4` (just like how `name > 3.*` must be higher than an arbitrarily low `*`, normalizing to `name > 3`).

edgarrmondragon:
@furlongm is there a reason not to install from `master` in the meantime if the feature you need is there?

furlongm:
> @furlongm is there a reason not to install from `master` in the meantime if the feature you need is there?

Mostly a desire to pin the poetry dependencies at a specific version so it's not a moving target - also don't always have access to github but do have a pypi mirror, so released versions are a lot more accessible.

PR #187, opened 2021-08-18
Issue 4398: Fix constraint and pep 660 hook typo

airstandley:
Resolves: python-poetry#4398

Would very much appreciate an expert reviewing the new tests to see if the behaviour checked is correct. I do not have time to parse through the code base in enough detail to be certain about what the expected behaviour should be so this is my best guess.


airstandley:
Fixed by #188

PR #194, opened 2021-08-22
Fix an additional case for python-poetry/poetry#4402

morne:
Resolves: python-poetry#4402

### DIFF
--- a/poetry/core/version/markers.py
+++ b/poetry/core/version/markers.py
@@ -297,6 +297,9 @@ class SingleMarker(BaseMarker):
         if self._name not in environment:
             return True
 
+        if self._constraint.is_empty():
+            return True
+
         return self._constraint.allows(self._parser(environment[self._name]))
 
     def without_extras(self) -> MarkerTypes:


radoering:
Can you give more detail on why this fix is required, a concrete use-case for example?

The linked issue has been closed by another pull request that introduced an `allows` method for EmptyConstraint. However, this method returns `False` instead of `True` as proposed in your fix. In theory, returning `False` seems to be correct when comparing to AnyConstraint. (AnyConstraint allows everything and EmptyConstraint allows nothing.) But my theory may be wrong...

Secrus:
@morne any update on this?

morne:
Apologies for the very late reply, i missed the notifications.

> The linked issue has been closed by another pull request that introduced an allows method for EmptyConstraint. However, this method returns False instead of True as proposed in your fix. In theory, returning False seems to be correct when comparing to AnyConstraint. (AnyConstraint allows everything and EmptyConstraint allows nothing.) But my theory may be wrong...

I think this is the correct fix, and my pull request can be ignored. I seems that I had not properly updated some files at the time.

PR #196, opened 2021-09-08
support specifying algorithm when calculating file dependency hash.

kaos:
Signed-off-by: Andreas Stenius <andreas.stenius@svenskaspel.se>

Pre-work for: python-poetry/poetry#4085

Currently, poetry assumes all archive file hashes are using `sha256`. This opens for the ability to support other algorithms. (I have a real use case, where the archive uses md5, for instance.)

### DIFF
--- a/poetry/core/packages/file_dependency.py
+++ b/poetry/core/packages/file_dependency.py
@@ -69,8 +69,8 @@ class FileDependency(Dependency):
     def is_file(self) -> bool:
         return True
 
-    def hash(self) -> str:
-        h = hashlib.sha256()
+    def hash(self, algorithm: str = "sha256") -> str:
+        h = hashlib.new(algorithm)
         with self._full_path.open("rb") as fp:
             for content in iter(lambda: fp.read(io.DEFAULT_BUFFER_SIZE), b""):
                 h.update(content)
--- a/tests/packages/test_file_dependency.py
+++ b/tests/packages/test_file_dependency.py
@@ -1,3 +1,5 @@
+from contextlib import contextmanager
+from io import BytesIO
 from pathlib import Path
 
 import pytest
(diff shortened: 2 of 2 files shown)


kaos:
Ping. Any feedback on this?

wcn00:
I am currently blocked because all the packages in my gemfury repo are md5 signed so we cannot upgrade/install anything using them.  Is there any circumvention I can use while we wait for this pr to land?  This is a significant blocker for my company.

wcn00:
Thankyou @kaos   You're a gentleperson and a scholar :)  Disabling that feature gets me out of this bind which was threatening to ruin my week!
Thanks again

kaos:
Ping. @finswimmer  @abn  
I've got no reaction at all from any maintainers in the past two weeks on any of my 3 PRs. If you are really busy, a quick comment as such, would be appreciated, acknowledging that you've seen this.

Thank you.

kaos:
Bump.

I can address conflicts etc, if there is any sign this PR will be getting any review activity.
As I've found a work-around for the issues I had (with simply using the old resolver) I will not fret over these fixes..

FlorianLudwig:
@kaos Patience. I also once wanted wrote a patch for the same feature. I would estimate you have to wait another year to get it merged. (ref: https://github.com/python-poetry/poetry/issues/2422 for ETA for changes like this)

Writing "bump" here just gives notifications to those interested in the feature (like me) but doesn't make things faster

neersighted (approved):
This change is very small and well-tested. While overall I'm not enthusiastic about md5 support, I think we need to take the pragmatic approach here.

kaos:
This was fixed in 68bf05242e16f46640b0063677a80a1b8cf7f56e

neersighted:
Yep -- you're right. I'll take a look at the `poetry` changes next.

PR #198, opened 2021-09-15
use flake8-bugbear for linting

danieleades:
bugbear is a collection of fairly uncontroversial lints


neersighted (approved):
I planned to add this myself (working on getting poetry's test suite perfected and then planning to port it), but I think this is good incremental progress (i.e. get it in and fix the lints). It won't conflict with my later changes, so LGTM!

PR #199, opened 2021-09-15
set up mypy hook for incremental adoption

danieleades:
as per the comment here - https://github.com/python-poetry/poetry/pull/4510

for some reason i had to use different config to successfully exclude the `_vendor` directory


neersighted (changes requested):
Same as on poetry -- lets add some comments and rebase to fix the merge conflicts.

neersighted (dismissed):

danieleades:
@neersighted this should be good to go. Not sure what's caused that python 3.10 workflow failure

branchvincent:
The mypy error is an open issue (see https://github.com/python/mypy/issues/1393) and only occurs on 3.10 since `collections.Mapping` was [removed](https://docs.python.org/3/whatsnew/3.10.html#removed) (8th bullet). We could either:

1. Ignore it with `# type: ignore`
2. Replace it with a non-guarded `from collections.abc import Mapping`, given it's unneeded compatibility for python 2

neersighted:
Go ahead with `collections.abc` -- we're slowly trying to rip out Python 2 compatibility from poetry and poetry core -- there's no reason not to use the correct import path these days.

danieleades:
@neersighted do you mind triggering the CI for me?

mypy.ini
```diff
@@ -0,0 +1,37 @@
+[mypy]
```
neersighted, line 1: Hate to do this to you in the midnight hour, I didn't notice that this was a new file until now. Can you please implement this ala-poetry in `pyproject.toml`?

.pre-commit-config.yaml
```diff
@@ -16,6 +16,21 @@ repos:
               | ^poetry/core/_vendor
           )
 
+  - repo: https://github.com/pre-commit/mirrors-mypy
+    rev: v0.910
+    hooks:
+      - id: mypy
+        additional_dependencies:
+          - types-requests
+        files: |
```
neersighted, line 25: Let's codify these paths in `pyproject.toml` ala-poetry so that editor-invoked mypy will respect them.

poetry/core/_vendor/tomlkit/_utils.py
```diff
@@ -10,10 +10,7 @@
 from ._compat import timezone
 
 
-try:
-    from collections.abc import Mapping
-except ImportError:
-    from collections import Mapping
+from collections.abc import Mapping
```
branchvincent, line 13: i assume we don't want to make any changes to the vendored code in `poetry/core/_vendor`

danieleades: ah rats. good catch.
reverted

branchvincent (approved):
fwiw, this looks good to me

neersighted (changes requested):
Please note that when I say 'like in Poetry,' I really mean 'like in the linked PR' that I am working to get merged.

https://github.com/python-poetry/poetry/pull/4750

mypy.ini
danieleades: i've moved the config to `pyproject.toml`, but i'm not totally happy with it. running `poetry run mypy` results in a bunch of errors in the `_vendor` module, even if it's added to the 'whitelist'. not 100% sure why that is. potentially have to silence the import errors? will try and find some time in the next few days for a more thorough investigation.

pyproject.toml
```diff
@@ -84,6 +85,35 @@ filter_files = true
 known_first_party = "poetry.core"
 known_third_party = ["poetry.core._vendor"]
 
+[tool.mypy]
+python_version = "3.6"
```
neersighted, line 89: Since we seem to get by with a lesser set of mypy arguments, I would just go with `follow_imports = silent` for now. The options are mostly cargo-cult so I'm comfortable changing later if the need presents itself, and then trying to unify poetry and python-poetry even later.

pyproject.toml
```diff
@@ -84,6 +85,35 @@ filter_files = true
...
+warn_redundant_casts = true
+warn_unused_configs = true
+warn_unused_ignores = true
+files = "poetry"
+
+# The following whitelist is used to allow for incremental adoption
+# of Mypy. Modules should be removed from this whitelist as and when
+# their respective type errors have been addressed. No new modules
+# should be added to this whitelist.
+# see https://github.com/python-poetry/poetry/pull/4510.
```
neersighted, line 101: This URL should be updated to match this PR (#199 on poetry-core).

neersighted (changes requested):

pyproject.toml
```diff
@@ -84,6 +85,35 @@ filter_files = true
...
 known_third_party = ["poetry.core._vendor"]
 
+[tool.mypy]
+python_version = "3.6"
+check_untyped_defs = true
+ignore_missing_imports = true
+warn_redundant_casts = true
+warn_unused_configs = true
+warn_unused_ignores = true
+files = "poetry"
```
branchvincent, line 95: to exclude typechecking `_vendor`:

```suggestion
files = "poetry"
exclude = "poetry/core/_vendor"
```

mypy.ini
neersighted: Instead of using a module whitelist (that won't work as the modules in `_vendor` are not part of the `poetry.core` namespace), have you tried a file-glob exclude in `pyproject.toml`?

https://mypy.readthedocs.io/en/stable/config_file.html#confval-exclude

branchvincent: does https://github.com/python-poetry/poetry-core/pull/199#discussion_r748913374 help?

pyproject.toml
neersighted: (keep the `python_version` however)

.pre-commit-config.yaml
```diff
@@ -16,6 +16,21 @@ repos:
...
+          (?x)(
+            ^poetry
+          )
+        exclude: |
+          (?x)(
+              ^poetry/core/_vendor
+          )
```
branchvincent, line 32: since we've moved our include/exclude rules to `pyproject.toml`, this can be simplified and closer aligned to the [config in poetry](https://github.com/python-poetry/poetry/blob/f7bbd54f5e543523d4d81221c9a8a4daee2e1315/.pre-commit-config.yaml#L43-L50)

```suggestion
    rev: v0.910-1
    hooks:
      - id: mypy
        pass_filenames: false
        additional_dependencies:
          - types-requests
```

mypy.ini
neersighted: > 
> 
> does [#199 (comment)](https://github.com/python-poetry/poetry-core/pull/199#discussion_r748913374) help?

It was exactly what I meant!

danieleades:
i've aligned the pre-commit config with the `pyproject.toml` file. my only concern is that the blanket `allow_missing_imports` flag might be slightly too unrestrictive/untargeted. Then again, it appears to be baked into the pre-commit hook.

pyproject.toml
danieleades: done

.pre-commit-config.yaml
```diff
@@ -16,6 +16,20 @@ repos:
               | ^poetry/core/_vendor
           )
 
+  - repo: https://github.com/pre-commit/mirrors-mypy
+    rev: v0.910
+    hooks:
+      - id: mypy
+        pass_filenames: false
+        # files: |
```
neersighted, line 24: Let's go ahead and elide this entirely -- there should be no reason to directly exclude files in pre-commit at this time.

neersighted (changes requested):

neersighted (approved):

neersighted:
Looks like we get a failure on pypy, which is interesting. Is this rebased onto the latest master?

danieleades:
> Looks like we get a failure on pypy, which is interesting. Is this rebased onto the latest master?

it is now

danieleades:
> Looks like we get a failure on pypy, which is interesting. Is this rebased onto the latest master?

https://github.com/python/typed_ast/issues/111

danieleades:
rebased on master again. let's see if that helps

neersighted:
> 
> 
> > Looks like we get a failure on pypy, which is interesting. Is this rebased onto the latest master?
> 
> [python/typed_ast#111](https://github.com/python/typed_ast/issues/111)

I think the better thing to do here is to not depend on mypy in `pyproject.toml` at all -- we run it only under pre-commit anyway. If the user chooses to install mypy separately, that can be their (partially unsupported) business I think.

danieleades:
> > > Looks like we get a failure on pypy, which is interesting. Is this rebased onto the latest master?
> > 
> > 
> > [python/typed_ast#111](https://github.com/python/typed_ast/issues/111)
> 
> I think the better thing to do here is to not depend on mypy in `pyproject.toml` at all -- we run it only under pre-commit anyway. If the user chooses to install mypy separately, that can be their (partially unsupported) business I think.
> 
> Alternatively, we could add a market to only install on CPython (see the black dep above) -- but I'm thinking we should just remove the black and isort deps in the future and rely on pre-commit as the main poetry repo does.

i've removed the mypy dep for now.

neersighted (approved):

PR #179, opened 2021-05-18
chore: upgrade codebase to 3.6+

branchvincent:
Relates-to: python-poetry/poetry#3860

- [ ] Added **tests** for changed code.
- [ ] Updated **documentation** for changed code.

In the same spirit as python-poetry/poetry#3988, this:
* Applies `pyupgrade --py36-plus` to the codebase
* Adds the above to `.pre-commit-config.yaml`

PR #200, opened 2021-09-15
use pyupgrade to update syntax

danieleades:


danieleades:
this is a duplicate of #179, this can be closed

PR #207, opened 2021-09-20
Support most of the guaranteed hashes

awilkins:
Works toward resolution of python-poetry/poetry#4523
Also needs client changes

- [X] Added tests for changed code
- [X] Added documentation for changed code

Documentation is at least equivalent to the existing stuff.

- Implement most of the "guaranteed" hashes as per PEP 503
- Leave out shake, because they're variable length
  - Feel it's unlikely anyone's repo manager is using these

No offence to @hydra-zim who has produced pretty much the same PR @ #206 : this one has tests, and also targets the current mainline release branch instead of the future. The future is great, I would prefer this was fixed sooner.

### DIFF
--- a/poetry/core/packages/file_dependency.py
+++ b/poetry/core/packages/file_dependency.py
@@ -68,8 +68,8 @@ class FileDependency(Dependency):
     def is_file(self):  # type: () -> bool
         return True
 
-    def hash(self):  # type: () -> str
-        h = hashlib.sha256()
+    def hash(self, hash_name = "sha256"):  # type: (str) -> str
+        h = hashlib.new(hash_name)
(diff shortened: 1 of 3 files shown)


timbedard:
@sdispater python-poetry/poetry#4523 introduces some pretty critical breakage for anyone using private package repositories, and in a situation where version pins are not common (pinning a subdependency of a build tool). Do you or another maintainer have the bandwidth to review this and aim for a 1.0.6 release ASAP?

sdispater:
@awilkins Thanks for the PR!

You will need to fix the linting issues before I can merge this.

Also, it seems that some of the hash types are not available on Python 2.7 and 3.5.

sdispater (approved):

PR #214, opened 2021-10-04
Keep _constraint and _pretty_constraint of Dependency synchronized when calling set_constraint()

radoering:
Resolves: python-poetry/poetry#4589

### DIFF
--- a/src/poetry/core/packages/dependency.py
+++ b/src/poetry/core/packages/dependency.py
@@ -56,9 +56,9 @@ class Dependency(PackageSpecification):
         )
 
         self._constraint = None
+        self._pretty_constraint = None
         self.set_constraint(constraint=constraint)
 
-        self._pretty_constraint = str(constraint)
         self._optional = optional
 
         if not groups:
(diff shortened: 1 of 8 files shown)


neersighted (commented):
Is there a compelling reason we can't remove the internal property and rely on the getter to call str()? I'm not super familiar with the code here, but this just seems to be much more error-prone than computing the value on demand.

radoering:
The only reason I can imagine (without having examined it) is performance.

neersighted:
Let's go ahead and drop the property then -- I think that implementation will be much more obviously correct and will not allow things to get out of sync in the future.

radoering:
After dropping the property, I had to adapt some tests because the constraint style changed. With the property pretty_constraint was determined before parsing the constraint, now it is determined after parsing the constraint.

neersighted:
> After dropping the property, I had to adapt some tests because the constraint style changed. With the property pretty_constraint was determined before parsing the constraint, now it is determined after parsing the constraint.

Gah, looks like it's a pretty leaky abstraction overall. I'm inclined to just roll back to your first changeset and plan to refactor this later.

radoering:
No problem. In any case, less risky.

neersighted (approved):

PR #144, opened 2021-03-25
Expand ci test matrix and add support for Python 3.10

abn:
- Add support for experimental 3.10 builds
- Updates integration test `setup-python` action to `v2`
- Expand python versions for integration tests

~3.10 is expected to fail due to poetry bootstrap issue (see python-poetry/poetry#3828).~ We resolve this here by switching ti pip install until the new bootstrap is available.

Relates-to: python-poetry/poetry#3187
Relates-to: python-poetry/poetry#3828

PR #215, opened 2021-10-04
[1.0] Automatically add Python 3.10 classifiers

sdispater:
Now that Python 3.10 is around the corner, we should automatically add the `Python 3.10` classifier.


abn:
Back-port from #144

PR #219, opened 2021-10-09
Require that package descriptions not include newlines

tipabu:
Previously, we would include the description-with-newlines directly as the PKG-INFO summary, which could cause subtly broken builds (for instance, the package may install, but none of the specified dependencies).

Now, raise a validation error during building, like:

    RuntimeError

    The Poetry configuration is invalid:
      - [description] 'First line\nSecond line (BOOOOOM)' does not match '^[^\n]+$'

Resolves: python-poetry/poetry#1372

### DIFF
--- a/poetry/core/json/schemas/poetry-schema.json
+++ b/poetry/core/json/schemas/poetry-schema.json
@@ -19,7 +19,8 @@
     },
     "description": {
       "type": "string",
-      "description": "Short package description."
+      "description": "Short package description.",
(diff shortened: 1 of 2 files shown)


neersighted (approved):
Looks pretty straightforward and solves the bug nicely. I have to admit the error message is not the best, but I'm in favor of merging this and cleaning up the error later if it's non-obvious to people, since the issue is longstanding.

PR #223, opened 2021-10-19
perf: optimise ``Release.__post_init__`` method

P403n1x87:
The `Release.__post_init__` method has been optimised by avoiding function calls in favour of a single list comprehension. Profiling data shows the benefits of such change. There is less time spent in `__post_init__` after with the list comprehension than before with the calls to `list`, `map` and `filter`.

## Before

release_post_init_before

## After

release_post_init_after


poetry/core/version/pep440/segments.py
```diff
@@ -45,15 +45,11 @@ def __post_init__(self):
...
-        parts = list(
-            map(
-                str,
-                filter(
-                    lambda x: x is not None,
-                    [self.major, self.minor, self.patch, *self.extra],
-                ),
-            )
-        )
+        parts = [
```
neersighted, line 48: The only thing I don't like here is the use of `_` as a variable name when the value is not discarded. This goes against general Python convention. `part` or something similar would be suitable and make it clear that the value is used.

neersighted (changes requested):

neersighted (approved):

PR #225, opened 2021-10-28
[Hotfix] Transform Link back into Path inside FileDependency

rsayn:
Resolves: python-poetry/poetry#4541 (& related issues)

## Context
Whenever I try installing dependencies with `poetry-core 1.0.7`, I run into an `AttributeError` since a `Link` object is passed to `FileDependency` instead of a `Path` object.

I saw that the [code that initializes the class in Poetry performs a cast](https://github.com/python-poetry/poetry/blob/82459bd6f38785882a8eff0fbce2a8b0119b205c/poetry/installation/executor.py#L678), but for some reason this cast is not working on my local machine. The fix I performed directly inside the `FileDependency` class is the only thing that made Poetry work consistently on my machine.

I tried running every Poetry version in the range `1.1.6 -> 1.1.11`, without success.

My local machine is a MacOS, not sure if this is the cause since I saw lots of issues about this behaviour in the Poetry repo.

## Checklist

### DIFF
--- a/poetry/core/packages/file_dependency.py
+++ b/poetry/core/packages/file_dependency.py
@@ -2,13 +2,14 @@ import hashlib
 import io
 
 from pathlib import Path
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, cast
 from typing import FrozenSet
 from typing import List
 from typing import Optional
 from typing import Union
 
 from poetry.core.packages.utils.utils import path_to_url
+from poetry.core.packages.utils.link import Link
 
 from .dependency import Dependency
 
@@ -30,6 +31,12 @@ class FileDependency(Dependency):
         self._path = path
         self._base = base or Path.cwd()
         self._full_path = path
+
+        if isinstance(self._path, Link):
+            link = cast(Link, self._path)
+            self._path = Path(link.filename)
+            if self._path.is_absolute():
+                self._full_path = self._path
 
         if not self._path.is_absolute():
             try:


radoering:
The cast has been introduced with `poetry 1.1.9`. If you check the [source](https://github.com/python-poetry/poetry/blob/1.1.8/poetry/installation/executor.py#L611) of `poetry 1.1.8,` you can see that there is no cast in this version. I can't imagine how a `Link` object could be passed to `FileDependency` in `poetry 1.1.9` and above with the cast. Maybe, you have experienced another issue with `1.1.9` to `1.1.11`?

In `1.1.12`, there has been [another fix](https://github.com/python-poetry/poetry/blob/1.1.12/poetry/installation/executor.py#L615-L624) considering this part of the code. This was [primarily for Windows](https://github.com/python-poetry/poetry/pull/4531), but you may want to check if `1.1.12` fixes your issue, too.

rsayn:
@radoering As I wrote in my initial comment:
> I tried running every Poetry version in the range 1.1.6 -> 1.1.11, without success

I feel this is not an issue with the individual Poetry package, but with the Poetry + Poetry-Core combination.
Anyway I ended up hard-coding the cast on my machine under `site-packages` and I've been running smoothly since then 😄

rsayn:
@radoering I can confirm this still happens with Poetry 1.1.12, this time on a Windows 10 machine on which I had not applied the fix I listed in my first comment

radoering:
At the moment, I can only imagine that your poetry installation has been corrupted somehow.

Maybe you can check, if the following files look like the tagged versions for poetry 1.1.12 resp. poetry-core 1.0.7: When looking into [poetry/installation/executor.py#L615-L624](https://github.com/python-poetry/poetry/blob/1.1.12/poetry/installation/executor.py#L615-L624) you should see that `url_to_path(archive.url)` is called if `archive` is a `Link`. And when looking into [poetry/core/packages/utils/utils.py#L67-L90](https://github.com/python-poetry/poetry-core/blob/1.0.7/poetry/core/packages/utils/utils.py#L67-L90) you should see that a `Path` object is returned by `url_to_path`. If the issue occurs even if both files seem to be ok, probably some debugging is required.

I would appreciate it if you could find out more. I think we need more information on the root cause of the issue.

PR #229, opened 2021-11-13
Refactor typing in version.pep440.segments

danieleades:
this PR is stacked on top of #199, so i'll leave it as a draft until that's merged

addresses the type errors in `version.pep440.segments`, and narrows the whitelist

### DIFF
--- a/poetry/core/version/pep440/segments.py
+++ b/poetry/core/version/pep440/segments.py
@@ -1,5 +1,6 @@
 import dataclasses
 
+from typing import Iterable
 from typing import Optional
 from typing import Tuple
 from typing import Union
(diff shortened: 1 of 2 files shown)


poetry/core/version/pep440/segments.py
```diff
@@ -30,20 +31,16 @@ class Release:
     minor: Optional[int] = dataclasses.field(default=None, compare=False)
     patch: Optional[int] = dataclasses.field(default=None, compare=False)
     # some projects use non-semver versioning schemes, eg: 1.2.3.4
-    extra: Optional[Union[int, Tuple[int, ...]]] = dataclasses.field(
-        default=None, compare=False
+    # (this appears to trigger a mypy false positive)
+    extra: Union[int, Tuple[int, ...]] = dataclasses.field(  # type: ignore
+        default_factory=tuple, compare=False
```
danieleades, line 36: can i get a quick sanity check here @neersighted?

mypy.ini
```diff
@@ -0,0 +1,43 @@
+[mypy]
```
branchvincent, line 1: as done in https://github.com/python-poetry/poetry/pull/4750, it may make more sense to configure this in `pyproject.toml`

danieleades: I agree that's a better solution, but I dont think that change belongs in this PR. happy to rebase though, if the linked PR is merged first

branchvincent: oops, i see now your pr's are stacked 👍

so happy to see mypy getting in the codebase!

PR #230, opened 2021-11-13
remove some of the python2 compat and update syntax

danieleades:


poetry/core/masonry/builders/builder.py
```diff
@@ -364,17 +363,12 @@ def convert_author(cls, author: str) -> Dict[str, str]:
     @classmethod
     @contextmanager
     def temporary_directory(cls, *args: Any, **kwargs: Any) -> ContextManager[str]:
```
neersighted, line 365: This is an example of a now-unnecessary wrapper function, and I would prefer to leave it in place on the first commit and then excise it entirely on a follow-up commit

poetry/core/masonry/builders/builder.py
```diff
@@ -364,17 +363,12 @@ def convert_author(cls, author: str) -> Dict[str, str]:
...
-                yield name
-        except ImportError:
-            name = tempfile.mkdtemp(*args, **kwargs)
+        from tempfile import TemporaryDirectory
 
+        with TemporaryDirectory(*args, **kwargs) as name:
             yield name
 
-            shutil.rmtree(name)
+        shutil.rmtree(name)
```
neersighted, line 371: This change is strictly speaking incorrect -- the TemporaryDirectory contextmanager will clean up on its own. Removal of the entire wrapper is better than removing this line however.

neersighted (changes requested):
Looks good overall with one exception -- the refactor of the TemporaryDirectory wrapper is incorrect.

danieleades:
is there a reason not to simply add the pyupgrade hook?

branchvincent:
https://github.com/python-poetry/poetry/pull/4758 adds the hook for `poetry`, so i imagine we should do the same here

neersighted:
> 
> 
> [python-poetry/poetry#4758](https://github.com/python-poetry/poetry/pull/4758) adds the hook for `poetry`, so i imagine we should do the same here

The intention is to copy the automation from poetry once I've got it dialled in to everyone's satisfaction. Still, a PR adding pyupgrade would not be unwelcome -- I just plan to replace the `.pre-commit-config.yaml` in this repo in a day or two.

neersighted (dismissed):

poetry/core/masonry/builders/builder.py
neersighted: It looks like `rmtree` line is still around.

poetry/core/masonry/builders/builder.py
neersighted, line 365: We should completely eliminate this `temporary_directory` method in the first commit.

neersighted (changes requested):

poetry/core/masonry/builders/builder.py
danieleades: that's reared it's ugly head again after a rebase. expunged again now

neersighted (approved):

PR #233, opened 2021-11-15
Force type hints

finswimmer:
This PR add missing type hints to both package code and test code. Furthermore `flake-annotations` is added as a flake8 plugin to enforce type hints in the future.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,22 +1,20 @@
 [flake8]
 max-line-length = 88
-ignore = E501, E203, W503
-per-file-ignores = __init__.py:F401
+ignore = E501, E203, W503, ANN101, ANN102
+mypy-init-return = True
+per-file-ignores =
+    __init__.py:F401
+    tests/test_*:ANN201
+    tests/**/test_*:ANN201
 exclude =
     .git
     __pycache__
-    setup.py
     build
     dist
-    releases
     .venv
     .tox
-    .mypy_cache
     .pytest_cache
-    .vscode
     .github
-    poetry_core/_vendor/
-    poetry_core/utils/_compat.py
-    poetry_core/utils/_typing.py
-    tests/fixtures/
-    tests/masonry/fixtures/
+    poetry/core/_vendor/*
+    tests/fixtures/*
+    tests/**/fixtures/*
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -10,14 +10,9 @@ repos:
     hooks:
       - id: flake8
         additional_dependencies:
+          - flake8-annotations
           - flake8-bugbear
           - flake8-comprehensions
-        exclude: |
-          (?x)(
-              ^poetry/core/utils/_typing.py$
-              | ^poetry/core/utils/_compat.py$
-              | ^poetry/core/_vendor
-          )
 
   - repo: https://github.com/pre-commit/mirrors-mypy
     rev: v0.910
--- a/poetry/core/version/pep440/parser.py
+++ b/poetry/core/version/pep440/parser.py
@@ -61,7 +61,9 @@ class PEP440Parser:
         )
 
     @classmethod
-    def parse(cls, value: str, version_class: Optional[Type["PEP440Version"]] = None):
+    def parse(
+        cls, value: str, version_class: Optional[Type["PEP440Version"]] = None
+    ) -> "PEP440Version":
         match = cls._regex.search(value) if value else None
         if not match:
             raise InvalidVersion(f"Invalid PEP 440 version: '{value}'")
--- a/poetry/core/version/pep440/version.py
+++ b/poetry/core/version/pep440/version.py
@@ -1,6 +1,7 @@
 import dataclasses
 import math
 
+from typing import Any
 from typing import Optional
 from typing import Tuple
 from typing import Union
@@ -32,7 +33,7 @@ class PEP440Version:
         int, Release, ReleaseTag, ReleaseTag, ReleaseTag, Tuple[Union[int, str], ...]
     ] = dataclasses.field(default=None, init=False, compare=True)
 
-    def __post_init__(self):
+    def __post_init__(self) -> None:
         if self.local is not None and not isinstance(self.local, tuple):
             object.__setattr__(self, "local", (self.local,))
 
@@ -46,7 +47,11 @@ class PEP440Version:
 
         object.__setattr__(self, "_compare_key", self._make_compare_key())
 
-    def _make_compare_key(self):
+    def _make_compare_key(
+        self,
+    ) -> Tuple[
+        int, Release, ReleaseTag, ReleaseTag, ReleaseTag, Tuple[Tuple[float, str], ...]
+    ]:
         """
         This code is based on the implementation of packaging.version._cmpkey(..)
         """
@@ -103,7 +108,7 @@ class PEP440Version:
     def non_semver_parts(self) -> Optional[Tuple[int]]:
         return self.release.extra
 
-    def to_string(self, short=False):
+    def to_string(self, short: bool = False) -> str:
         dash = "-" if not short else ""
 
         version_string = dash.join(
@@ -209,7 +214,7 @@ class PEP440Version:
             epoch=self.epoch, release=self.release, pre=ReleaseTag(RELEASE_PHASE_ALPHA)
         )
 
-    def replace(self, **kwargs):
+    def replace(self, **kwargs: Any) -> "PEP440Version":
         return self.__class__(
             **{
                 **{
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,6 +1,7 @@
 import sys
 
 from pathlib import Path
+from typing import TYPE_CHECKING
 from typing import Callable
 
 import pytest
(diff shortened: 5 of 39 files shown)


.flake8
```diff
@@ -1,7 +1,11 @@
 [flake8]
```
neersighted, line 1: Thsi whole file is very cargo-culted. Can we clean it up while we're in here?

.flake8
```diff
@@ -1,7 +1,11 @@
...
 max-line-length = 88
-ignore = E501, E203, W503
-per-file-ignores = __init__.py:F401
+ignore = E501, E203, W503, ANN101, ANN102
+mypy-init-return = True
```
neersighted, line 4: Is this a mypy config setting? If so, please put it in pyproject.toml (see #199).

.pre-commit-config.yaml
```diff
@@ -9,11 +9,13 @@ repos:
     rev: 3.9.2
     hooks:
       - id: flake8
+        additional_dependencies: [flake8-annotations]
```
neersighted, line 12: We shouldn't need additional exclusions here if the `.flake8` ones are good.

poetry/core/version/pep440/version.py
```diff
@@ -46,7 +47,21 @@ def __post_init__(self):
 
         object.__setattr__(self, "_compare_key", self._make_compare_key())
 
-    def _make_compare_key(self):
+    def _make_compare_key(
```
neersighted, line 50: Wow, this is a complicated type. Is there any way to clean this up with NewType, a refactor with an object, or similar?

neersighted (changes requested):

tests/conftest.py
```diff
@@ -10,7 +11,12 @@
 from tests.testutils import tempfile
 
 
-def pytest_addoption(parser):
+if TYPE_CHECKING:
```
danieleades, line 14: you shouldn't need the `if TYPE_CHECKING` pattern here- that's really only for preventing circular imports. You can just import these directly, and use them below without quoting the types

tests/conftest.py
```diff
@@ -20,15 +26,15 @@ def pytest_addoption(parser):
     )
 
 
-def pytest_configure(config):
+def pytest_configure(config: "Config") -> None:
```
danieleades, line 29: you can use `Config` unquoted if you remove the `if TYPE_CHECKING` pattern above

.flake8
finswimmer: Nope, it's a `flake8-annotation` setting. See here: https://pypi.org/project/flake8-annotations/

> Allow omission of a return type hint for `__init__` if at least one argument is annotated. See mypy's documentation for additional details.

tests/json/test_poetry_schema.py
```diff
@@ -4,7 +4,7 @@
 
 
 @pytest.fixture
-def base_object():
+def base_object() -> dict:
```
danieleades, line 7: this syntax is python 3.9+. less than that, the correct return type would be `Dict`, rather than `dict`.

I guess we can get away with it, since `mypy` is being run with a recent interpreter, and the type is ignored during other testing, but whichever way we go, it should be consistent throughout the codebase

tests/masonry/builders/test_builder.py
```diff
@@ -10,7 +13,11 @@
 from poetry.core.utils._compat import PY37
 
 
-def test_builder_find_excluded_files(mocker):
+if TYPE_CHECKING:
```
danieleades, line 16: is there a circular import if you do not guard the import of `MockerFixture`?

.pre-commit-config.yaml
finswimmer: I thought that too. But it looks like when running `flake8` via `pre-commit` the exclusions in `.flake8` are ignored :man_shrugging: Would be happy to see only one single source of truth.

danieleades (changes requested):
mypy itself can be configured to return errors on missing annotations, i believe. is the 'flake8' plugin somehow preferable?

- `if TYPE_CHECKING` pattern doesn't appear to be required throughout
- inconsistent use of `dict`/`Dict`

poetry/core/version/pep440/version.py
finswimmer: I was happy, that PyCharm was doing this for me :laughing: IMO refactoring should be done in another PR, because the purpose of this PR is to enforce type hints in the future.

.pre-commit-config.yaml
neersighted: That was not my experience with the main repo. I suspect what's happening is partial vs. full paths being passed. Try fragments like e.g. `*/tests/fixtures/*/setup.py`. Some experimentation is likely needed.

poetry/core/version/pep440/version.py
neersighted: That is a brutal type signature :)

tests/json/test_poetry_schema.py
danieleades: while we're at it, these `dict` annotations are equivalent to `Dict[Any, Any]`- we should be able to do better than that

tests/testutils.py
```diff
@@ -70,8 +70,8 @@ def subprocess_run(*args, **kwargs):  # type: (str, Any) -> subprocess.Completed
 
 
 def validate_wheel_contents(
-    name, version, path, files=None
-):  # type: (str, str, str, Optional[List[str]]) -> None
+    name: str, version: str, path: str, files: Optional[List[str]] = None
+) -> None:
     dist_info = "{}-{}.dist-info".format(name, version)
     files = files or []
```
danieleades, line 76: does the following work?

```suggestion
    name: str, version: str, path: str, files: [List[str] = []
) -> None:
    dist_info = "{}-{}.dist-info".format(name, version)
```

finswimmer: [Python Mutable Defaults Are The Source of All Evil](https://florimond.dev/en/posts/2018/08/python-mutable-defaults-are-the-source-of-all-evil/) :smiley:

danieleades: > [Python Mutable Defaults Are The Source of All Evil](https://florimond.dev/en/posts/2018/08/python-mutable-defaults-are-the-source-of-all-evil/) smiley

ah. yep.

.pre-commit-config.yaml
finswimmer: Ok, I've got it now. I removed the exclusion definitions completely from from yml file and make the `.flake8` to the singe source if truth.

tests/json/test_poetry_schema.py
finswimmer: As you find out `dict` is valid as well, but don't let us define how keys and values look like in python<3.9. I agree it's in inconsistent use if we use `Dict` in other places, so I changed it.

> while we're at it, these dict annotations are equivalent to Dict[Any, Any]- we should be able to do better than that

The only way to do it better here is to use `Dict[str, Any]` (which doesn't add much benefit) or use `TypedDict`. But this is only available for python >=3.8 and would require to add `typing-extensions` as dependency for python <3.8. Not sure if we want to introduce another dependency.

poetry/core/version/pep440/parser.py
```diff
@@ -61,7 +61,9 @@ def _get_local(cls, match: Optional[Match[AnyStr]]) -> Optional[LocalSegmentType
         )
 
     @classmethod
-    def parse(cls, value: str, version_class: Optional[Type["PEP440Version"]] = None):
+    def parse(
+        cls, value: str, version_class: Optional[Type["PEP440Version"]] = None
+    ) -> "PEP440Version":
```
danieleades, line 66: if you remove the `if TYPE_CHECKING` guard, this can be written as-

```suggestion
    def parse(
        cls, value: str, version_class: Optional[PEP440Version] = None
    ) -> PEP440Version:
```

tests/conftest.py
finswimmer: That's a convention I'm used to. If you need a module just for type hints, import it only in the `if TYPE_CHECKING` block. 

It's easier to teach to other people ("Why are you sometime put import in that block and sometimes not?") and the overall chance that you have a circular import if you need this import just for type hint is quite big.

TL;DR: I prefer to leave it as it is.

poetry/core/version/pep440/version.py
finswimmer: I've tried to understand the code and IMO the signature can be written more easily. See code change.

tests/conftest.py
danieleades: fair enough. there's some interesting background on it here - https://github.com/larryhastings/co_annotations/issues/1

in general, it seems that `if TYPE_CHECKING` is used iff

- you would otherwise have a circular import
- the import is 'expensive'

the Benevolent Dictator for Life seems to regard it is a painfully necessary hack.

that said, it's also clear from the thread that it's far from settled. And using it *everywhere* instead of the small number of cases where you need it (do you definitely need it anywhere in this codebase?) is at least consistent.

danieleades:
might be a good idea to merge #199 before this is merged, in order to catch any type errors that could be interested by this PR

neersighted:
re: `is TYPE_CHECKING`, check out https://github.com/asottile/flake8-typing-imports

Looks like a useful set of lints we could apply here.

neersighted (approved):

PR #235, opened 2021-11-15
Fix ci and revert #118

abn:
Reverting #234 because this was an incorrect change as the change meant the wrong version of `poetry-core` was used for testing against downstream poetry.

Reverting #118 as this is a breaking change. The change needs to be reworked to support existing downstream usage. Alternatively, a change to poetry first to support list of files can also be made.


abn:
fyi: @neersighted @wagnerluis1982

wagnerluis1982:
@abn how should I address the rework? I mean, do you have any suggestions, apart from avoid to rename the attributes?

BTW, do you have any instant messaging channel for discussing?

finswimmer:
> BTW, do you have any instant messaging channel for discussing?

You can join our [discord server] 
(https://discord.com/invite/awxPgve).

abn:
I'm on the Poetry discord. 

As suggestion, I'd say do not rename the attribute. And follow a sequence like this to avoid the failure (think it's only an issue in tests). This is off the top of my head, no real validation done. 

1. Make poetry aware that `package.readme` can be a string or a list. Unsure if there are usages that might have an issue with this.
2. Always listify the readme property in core.
3. Once new release of core is cut and used by poetry, remove "is str or list" checks.

PR #240, opened 2021-11-15
chore(factory): remove type errors

branchvincent:
Fixes mypy errors from `poetry.core.factory`

### DIFF
--- a/poetry/core/factory.py
+++ b/poetry/core/factory.py
@@ -17,6 +17,8 @@ if TYPE_CHECKING:
     from .packages.project_package import ProjectPackage
     from .packages.types import DependencyTypes
     from .poetry import Poetry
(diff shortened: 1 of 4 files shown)


poetry/core/factory.py
```diff
@@ -32,7 +34,7 @@ def create_poetry(
         from .poetry import Poetry
         from .pyproject.toml import PyProjectTOML
 
-        poetry_file = self.locate(cwd)
+        poetry_file = self.locate(cwd or Path.cwd())
```
branchvincent, line 37: perhaps this default should be contained within the `locate` method?

neersighted: Agreed -- let's go ahead and make that change.

neersighted (approved):

PR #245, opened 2021-11-16
remove overzealous mypy config

danieleades:
we might have gone slightly overboard with the mypy config. don't think these two are required

> **python_version**
> 
> Specifies the Python version used to parse and check the target program. The string should be in the format DIGIT.DIGIT – for example 2.7. The default is the version of the Python interpreter used to run mypy.

any reason not to use the default interpreter? I've not seen any syntax compatibility issues in this repo, or anything like that. and since the typings are ignored by the interpreter during normal execution, there's no reason that mypy can't use more recent typing syntax than the codebase itself, right?

>**ignore_missing_imports**

this doesn't seem to hiding any errors right now, so we might as well remove it


pyproject.toml
```diff
@@ -85,9 +85,7 @@ known_first_party = "poetry.core"
 known_third_party = ["poetry.core._vendor"]
 
 [tool.mypy]
-python_version = "3.6"
```
neersighted, line 88: I'd like to keep this, as I understand it will matter for 3.7's `from __future__ import annotations` and similar features. In general with our linting tools I think it's a good idea to call out the minimum Python version.

neersighted (changes requested):

pyproject.toml
danieleades: i don't think this configuration setting sets the minimum version for the syntax, just the interpreter version used to run mypy, which is not *quite* the same thing.

Can you explain to me the interaction with `from __future__ import annotations`? i'm not sure I understand that. it won't impact us until the move to 3.7+, and even though i'm not sure I understand why it's an issue?

My impression was-

> In general with our linting tools I think it's a good idea to call out the minimum Python version

I'm not sure i agree with this. the approach i would usually take is to

1. test with the oldest version I can get away with
2. lint with the newest version I can get away with (without breaking 1.!)

neersighted: Fair enough, looks like I got misled.

neersighted (approved):

PR #246, opened 2021-11-16
use mypy to lint 'tests' directory

danieleades:
what it says on the tin.

is this something we want or are we getting carried away now?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -89,8 +89,8 @@ follow_imports = "silent"
 explicit_package_bases = true
 namespace_packages = true
 mypy_path = "src"
-files = "src"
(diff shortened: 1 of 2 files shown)


neersighted:
I think this is a bit longer off -- long-term it would be good, but for now I think it's tilting and windmills and not that productive. Happy to hear from others, however.

danieleades:
Perhaps a little quixotic

neersighted:
> 
> 
> Perhaps a little quixotic

You know what, we're doing this in `poetry` itself, so I'm personally open to it.

danieleades:
rebased

neersighted (approved):

PR #247, opened 2021-11-16
lint and fix pep8 naming conventions

danieleades:
i'll see what the CI does, but i'm expecting there's a reasonable chance this is a breaking change for downstream

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -17,6 +17,7 @@ repos:
           - flake8-simplify
           - flake8-tidy-imports
           - flake8-use-fstring
+          - pep8-naming
 
   - repo: https://github.com/pre-commit/mirrors-mypy
     rev: v0.910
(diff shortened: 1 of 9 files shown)


neersighted:
This will probably cause some breakage, yeah. It's a bigger codebase, but I think this would be good to try in poetry itself first since it has no consumers outside of plugins (and the plugin API is not yet stable).

See python-poetry/poetry#4776.

danieleades:
> This will probably cause some breakage, yeah. It's a bigger codebase, but I think this would be good to try in poetry itself first since it has no consumers outside of plugins (and the plugin API is not yet stable).
> 
> See [python-poetry/poetry#4776](https://github.com/python-poetry/poetry/issues/4776).

downstream tests passed. wasn't expecting that

danieleades:
grepping the code, it's clear that the Exception class is used downstream. I guess the 'downstream' test case doesn't cover it.

I guess both repos will need to be changed together, and it's a major version bump for `core`

danieleades:
rebased

finswimmer:
We could ignore the `N818` error for now, to introduce `pep-naming` anyway.

Secrus:
Superseded by #767, this PR was too much of a rebase hell.

PR #250, opened 2021-11-17
chore: use fstrings

branchvincent:
Towards https://github.com/python-poetry/poetry/issues/4776

### DIFF
--- a/src/poetry/core/packages/utils/link.py
+++ b/src/poetry/core/packages/utils/link.py
@@ -40,7 +40,7 @@ class Link:
 
     def __str__(self) -> str:
         if self.requires_python:
-            rp = " (requires-python:%s)" % self.requires_python
+            rp = f" (requires-python:{self.requires_python})"
         else:
             rp = ""
         if self.comes_from:
@@ -49,7 +49,7 @@ class Link:
             return str(self.url)
 
     def __repr__(self) -> str:
-        return "<Link %s>" % self
+        return f"<Link {self!s}>"
 
     def __eq__(self, other: Any) -> bool:
         if not isinstance(other, Link):
--- a/.flake8
+++ b/.flake8
@@ -3,6 +3,7 @@ max-line-length = 88
 ignore = E501, E203, W503, ANN101, ANN102
 ban-relative-imports = True
 mypy-init-return = True
+format-greedy = 1
 per-file-ignores =
     __init__.py:F401
     tests/test_*:ANN201
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -14,6 +14,7 @@ repos:
           - flake8-bugbear
           - flake8-comprehensions
           - flake8-tidy-imports
+          - flake8-use-fstring
 
   - repo: https://github.com/pre-commit/mirrors-mypy
     rev: v0.910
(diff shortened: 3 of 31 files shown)


poetry/core/json/__init__.py
```diff
@@ -31,9 +31,7 @@ def validate_object(obj: dict, schema_name: str) -> List[str]:
     for error in validation_errors:
         message = error.message
         if error.path:
-            message = "[{}] {}".format(
-                ".".join(str(x) for x in error.absolute_path), message
-            )
+            message = f"[{'.'.join(str(x) for x in error.absolute_path)}] {message}"
```
danieleades, line 34: i think this is less clear than the original. consider splitting into two lines
```suggestion
            path = '.'.join(str(x) for x in error.absolute_path)
            message = f"[{path}] {message}"
```

poetry/core/masonry/builders/sdist.py
```diff
@@ -123,7 +121,7 @@ def build_setup(self) -> bytes:
         # If we have a build script, use it
         if self._package.build_script:
             after += [
-                "from {} import *".format(self._package.build_script.split(".")[0]),
+                f"from {self._package.build_script.split('.')[0]} import *",
```
danieleades, line 124: as above- this is harder to read than the original

poetry/core/packages/constraints/constraint.py
```diff
@@ -126,6 +126,4 @@ def __hash__(self) -> int:
         return hash((self._operator, self._version))
 
     def __str__(self) -> str:
-        return "{}{}".format(
-            self._operator if self._operator != "==" else "", self._version
-        )
+        return f"{self._operator if self._operator != '==' else ''}{self._version}"
```
danieleades, line 129: as above

poetry/core/packages/dependency_group.py
```diff
@@ -49,6 +49,4 @@ def __eq__(self, other: object) -> bool:
         )
 
     def __repr__(self) -> str:
-        return "{}({}, optional={})".format(
-            self.__class__.__name__, self._name, self._optional
-        )
+        return f"{self.__class__.__name__}({self._name}, optional={self._optional})"
```
danieleades, line 52: as above

danieleades (changes requested):
there's quite a few places where the f-strings are *harder* to read than the original syntax. This occurs when the 'structure' of the original message is obscured by the long, inline code.

I commented on the first few, but there's a lot more places here where i think this holds.

In those cases, I would suggest improving clarity by adding intermediate calculations of the message components, and then using the local variables.

the point of this lint is to make things easier to grok, so we need to be careful about cargo-culting in the lint, and missing the point of why we're using it in the first place

branchvincent:
thanks @danieleades, hopefully this is more readable

danieleades (approved):
Looks good to me!

src/poetry/core/packages/utils/link.py
```diff
@@ -49,7 +49,7 @@ def __str__(self) -> str:
             return str(self.url)
 
     def __repr__(self) -> str:
-        return "<Link %s>" % self
+        return f"<Link {self}>"
```
neersighted, line 52:
```suggestion
        return f"<Link {self!s}>"
```

To keep this code consistent (`%s` implies the same `__str__` but not `__format__` call).

neersighted (changes requested):
LGTM after one tiny correctness fix! Awesome work!

neersighted (approved):

PR #251, opened 2021-11-17
Source layout and implicit namespace package

abn:
This change moves the code base to using src layout. Additionally, this change also converts the `poetry` package to be an implicit namespace package.


.pre-commit-config.yaml
```diff
@@ -3,7 +3,7 @@ repos:
     rev: 21.9b0
     hooks:
       - id: black
-        exclude: ^poetry/core/_vendor
+        exclude: ^src/poetry/core/_vendor
```
branchvincent, line 6: unrelated to this pr, but `^src/poetry/core/_vendor` could be moved to the [top-level `exclude`](https://pre-commit.com/#pre-commit-configyaml---top-level)

abn: ack, we can do that in another PR. I think @neersighted is doing a general cleanup once the config for `poetry` stabilises.

neersighted: I'm personally fine leaving it for now until we start bringing poetry's `.pre-commit-config.yaml` over.

pyproject.toml
```diff
@@ -19,7 +19,7 @@ classifiers = [
 ]
 
 packages = [
-    {include = "poetry"},
+    {include = "poetry", from = "src"},
```
neersighted, line 22: Let's add whitespace inside the curl braces here.

neersighted (changes requested):

neersighted:
I'd also squash all the comments down, and then this LGTM.

abn:
> I'd also squash all the comments down, and then this LGTM.

I squashed the tox change in because I guess it is kinda related to the src layout change :) The namespace package change needs to be separate.

neersighted (approved):

PR #253, opened 2021-11-20
Type checking

dimbleby:
Resolves: python-poetry#

More typechecking.

Things that I am not entirely comfortable with:

- returning empty strings from `VCSDependency.reference()` and `VCSDependency.pretty_constraint()`
- https://github.com/python-poetry/poetry-core/blob/26e978d3b72ff58039ac921524b5ed47cb8b7353/src/poetry/core/packages/constraints/any_constraint.py#L25 looks like nonsense, what we surely want to return here is some sort of negation of other?
- I've thrown in a few exceptions for places in constraints where cases simply weren't handled.  The code would have failed anyway, I've just made it explicit: but still WIBNI these cases _were_ handled


dimbleby:
Looks as though I made mypy settings stricter, at the same time as @danieleades made mypy run on test files.  Hence the failing pipeline: more tests are not clean under the stricter settings.

I think it's more valuable to clamp down on the src than on the tests, so I've resolved by making more exceptions in the tests rather than by reverting the strictness.

dimbleby:
Re my worries in the `constraints`... I wonder whether constraints are even still used?  It looks as though most of this is duplicated in `semver`: is it that `semver` replaces `constraints`?

While I'd rue the work done to make it typecheck, if we can delete all of the `constraints`... let's do it.

src/poetry/core/packages/constraints/any_constraint.py
```diff
@@ -1,33 +1,27 @@
...
 
-    def difference(self, other: "ConstraintTypes") -> "ConstraintTypes":
+    def difference(self, other: "BaseConstraint") -> "BaseConstraint":
         if other.is_any():
             return EmptyConstraint()
 
         return other
 
-    def intersect(self, other: "ConstraintTypes") -> "ConstraintTypes":
+    def intersect(self, other: "BaseConstraint") -> "BaseConstraint":
```
finswimmer, line 21: Hey,

just a quick comment on that. While returning "BaseConstraint" is technical correct, this might lead missing code completion and "Unresolved attribute errors" in the IDE, because the several Constraint Types have different additional attributes. That's the reason why we used "ConstraintTypes" until now. Another solution to this, is to typecast the return values of these methods in the place where they were called.

fin swimmer

dimbleby: Seeing as how you don't know which type of constraint you have in hand, you mustn't rely on any of those completions until you have refined (by casting or isinstance or whatever) - right?

ie it only leads to missing completions in places where completions _should_ be missing

danieleades (changes requested):

src/poetry/core/packages/constraints/__init__.py
```diff
@@ -11,14 +11,9 @@
...
 BASIC_CONSTRAINT = re.compile(r"^(!?==?)?\s*([^\s]+?)\s*$")
-ConstraintTypes = Union[
-    AnyConstraint, Constraint, UnionConstraint, EmptyConstraint, MultiConstraint
-]
 
 
-def parse_constraint(
-    constraints: str,
-) -> Union[AnyConstraint, UnionConstraint, Constraint]:
+def parse_constraint(constraints: str) -> BaseConstraint:
```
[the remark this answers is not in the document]
dimbleby: if you go through the branch with a `constraint.intersect()` in it, then the original type just isn't true.

pyproject.toml
```diff
@@ -85,7 +85,8 @@ known_first_party = "poetry.core"
 known_third_party = ["poetry.core._vendor"]
 
 [tool.mypy]
-follow_imports = "silent"
+strict = true
+ignore_missing_imports = true
```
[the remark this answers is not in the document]
dimbleby: `silent` is the same as the default `normal` for follow-imports, except that it suppresses errors.  So this is unsuppressing errors.

`strict` is kinda self-explanatory, it turns on all the strictness flags.

`ignore-missing-imports` prevents you from getting warnings for importing third-party libraries that don't have type-checking.

ie I've turned up all the strictness flags, but then turned just one of them off again.

- https://mypy.readthedocs.io/en/stable/running_mypy.html#follow-imports
- https://mypy.readthedocs.io/en/stable/command_line.html?highlight=strict#cmdoption-mypy-strict
- https://mypy.readthedocs.io/en/stable/command_line.html?highlight=ignore-missing-imports#cmdoption-mypy-ignore-missing-imports

src/poetry/core/packages/constraints/any_constraint.py
```diff
@@ -1,33 +1,27 @@
...
-    def allows_all(self, other: "ConstraintTypes") -> bool:
+    def allows_all(self, other: "BaseConstraint") -> bool:
         return True
 
-    def allows_any(self, other: "ConstraintTypes") -> bool:
+    def allows_any(self, other: "BaseConstraint") -> bool:
         return True
 
-    def difference(self, other: "ConstraintTypes") -> "ConstraintTypes":
+    def difference(self, other: "BaseConstraint") -> "BaseConstraint":
```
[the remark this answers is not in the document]
dimbleby: Unless we can make the same change on the `BaseConstraint` then this I'd suggest this is somewhere in the region of confusing / not useful!

What you probably have in hand when performing `difference()` is a `BaseConstraint` anyway, so the specific type of the method on the subclass usually won't be interesting anyway.  It just needs to agree with the base type so that the typechecker is happy.

The proposed parameterisation is not true for all base types eg it doesn't make sense for an `AnyConstraint`.  (The current implementation on `AnyConstraint` _does_ satisfy this: but as previously noted, that's surely because it's a bogus implementation!)

pyproject.toml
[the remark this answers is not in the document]
danieleades: sorry, i more meant why did you decide to change these as part of this pull request?

I'm guessing removing `follow_imports=silent" meant you had to add a few more files to the whitelist? I think that setting can cause errors in files that don't contain errors, because you're *importing* problematic typing from other files.

`strict` i'm all for!

as for `ignore-missing-imports`, is there a chance that could suppress warnings about installing typings from a 3rd-party package? for example would it have suppressed the warning suggesting installing `request-types`? Would it be better to allow these errors, and then selectively suppress them in the mypy config file?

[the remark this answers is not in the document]
dimbleby: I don't think I added anything to the whitelist on removing `follow_imports=silent`.  AFAICS the only non-test thing that I have added to the whitelist is `poetry.core.toml`, and that shows the same errors with or without this setting.

mypy is smart about libraries for which it knows that stubs exist (such as requests): 

> If mypy can’t find stubs for a third-party library, and it knows that stubs exist for the library, you will get a message like this:

>    main.py:1: error: Library stubs not installed for "yaml" (or incompatible with Python 3.8)
>    main.py:1: note: Hint: "python3 -m pip install types-PyYAML"
>    main.py:1: note: (or run "mypy --install-types" to install all missing stub packages)

(https://mypy.readthedocs.io/en/stable/running_mypy.html#library-stubs-not-installed)

So I think that's also not a problem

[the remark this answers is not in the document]
dimbleby: having said which I've no particular objection to changing it to
```
[[tool.mypy.overrides]]
module = [
  'jsonschema.*',
  'setuptools.*',
  'tomlkit.*',
]
ignore_missing_imports = true
```
if it would make you happier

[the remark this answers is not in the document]
dimbleby: Oh, as for why did I change it?  Well it just seems to me that if you're going to turn typechecking on you might as well do it properly...

dimbleby:
I'm so convinced that the `AnyConstraint.difference()` method is faulty that I've partially replaced it with raising an exception.  Also pushed a handful of other minor markups following @danieleades comments.

pyproject.toml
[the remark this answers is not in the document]
danieleades: all of that makes sense to me. I don't have any particular issue with the above, just wanted to check your reasoning.

thanks for the explanation

src/poetry/core/json/__init__.py
```diff
@@ -24,7 +26,7 @@ def validate_object(obj: dict, schema_name: str) -> List[str]:
     from jsonschema import Draft7Validator
 
     validator = Draft7Validator(schema)
-    validation_errors = sorted(validator.iter_errors(obj), key=lambda e: e.path)
+    validation_errors = sorted(validator.iter_errors(obj), key=lambda e: e.path)  # type: ignore
```
finswimmer, line 29: Could you add the error code to `# type: ignore`? 

I think we should add the `python-check-blanket-type-ignore` check to `pre-commit` like we have in `poetry`.

src/poetry/core/packages/constraints/empty_constraint.py
```diff
@@ -1,38 +1,35 @@
...
+    def allows_any(self, other: "BaseConstraint") -> bool:
         return True
 
-    def intersect(self, other: "ConstraintTypes") -> "ConstraintTypes":
+    def intersect(self, other: "BaseConstraint") -> "BaseConstraint":
         return other
 
-    def difference(self, other: "ConstraintTypes") -> None:
-        return
+    def difference(self, other: "BaseConstraint") -> "BaseConstraint":
```
finswimmer, line 26: I guess we can be more specific here by setting the return type to `EmptyConstraint`.

src/poetry/core/packages/constraints/multi_constraint.py
```diff
@@ -71,18 +66,20 @@ def allows_any(self, other: "ConstraintTypes") -> bool:
...
 
-    def intersect(self, other: Constraint) -> "MultiConstraint":
-        if isinstance(other, Constraint):
-            constraints = self._constraints
-            if other not in constraints:
-                constraints += (other,)
-            else:
-                constraints = (other,)
+    def intersect(self, other: BaseConstraint) -> "BaseConstraint":
+        if not isinstance(other, Constraint):
```
finswimmer, line 70: If we only allow `Constraint` we should tell it so by the type hint.

src/poetry/core/packages/constraints/union_constraint.py
```diff
@@ -1,45 +1,43 @@
...
         if other.is_empty():
             return False
 
         if other.is_any():
             return True
 
-        if isinstance(other, Constraint):
-            constraints = [other]
-        else:
+        constraints: Tuple[BaseConstraint, ...]
```
finswimmer, line 36: Can you explain why this is needed here?

src/poetry/core/packages/constraints/union_constraint.py
```diff
@@ -48,17 +46,18 @@ def allows_any(self, other: "ConstraintTypes") -> bool:
...
         if other.is_any():
             return False
 
         if other.is_empty():
             return True
 
-        if isinstance(other, Constraint):
-            constraints = [other]
-        else:
+        constraints: Tuple[BaseConstraint, ...]
```
finswimmer, line 56: Same question as above.

src/poetry/core/packages/constraints/union_constraint.py
```diff
@@ -99,22 +100,22 @@ def intersect(self, other: "ConstraintTypes") -> "ConstraintTypes":
...
         return UnionConstraint(*new_constraints)
 
-    def union(self, other: Constraint) -> "UnionConstraint":
-        if isinstance(other, Constraint):
-            constraints = self._constraints
-            if other not in self._constraints:
-                constraints += (other,)
+    def union(self, other: BaseConstraint) -> "UnionConstraint":
+        if not isinstance(other, Constraint):
+            raise ValueError("Unimplemented constraint union")
```
finswimmer, line 105: If we only allow Constraint we should tell it so by the type hint.

src/poetry/core/spdx/license.py
```diff
@@ -3,6 +3,10 @@
 
 
 class License(namedtuple("License", "id name is_osi_approved is_deprecated")):
+    id: str
+    name: str
+    is_osi_approved: bool
+    is_deprecated: bool
```
finswimmer, line 9: What do you think about using [typing.NamedTuple](https://docs.python.org/3/library/typing.html#typing.NamedTuple) here?

src/poetry/core/utils/toml_file.py
```diff
@@ -15,4 +16,4 @@ def __new__(cls, *args: Any, **kwargs: Any) -> TOMLFile:
             category=DeprecationWarning,
             stacklevel=2,
         )
-        return super().__new__(cls)
+        return super().__new__(cls)  # type: ignore
```
finswimmer, line 19: See comment above about adding error code to `# type: ignore`

finswimmer (changes requested):
Looks quite good to me :+1: Just some minor suggestions/questions.

src/poetry/core/json/__init__.py
dimbleby: I'll add the error code

src/poetry/core/packages/constraints/empty_constraint.py
dimbleby: We could, but that would be different from the signature on the BaseConstraint.  I prefer to keep the signatures in the various constraints consistent; but if you have strong opinions that it's useful to try and make them more precise in the variousl cases where this is possible... well, maybe another time?

src/poetry/core/packages/constraints/multi_constraint.py
dimbleby: But then this signature would definitely be inconsistent with the parent signature, which mypy would rightly complain about.

It is logically allowed to provide a `Constraint` here, it's just that code to support this case has not been written.  (And, presumably, it does not arise in practice.)

src/poetry/core/packages/constraints/union_constraint.py
dimbleby: seems to be not needed, I expect I refactored at some point and missed that this could be removed.

src/poetry/core/packages/constraints/union_constraint.py
dimbleby: and same answer

src/poetry/core/packages/constraints/union_constraint.py
dimbleby: As elsewhere: this would be inconsistent with the base type which mypy would rightly object to.  Again, a `Constraint` is logically allowed here: but we have no implementation for that case.

src/poetry/core/spdx/license.py
dimbleby: No view!  Probably would be fine.

src/poetry/core/utils/toml_file.py
dimbleby: ack

src/poetry/core/packages/constraints/empty_constraint.py
dimbleby: cf previous discussion https://github.com/python-poetry/poetry-core/pull/253/files#r757057577

dimbleby:
re-reading some of the code this evening, I see that various implementations on the EmptyConstraint are nonsense. I've pushed a commit that makes them sensible.

Implementation for allows_all() is a bit non-obvious: does an empty constraint allow all the things in another empty constraint? I've gone yes, for consistency with the semver.EmptyConstraint.

(The existence of these parallel implementations is I suppose a sign of something gone wrong, but that's more than I want to take on for the moment...)

src/poetry/core/spdx/license.py
finswimmer: Would be fine if `typing.NamedTuple` will then be used.

finswimmer (changes requested):

pyproject.toml
```diff
@@ -102,29 +110,27 @@ exclude = "(?x)(^tests/.*/fixtures | ^src/poetry/core/_vendor)"
...
   'poetry.core.packages.project_package',
-  'poetry.core.packages.url_dependency',
-  'poetry.core.packages.vcs_dependency',
-  'poetry.core.pyproject.*',
   'poetry.core.semver.*',
+  'poetry.core.toml.*',
   'poetry.core.version.*',
   # test modules
+  'tests.conftest',
+  'tests.integration/test_pep517',
```
danieleades, line 122: should this be 
```suggestion
  'tests.integration.test_pep517',
```
?

danieleades:
this looks really close! what's left to get it over the line?

pyproject.toml
dimbleby: Either seems to work fine, though I agree that a dot would be consistent with all the others.

This line is moved, rather than introduced by this MR.

dimbleby:
> this looks really close! what's left to get it over the line?

Nothing that I know of.  If I've missed a change request somewhere in the thread, please let me know.

finswimmer:
> > this looks really close! what's left to get it over the line?
> 
> Nothing that I know of.  If I've missed a change request somewhere in the thread, please let me know.

Beside the suggestion by @danieleades there is only my wish about using `typing. NamedTuple` left (if you haven't a strong opinion about it).

dimbleby:
Ah, I hadn't realised you were actually asking for `NamedTuple`.

Well it turns out that it's not completely straightforward in this case: the `CLASSIFIER_SUPPORTED` and `CLASSIFIER_NAMES` don't fit the `NamedTuple` pattern, so some other refactoring would be needed.

```
Invalid statement in NamedTuple definition; expected "field_name: field_type [= default]"
```

So I propose not to do that.

Happy to change the odd-slash-out in the mypy exceptions.

finswimmer (dismissed):
LGTM :+1: 

Thanks a lot for your patience :)

finswimmer:
Downstream tests are now running fine. But now there's a conflicting file :disappointed:  Could you please rebase once more @dimbleby ? :pray:

dimbleby:
Sigh, done.  I would be glad to see this merged!

finswimmer (approved):

PR #255, opened 2021-11-28
special case union of single marker with inverse

dimbleby:
To avoid ending up with constraints like:

bar==7.8.9; platform_system != 'Windows' or platform_system == 'Windows'

Resolves: python-poetry#

Per comment at https://github.com/python-poetry/poetry/pull/4824#issuecomment-981058498.

I have verified that this has the desired effect in the testcase added at https://github.com/python-poetry/poetry/pull/4686

### DIFF
--- a/src/poetry/core/version/markers.py
+++ b/src/poetry/core/version/markers.py
@@ -284,6 +284,9 @@ class SingleMarker(BaseMarker):
             if self == other:
                 return self
 
+            if self == other.invert():
+                return AnyMarker()
+
             return MarkerUnion.of(self, other)
 
         return other.union(self)
--- a/tests/version/test_markers.py
+++ b/tests/version/test_markers.py
@@ -212,6 +212,12 @@ def test_single_marker_union_with_union_duplicate():
     assert str(union) == 'sys_platform == "darwin" or python_version <= "3.6"'
 
 
+def test_single_marker_union_with_inverse():
+    m = parse_marker('sys_platform == "darwin"')
+    union = m.union(parse_marker('sys_platform != "darwin"'))
+    assert union.is_any()
+
+
 def test_multi_marker():
     m = parse_marker('sys_platform == "darwin" and implementation_name == "cpython"')
 


dimbleby:
Is "Poetry Downstream Tests" new?  The first failure I think is a consequence of af08f1ce720da467c9bf3d43eed3d9ebaf4ad7fb; and the second failure is of course the point of this fix.

neersighted:
The Downstream tests are not new per-se. I think what we should do is 1. PR a fix for the `_pretty_constraint` fix, and 2. merge the exporter plugin, then merge this fix (making any needed changes to the plugin tests after).

dimbleby:
re-reading some of the code this evening, I see that various implementations on the EmptyConstraint are nonsense.  I'll push a commit that makes them sensible.

Implementation for `allows_all()` is a bit non-obvious: does an empty constraint allow all the things in another empty constraint?  I've gone yes, for consistency with the `semver.EmptyConstraint`.  

(The existence of these parallel implementations is I suppose a sign of something gone wrong, but that's more than I want to take on for the moment...)

src/poetry/core/version/markers.py
```diff
@@ -284,6 +284,9 @@ def union(self, other: MarkerTypes) -> MarkerTypes:
             if self == other:
                 return self
 
+            if self == other.invert():
+                return AnyMarker()
```
neersighted, line 288: Is `AnyMarker` really correct here, or would `EmptyMarker` be more appropriate?

dimbleby: That feels backwards to me, but perhaps I am mistaken about how this works.  Isn't an empty marker intended to represent a marker that allows nothing? ie to my understanding the empty marker would be the correct value for the intersection of a marker and its inverse, not to the union of those things.

However I am certainly unsure about the intended relationship between

- `poetry.core.packages.constraints.Constraint`
- `poetry.core.semver.version_constraint.VersionConstraint`
- `poetry.core.semver.version.Version`
- `poetry.core.version.BaseMarker`

some of which seem to have overlapping implementations and purposes.  Not to mention that much of the time the constraint-handling code deals with plain strings and none of the above anyway.  So it's _very_ possible that I'm wrong.  

Perhaps it would be a 'fun' project for someone to either write up the design or tidy this all up...

TLDR: I think what I've done is right but I'm more than willing to be corrected.

radoering: I think @dimbleby is right: The intersection of a marker and its inverse is empty, but the union of a marker and its inverse is "any" (AnyMarker stands for `"*"`).

radoering (approved):

neersighted (approved):

PR #258, opened 2021-12-21
address sonarqube 'code smells'

danieleades:


src/poetry/core/packages/constraints/base_constraint.py
```diff
@@ -6,23 +6,23 @@
 
 
 class BaseConstraint:
-    def allows(self, other: "ConstraintTypes") -> bool:
-        raise NotImplementedError()
+    def allows(self, _other: "ConstraintTypes") -> bool:
+        return NotImplemented
```
dimbleby, line 10: What's the reason for this change?  I don't think it's right.   Compare https://docs.python.org/3/library/exceptions.html#NotImplementedError and https://docs.python.org/3/library/constants.html#NotImplemented

>  In user defined base classes, abstract methods should raise this exception when they require derived classes to override the method ...

vs

> A special value which should be returned by the binary special methods (e.g. __eq__(), __lt__(), __add__(), __rsub__(), etc.) to indicate that the operation is not implemented with respect to the other type ...

danieleades: good catch. either a false positive from sonarqube, or an overly-enthusiastic edit on my part. probably the latter.

Secrus:
@danieleades are you coming back to this PR? If not, please close it.

PR #261, opened 2022-01-02
fix populate python constraints from markers

maciejskorski:
Resolves: python-poetry#

### DIFF
--- a/src/poetry/core/factory.py
+++ b/src/poetry/core/factory.py
@@ -247,6 +247,7 @@ class Factory:
         from poetry.core.semver.helpers import parse_constraint
         from poetry.core.version.markers import AnyMarker
         from poetry.core.version.markers import parse_marker
+        from poetry.core.packages.utils.utils import get_python_constraint_from_marker
 
         if groups is None:
             groups = ["default"]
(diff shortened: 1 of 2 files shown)


src/poetry/core/factory.py
```diff
@@ -366,6 +367,10 @@ def create_dependency(
                     )
             else:
                 marker = parse_marker(markers)
+                # possibility that python constraint appears in markers, not keywords
+                if not python_versions and markers:
```
dimbleby, line 371: `markers` is known truth-y here, isn't it, making this check redundant?

Having said which I wonder whether it would be ever so slightly clearer to do this up at line 260-ish, immediately after
```python
            python_versions = constraint.get("python")
            platform = constraint.get("platform")
            markers = constraint.get("markers")
```
(then the check would not be redundant)

src/poetry/core/factory.py
```diff
@@ -366,6 +367,10 @@ def create_dependency(
...
+                    python_versions_marker = get_python_constraint_from_marker(marker)
```
dimbleby, line 372: why not just `python_versions = get_python_constraint_from_marker(marker)`?

src/poetry/core/factory.py
```diff
@@ -366,6 +367,10 @@ def create_dependency(
...
+                    python_versions = python_versions_marker
```
dimbleby, line 373: The result of this parse is never used, is it?

The test that you have added passes without the fix, which is consistent with this not doing anything.  

Can you add a test that fails before the fix, and passes after it?

radoering:
Should be fixed by #284 and/or #324. If there's still an issue feel free to reopen.

PR #265, opened 2022-01-22
CI (python matrix): Update to 3.10

tony:
Resolves: https://github.com/python-poetry/poetry/issues/5087

- [x] Added **tests** for changed code. N/A
- [x] Updated **documentation** for changed code. N/A

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -10,12 +10,7 @@ jobs:
     strategy:
       matrix:
         os: [Ubuntu, MacOS, Windows]
-        python-version: [3.7, 3.8, 3.9]
-        experimental: [false]
-        include:
-          - os: Ubuntu
-            python-version: "3.10.0-alpha - 3.10.0"
-            experimental: true
+        python-version: [3.7, 3.8, 3.9, 3.10]
     steps:
       - uses: actions/checkout@v2
 
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -6,7 +6,6 @@ jobs:
   Tests:
     name: ${{ matrix.os }} / ${{ matrix.python-version }}
     runs-on: ubuntu-latest
-    continue-on-error: ${{ matrix.experimental }}
     strategy:
       matrix:
         os: [Ubuntu, MacOS, Windows]
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -10,7 +10,7 @@ jobs:
     strategy:
       matrix:
         os: [Ubuntu, MacOS, Windows]
-        python-version: [3.7, 3.8, 3.9, 3.10]
+        python-version: [3.7, 3.8, 3.9, "3.10"]
     steps:
       - uses: actions/checkout@v2
 
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       matrix:
         os: [Ubuntu, MacOS, Windows]
-        python-version: [3.7, 3.8, 3.9, 3.10]
+        python-version: [3.7, 3.8, 3.9, "3.10"]
         experimental: [false]
         bootstrap-args: [""]
         include:
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       matrix:
         os: [Ubuntu, MacOS, Windows]
-        python-version: [3.7, 3.8, 3.9]
+        python-version: [3.7, 3.8, 3.9, 3.10]
         experimental: [false]
         bootstrap-args: [""]
         include:
@@ -21,7 +21,7 @@ jobs:
             python-version: pypy3
             experimental: false
           - os: Ubuntu
-            python-version: "3.10.0-alpha - 3.10.0"
+            python-version: "3.10"
             experimental: true
             bootstrap-args: "--git https://github.com/python-poetry/poetry.git"
       fail-fast: false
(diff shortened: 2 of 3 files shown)


.github/workflows/tests.yml
```diff
@@ -13,15 +13,15 @@ jobs:
...
         experimental: [false]
         bootstrap-args: [""]
         include:
           - os: Ubuntu
             python-version: pypy3
             experimental: false
           - os: Ubuntu
-            python-version: "3.10.0-alpha - 3.10.0"
+            python-version: "3.10"
             experimental: true
```
danieleades, line 25: why is this marked `experimental` in the 'tests' workflow, but not the 'integration' workflow?

danieleades (changes requested):

.github/workflows/tests.yml
tony: Unlike the integration test, this points to master poetry git repo. Here's the block in full from `tests.yml`:

```yaml
          - os: Ubuntu
            python-version: "3.10"
            experimental: true
            bootstrap-args: "--git https://github.com/python-poetry/poetry.git"
```

So the question may be: Why doesn't tox also check against git master? I propose that could be better to tackle in a separate issue, but we can also do it here if there's an agreement we could come to.

`workflow/integration.yml`:

[Integration test action](https://github.com/python-poetry/poetry-core/blob/25f6ded/.github/workflows/integration.yml#L31-L33) ([tox.ini](https://github.com/python-poetry/poetry-core/blob/25f6dedcd142148daf3e2f4f495d30055715ddd7/tox.ini)) run [tests/integration/](https://github.com/python-poetry/poetry-core/tree/c6dd2487947a450b030f9649c931964ac4e4a83a/tests/integration).

The reason it had `--experimental` before may be because 3.10 support itself wasn't stable.

danieleades: that makes sense. might be good to raise a separate ticket for discussion about the potential inconsistency

danieleades (dismissed):
looks ok to me. I don't have perms to trigger the CI i'm afraid

tony:
@danieleades 

> @tony tony dismissed danieleades’s stale review via e115389 now

Same, just a rebase

finswimmer:
@tony Could you please rebase :pray:

.github/workflows/integration.yml
```diff
@@ -10,12 +10,7 @@ jobs:
     strategy:
       matrix:
         os: [Ubuntu, MacOS, Windows]
-        python-version: [3.7, 3.8, 3.9]
-        experimental: [false]
-        include:
-          - os: Ubuntu
-            python-version: "3.10.0-alpha - 3.10.0"
-            experimental: true
+        python-version: [3.7, 3.8, 3.9, 3.10]
```
branchvincent, line 13: just needs to be quoted (else it's interpreted as the float `3.1`):
```suggestion
        python-version: [3.7, 3.8, 3.9, "3.10"]
```

.github/workflows/tests.yml
```diff
@@ -13,15 +13,15 @@ jobs:
     strategy:
       matrix:
         os: [Ubuntu, MacOS, Windows]
-        python-version: [3.7, 3.8, 3.9]
+        python-version: [3.7, 3.8, 3.9, 3.10]
```
branchvincent, line 16:
```suggestion
        python-version: [3.7, 3.8, 3.9, "3.10"]
```

.github/workflows/integration.yml
tony: Good catch!

.github/workflows/tests.yml
tony: [763240f](https://github.com/python-poetry/poetry-core/pull/265/commits/763240f96c80432f012c705234b86ce2a9f7061f)

.github/workflows/integration.yml
```diff
@@ -10,12 +10,7 @@ jobs:
     strategy:
       matrix:
         os: [Ubuntu, MacOS, Windows]
-        python-version: [3.7, 3.8, 3.9]
-        experimental: [false]
-        include:
-          - os: Ubuntu
-            python-version: "3.10.0-alpha - 3.10.0"
-            experimental: true
+        python-version: [3.7, 3.8, 3.9, "3.10"]
```
branchvincent, line 13: seems everything is passing but let's keep `experimental` as it's referenced above:
```suggestion
        python-version: [3.7, 3.8, 3.9, "3.10"]
        experimental: [false]
```

tony: I think we should remove `continue-on-error: ${{ matrix.experimental }}` or keep it as is.

The reason why is python 3.10 is stable now and considered officially supported in poetry.

I think experimental would be for either or both of these conditions:
- Alpha python release, e.g. 3.11 alpha
- python release unsupported by poetry, e.g. if 3.10 wasn't considered ready for poetry despite python release 3.10.0 officially
- Bootstrapping poetry trunk, e.g. `bootstrap-args: "--git https://github.com/python-poetry/poetry.git"` (but `tox.ini` doesn't support that and it may be best to defer that to a separate PR)

branchvincent: agreed, let's remove `continue-on-error: ${{ matrix.experimental }}` for now. 

We'll prob add it back to test 3.11 alpha/beta in the future like you said, but we can handle that then

tony: [c3bc556](https://github.com/python-poetry/poetry-core/pull/265/commits/c3bc556a0e7b683553f3a824a13077fcb644322e)

branchvincent (approved):
@tony thanks again! LGTM but like last time, we'll have to wait for another maintainer to merge

tony:
@branchvincent Thank you for the review!

I will sit tight and rebase as needed to make sure the PR stays fresh 👍

finswimmer (approved):
Thanks a lot :+1:

tony:
@finswimmer always.

PR #267, opened 2022-01-24
Use flake8-simplify

radoering:
Relates-to: python-poetry/poetry#4776

flake8-simplify has already been introduced for poetry in python-poetry/poetry#4798

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,6 +1,6 @@
 [flake8]
 max-line-length = 88
-ignore = E501, E203, W503, ANN101, ANN102
+ignore = E501, E203, W503, ANN101, ANN102, SIM106
 ban-relative-imports = True
 mypy-init-return = True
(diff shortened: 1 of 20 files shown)


.flake8
```diff
@@ -1,6 +1,6 @@
 [flake8]
 max-line-length = 88
-ignore = E501, E203, W503, ANN101, ANN102
+ignore = E501, E203, W503, ANN101, ANN102, SIM106
```
finswimmer, line 3: I see that we ignore `SIM106` in `poetry` as well. So I'm fine here. In many cases this warning helps to improve the code. But I've no idea how often it is not helpful in our cases. Maybe we should evaluate it some point later, and only ignore it case-by-case then.

finswimmer (approved):
Beautiful :+1:

PR #269, opened 2022-01-24
Add `py.typed` files 

NiklasRosenstein:
Fixes https://github.com/python-poetry/poetry/issues/5099

This PR adds a `py.typed` file to mark the `poetry-core` source code as containing type hints.

Having poetry-core typed allows Poetry plugins to be more thoroughly type checked with Mypy.


finswimmer:
Hello @NiklasRosenstein,

thanks a lot for your contribution :+1: 

We would like to wait for this until we thoroughly type checked `poetry-core` itself with `mypy` to give you the best experience with it. :smiley: 

fin swimmer

NiklasRosenstein:
Hey @finswimmer,

Don't you think that having _some wrong_ types is better than no types to check against?

It also opens up the opportunity for consumers for poetry-core to detect typing issues; see #270

dimbleby:
I think on the whole this would be better in than not in

- we probably have made progress on the typechecking in this project since this MR was originally raised
- most remaining errors are probably mostly on internal methods and won't harm users of this library
- but anyway surely the only user of this library that counts is `poetry` itself, and if `poetry` would find it helpful to have typing here then that's the main thing

And if there are still mistakes then having a py.typed will encourage putting them right.

We could be still further ahead on the typechecking if the MRs that helped to progress this would be merged.  https://github.com/python-poetry/poetry/issues/5017 asks for community help, but poetry is slow to accept that help when it arrives.  I added a list of open MRs that I know about at https://github.com/python-poetry/poetry/issues/5017#issuecomment-1110052823.

radoering:
I suppose this can be closed because `py.typed` has already sneaked in via #331. 😉

dimbleby:
> py.typed has already sneaked in via https://github.com/python-poetry/poetry-core/pull/331.

I just noticed that myself!  This was an accident, I'd been trying the effect of adding this and... well, I guess an empty file is an easy thing to miss in a diff.

Anyway, I came here to say that it had gone in accidentally but I reckon that it should stay anyway, happy to see that this was already the thought.

PR #270, opened 2022-01-24
Update type hints of `PEP440Version` class

NiklasRosenstein:
This change allows Mypy to infer that methods like `parse()` or `replace()` on a subclass of `PEP440Version` return the same type and not falling back to the `PEP440Version` base type.

Example:

```py
from poetry.core.semver.version import Version

version = Version.parse('1.0.0')
next_version = version.next_major()

reveal_type(version)  # poetry.core.semver.version.Version
reveal_type(next_version)  # poetry.core.semver.version.Version  (used to be PEP440Version)
```

Note that poetry-core can only be used with Mypy after merging #269 anyway.

### DIFF
--- a/src/poetry/core/version/pep440/version.py
+++ b/src/poetry/core/version/pep440/version.py
@@ -217,7 +217,7 @@ class PEP440Version:
             epoch=self.epoch, release=self.release, pre=ReleaseTag(RELEASE_PHASE_ALPHA)
         )
 
-    def replace(self: T, **kT: Any) -> "PEP440Version":
+    def replace(self: T, **kwargs: Any) -> "PEP440Version":
         return self.__class__(
             **{
                 **{
--- a/src/poetry/core/version/pep440/version.py
+++ b/src/poetry/core/version/pep440/version.py
@@ -15,6 +15,7 @@ from poetry.core.version.pep440.segments import LocalSegmentType
 from poetry.core.version.pep440.segments import Release
 from poetry.core.version.pep440.segments import ReleaseTag
 
+
 T = TypeVar("T")
 
 # we use the phase "z" to ensure we always sort this after other phases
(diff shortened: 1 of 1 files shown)


src/poetry/core/version/pep440/version.py
```diff
@@ -14,6 +16,8 @@
 from poetry.core.version.pep440.segments import ReleaseTag
 
 
+T = TypeVar("T")
```
dimbleby, line 19: shouldn't this be
```python
T = TypeVar("T", bound="PEP440Version")
```

NiklasRosenstein: Hm maybe; the current version works as well. I think just `T` is fine because `self` will only ever be a `PEP440Version` or one of it's subclasses. But I can understand if what you suggest is preferred (more obvious). I also haven't tried what happens with `Version.next_major('foobar')` (whether Mypy acepts it because `self` is a `str` and that is compatible with the unbounded type variable `T`)

dimbleby: https://mypy.readthedocs.io/en/stable/generics.html#generic-methods-and-generic-self uses the bound

dimbleby:
Pretty annoying to have to implement derived methods just for the sake of the typing, though?

The link is hidden in a now-resolved thread but https://mypy.readthedocs.io/en/stable/generics.html#generic-methods-and-generic-self shows more or less exactly this, albeit with a warning that this is 'experimental' meaning that mypy might say wrong things.

danieleades:
> Pretty annoying to have to implement derived methods just for the sake of the typing, though?

absolutely. I don't have any particular issue with this, just not something i've seen in the wild.
Python really needs something like Rust's `Self` type

NiklasRosenstein:
@danieleades  A `Self` type is coming in Python 3.11 ([PEP 673](https://www.python.org/dev/peps/pep-0673/))

neersighted (approved):

PR #263, opened 2022-01-16
Deprecate Python 3.6

tony:
Resolves: https://github.com/python-poetry/poetry/issues/4997
See also: https://github.com/python-poetry/poetry/pull/5055, https://github.com/python-poetry/install.python-poetry.org/pull/11

As I'm getting acclimated with the code base, this may have some back and forth to it

- [x] Added **tests** for changed code.
- [x] Updated **documentation** for changed code.

Later?
- [ ] tomlkit: Potentially update port of tomlkit to [0.8.0](https://github.com/sdispater/tomlkit/releases/tag/0.8.0), diff since: [0.7.2](https://github.com/sdispater/tomlkit/compare/0.7.2...0.8.0), [0.7.2](https://github.com/sdispater/tomlkit/compare/0.7.2...0.8.0), [0.6.0](https://github.com/sdispater/tomlkit/compare/0.6.0...0.8.0), [0.5.11](https://github.com/sdispater/tomlkit/compare/0.5.11...0.8.0)

PR #271, opened 2022-01-24
make .pre-commit-config.yaml similar to poetry repo

finswimmer:
* sort and update pre-commit repos to be similar to the one used in `poetry`.
* add some more `pre-commit` ids
* add `yesqa`repo
* add `pygrep` repo
* use `experimental_string_processing` parameter for `black

Relates to: https://github.com/python-poetry/poetry/issues/4776

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,72 +1,81 @@
-repos:
-  - repo: https://github.com/psf/black
-    rev: 21.9b0
-    hooks:
-      - id: black
-        exclude: ^src/poetry/core/_vendor
-
-  - repo: https://gitlab.com/pycqa/flake8
-    rev: 3.9.2
-    hooks:
-      - id: flake8
-        additional_dependencies:
-          - flake8-annotations
-          - flake8-bugbear
-          - flake8-comprehensions
-          - flake8-eradicate
-          - flake8-simplify
-          - flake8-tidy-imports
-          - flake8-use-fstring
-
-  - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v0.910
-    hooks:
-      - id: mypy
-        pass_filenames: false
-
-  - repo: https://github.com/pycqa/isort
-    rev: 5.9.3
-    hooks:
-      - id: isort
-        additional_dependencies: [toml]
-        exclude: |
+exclude: |
           (?x)(
-              ^.*/?setup\.py$
+              ^tests/.*/fixtures/.*
               | ^src/poetry/core/_vendor
-              | tests/.*\.pyi$
           )
 
+repos:
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v4.1.0
+    hooks:
+      - id: trailing-whitespace
+      - id: end-of-file-fixer
+      - id: debug-statements
+      - id: check-merge-conflict
+      - id: check-case-conflict
+      - id: check-json
+      - id: check-toml
+      - id: check-yaml
+      - id: pretty-format-json
+        args:
+          - --autofix
+          - --no-ensure-ascii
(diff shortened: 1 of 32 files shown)


.pre-commit-config.yaml
```diff
@@ -1,72 +1,81 @@
...
     hooks:
-      - id: pyupgrade
-        args:
-          - --py36-plus
-        exclude: ^src/poetry/core/_vendor
+      - id: black
 
-  - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v4.0.1
+  - repo: https://gitlab.com/pycqa/flake8
```
radoering, line 71: In poetry, the repo from github is used. Probably, there is no reason to use gitlab, here.

finswimmer: :+1:

src/poetry/core/packages/utils/utils.py
```diff
@@ -32,15 +31,15 @@
...
+    import bz2  # noqa: F401
 
     SUPPORTED_EXTENSIONS += BZ2_EXTENSIONS
 except ImportError:
     pass
 
 try:
     # Only for Python 3.3+
-    import lzma  # noqa
+    import lzma  # noqa: F401
```
danieleades, line 42: is python `<3.3` compat still required here?

if we need it, we should use

```python
if sys.version_info >= (3, 3):
    import lzma
    ....
```

or similar

src/poetry/core/utils/_compat.py
```diff
@@ -12,7 +12,7 @@
 try:
     FileNotFoundError
 except NameError:
-    FileNotFoundError = IOError  # noqa
+    FileNotFoundError = IOError
```
danieleades, line 15: ditto here - can we use `if sys.version_info > ...`?

danieleades (changes requested):

src/poetry/core/packages/utils/utils.py
finswimmer: I would try to avoid logic change to the code within this PR, because its only about formatting. I think this is better handled in https://github.com/python-poetry/poetry-core/pull/263 for example.

src/poetry/core/utils/_compat.py
finswimmer: same as above

src/poetry/core/packages/utils/utils.py
danieleades: ping @tony

src/poetry/core/utils/_compat.py
danieleades: ping @tony

danieleades (approved):

neersighted (approved):

src/poetry/core/utils/_compat.py
tony: In #263 I pushed a change removing this

src/poetry/core/packages/utils/utils.py
tony: `lzma` is optionally built in.  You can compile CPython without lzma 

Could not figure out the flags, Not having the library seems to do it:

```
❯ sudo apt remove liblzma-dev
❯ make distclean
❯ ./configure
❯ make -j8
❯ ./python -V
Python 3.11.0a4+

❯ ./python -c 'import lzma'
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/home/user/projects/c/cpython/Lib/lzma.py", line 27, in <module>
    from _lzma import *
    ^^^^^^^^^^^^^^^^^^^
ModuleNotFoundError: No module named '_lzma'
```

danieleades: that makes sense

PR #272, opened 2022-01-25
de-glob mypy whitelist

danieleades:


neersighted (approved):

neersighted (approved):

src/poetry/core/semver/empty_constraint.py
```diff
@@ -18,19 +17,19 @@ def is_any(self) -> bool:
...
-    def allows_any(self, other: "VersionTypes") -> bool:
+    def allows_any(self, other: VersionConstraint) -> bool:
         return False
 
-    def intersect(self, other: "VersionTypes") -> "EmptyConstraint":
+    def intersect(self, other: VersionConstraint) -> "EmptyConstraint":
         return self
 
-    def union(self, other: "VersionTypes") -> "VersionTypes":
+    def union(self, other: VersionConstraint) -> VersionConstraint:
```
finswimmer, line 29: Maybe use Generic type bound to `VersionConstraint` to say "what goes in, goes out"?

finswimmer (changes requested):

src/poetry/core/semver/empty_constraint.py
danieleades: I'll take a look. Will need to be changed in the base class

danieleades: this *may* have the behaviour you're looking for, but i'm not 100% confident. Would appreciate some more eyes on this

dimbleby: I resisted this sort of thing in #253: and while I may have been wrong then, I can at least be consistent...

IMO the updated version is unnecessary complication, too clever for its own good.  I doubt that there should be code that is manipulating unions between constraints and that relies on anything other than the generic typing.

Eg for this particular case: if I already know that I'm union-ing with an `EmptyConstraint` - it would be strange to perform the union at all.  I could just as well not bother!

That is: the only reason for calling `union()` is that I _don't_ know that I'm dealing with an `EmptyConstraint`, and in that case the extra type information doesn't help.

And this, I think, is the expected use case.

danieleades: i could go either way on this. On one hand i feel like it's needless pedantry. On the other hand, i *like* needless pedantry ;)

finswimmer: I could follow @dimbleby argumentation, that derived classes should not change the method signature. I have missed that we have a derived class here.

I wonder if there is a tool or maybe mypy option that can check whether the signature changed in a derived class :thinking:

dimbleby: mypy checks that the signature on the derived class is compatible with the parent class.  

In #253 we only discussed options where that was not true and mypy would have complained.

This approach is cleverer and I think in some sense technically correct (the best kind of correct).  But my judgement would still be that it's indeed needless pedantry, not valuable enough to justify the trouble.

neersighted (dismissed):

src/poetry/core/semver/empty_constraint.py
finswimmer: > But my judgement would still be that it's indeed needless pedantry, not valuable enough to justify the trouble.

Now I'm confused :smiley: Do you vote for keeping the same signature as the parent class (because it's technical correct) or allow changing it if the return type is obvious (more intuitive at the first sight, but technical not clean)?

dimbleby: I vote for the simpler version, keeping the same signature as the parent class.

My view is that while the more complicated version is (also) correct - I am reasonably confident that mypy would complain if that were not so - the extra information that it brings is not useful; and the extra complexity of the code is therefore not worthwhile.

danieleades: i'm happy with that. i've reverted back to the original

finswimmer (approved):
LGTM :+1:

PR #274, opened 2022-01-29
typechecking poetry.core.masonry

dimbleby:
Resolves: python-poetry#

I've added mypy to the dev-dependencies and made the poetry.core.masonry code mypy-clean.

Slightly more invasive than some of the typechecking MRs we've seen:
- making the various `.build()` methods have the same signature required changes (and I've removed an unused parameter too)
- since mostly this code wants `Path` rather than `str` I've removed some odd-one-out `Union[Path, str]` 
- I've removed some cleverness where a `BuildIncludeFile` could be equal to a `Path` because it just seemed too much.  I've caught one place where this trickery was being exploited, and re-implemented it in a less magical way, but it's possible that I've missed something.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -98,9 +98,6 @@ ignore_missing_imports = true
 [[tool.mypy.overrides]]
 module = [
   # src modules
-  'poetry.core.masonry.builders.builder',
-  'poetry.core.masonry.builders.sdist',
-  'poetry.core.masonry.builders.wheel',
   'poetry.core.packages.utils.utils',
   'poetry.core.packages.dependency',
   'poetry.core.packages.package',
--- a/src/poetry/core/masonry/builders/builder.py
+++ b/src/poetry/core/masonry/builders/builder.py
@@ -32,8 +32,7 @@ class Builder:
     def __init__(
         self,
         poetry: Poetry,
-        ignore_packages_formats: bool = False,
-        executable: Path | str | None = None,
+        executable: Path | None = None,
     ) -> None:
         from poetry.core.masonry.metadata import Metadata
         from poetry.core.masonry.utils.module import Module
@@ -56,12 +55,7 @@ class Builder:
             if not isinstance(formats, list):
                 formats = [formats]
 
-            if (
-                formats
-                and self.format
-                and self.format not in formats
-                and not ignore_packages_formats
-            ):
+            if formats and self.format and self.format not in formats:
                 continue
 
             packages.append(p)
@@ -70,12 +64,7 @@ class Builder:
         for include in self._package.include:
             formats = include.get("format", [])
 
-            if (
-                formats
-                and self.format
-                and self.format not in formats
-                and not ignore_packages_formats
-            ):
+            if formats and self.format and self.format not in formats:
                 continue
 
             includes.append(include)
@@ -93,7 +82,7 @@ class Builder:
     def executable(self) -> Path:
         return self._executable
 
-    def build(self) -> None:
+    def build(self) -> Path:
         raise NotImplementedError()
 
     def find_excluded_files(self, fmt: str | None = None) -> set[str]:
@@ -201,10 +190,6 @@ class Builder:
                     continue
 
                 if file.suffix == ".pyc":
-                    continue
-
-                if file in to_add:
-                    # Skip duplicates
                     continue
 
                 logger.debug(f"Adding: {str(file)}")
@@ -349,6 +334,8 @@ class Builder:
     @classmethod
     def convert_author(cls, author: str) -> dict[str, str]:
         m = AUTHOR_REGEX.match(author)
+        if m is None:
+            raise RuntimeError(f"{author} does not match regex")
 
         name = m.group("name")
         email = m.group("email")
@@ -378,14 +365,11 @@ class BuildIncludeFile:
 
         self.path = self.path.resolve()
 
-    def __eq__(self, other: BuildIncludeFile | Path) -> bool:
-        if hasattr(other, "path"):
-            return self.path == other.path
+    def __eq__(self, other: object) -> bool:
+        if not isinstance(other, BuildIncludeFile):
+            return False
 
-        return self.path == other
-
-    def __ne__(self, other: BuildIncludeFile | Path) -> bool:
-        return not self.__eq__(other)
+        return self.path == other.path
 
     def __hash__(self) -> int:
         return hash(self.path)
--- a/src/poetry/core/masonry/builders/builder.py
+++ b/src/poetry/core/masonry/builders/builder.py
@@ -40,7 +40,7 @@ class Builder:
 
         self._poetry = poetry
         self._package = poetry.package
-        self._path = poetry.file.parent
+        self._path: Path = poetry.file.parent
         self._excluded_files: set[str] | None = None
         self._executable = Path(executable or sys.executable)
 
@@ -92,6 +92,10 @@ class Builder:
     @property
     def executable(self) -> Path:
         return self._executable
+
+    @property
+    def default_target_dir(self) -> Path:
+        return self._path / "dist"
 
     def build(self, target_dir: Path | None) -> Path:
         raise NotImplementedError()
--- a/src/poetry/core/masonry/builders/builder.py
+++ b/src/poetry/core/masonry/builders/builder.py
@@ -93,7 +93,7 @@ class Builder:
     def executable(self) -> Path:
         return self._executable
 
-    def build(self) -> Path:
+    def build(self, target_dir: Path | None) -> Path:
         raise NotImplementedError()
 
     def find_excluded_files(self, fmt: str | None = None) -> set[str]:
--- a/src/poetry/core/masonry/builders/builder.py
+++ b/src/poetry/core/masonry/builders/builder.py
@@ -32,6 +32,7 @@ class Builder:
     def __init__(
         self,
         poetry: Poetry,
+        ignore_packages_formats: bool = False,
         executable: Path | None = None,
     ) -> None:
         from poetry.core.masonry.metadata import Metadata
@@ -55,7 +56,12 @@ class Builder:
             if not isinstance(formats, list):
                 formats = [formats]
 
-            if formats and self.format and self.format not in formats:
+            if (
+                formats
+                and self.format
+                and self.format not in formats
+                and not ignore_packages_formats
+            ):
                 continue
 
             packages.append(p)
@@ -64,7 +70,12 @@ class Builder:
         for include in self._package.include:
             formats = include.get("format", [])
 
-            if formats and self.format and self.format not in formats:
+            if (
+                formats
+                and self.format
+                and self.format not in formats
+                and not ignore_packages_formats
+            ):
                 continue
 
             includes.append(include)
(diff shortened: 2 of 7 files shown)


src/poetry/core/masonry/builders/builder.py
```diff
@@ -206,10 +203,6 @@ def find_files_to_add(self, exclude_build: bool = True) -> Set["BuildIncludeFile
                 if file.suffix == ".pyc":
                     continue
 
-                if file in to_add:
-                    # Skip duplicates
```
dimbleby, line 210: `to_add` is a set, it couldn't contain duplicates if it tried.  Just need to be happy with `__hash__` and `__eq__` on the `BuildIncludeFile`.

pyproject.toml
```diff
@@ -50,6 +50,7 @@ vendoring = {version = "^0.3", python = "^3.8"}
 pep517 = "^0.11.0"
 black = {version = "^21.9b0", markers = "python_version >= '3.6.2' and python_version < '4.0' and implementation_name != 'pypy'"}
 isort = {version = "^5.9.3", python = "^3.6.1"}
+mypy = ">=0.931"
```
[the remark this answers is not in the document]
dimbleby: While I agree WIBNI these were in one place only, I am not sure that I am convinced that removing them from `pyproject.toml` is a good idea, though.

Just as a matter of workflow: after running `poetry install` in this project I'd sure like to be in an environment that has all the dev tools that I'll be using.

But also precommit is suboptimal, certainly in mypy, because it does not run in the correct virtual environment (per https://github.com/python-poetry/poetry/pull/4510#issuecomment-968125631).  

I wonder if the same is true for isort, which also behaves differently according to the environment.

pyproject.toml
```diff
@@ -39,6 +39,7 @@ python = "^3.7"
 importlib-metadata = {version = ">=1.7.0", python = "<3.8"}
 
 [tool.poetry.dev-dependencies]
+mypy = ">=0.931"
```
danieleades, line 42: adding mypy as a poetry dependency seems a separate concern to the other changes in this PR, no?

danieleades: just saw the above discussion around mypy. I agree it's nice to have it in the development environment, but i think perhaps that should be split out of this and into a separate PR so this one can be merged

danieleades (changes requested):
suggest moving the adding of `mypy` to a separate PR

pyproject.toml
dimbleby: If a maintainer would indicate that this is what's preventing merge then I'm willing to make it separate.  

I don't know have any reason to suspect that's the case though, this project generally just seems shy of hitting the merge button.

danieleades: Yeah that's not totally unfair.
My own issue is more semantic, than logistic. They seem like genuinely separate concerns to me, and therefore belong in separate PRs.
Ultimately that's between you and the maintainers

dimbleby:
Ugh, merge conflicts.  So de-motivating.  I don't think this is waiting on anything, can we get it merged please?  @finswimmer?

(I have removed mypy from pyproject.toml in case that was controversial, though if that was a problem it's frustrating that no-one with the commit bit would say so)

src/poetry/core/masonry/builders/builder.py
```diff
@@ -70,12 +65,7 @@ def __init__(
         for include in self._package.include:
             formats = include.get("format", [])
 
-            if (
-                formats
-                and self.format
-                and self.format not in formats
-                and not ignore_packages_formats
-            ):
+            if formats and self.format and self.format not in formats:
```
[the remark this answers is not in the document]
dimbleby: just simplification: no caller ever sets this parameter true.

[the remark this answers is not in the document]
abn: Let's not remove it for that reason. Core is to be treated as a library for now.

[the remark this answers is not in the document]
dimbleby: Are you sure that's the right call?  Of course this can go back; but accruing dead code doesn't seem desirable to me?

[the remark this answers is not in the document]
dimbleby: of course I intended to remove the parameter on the `__init__()` method as well as the implementation, leaving one but not the other makes no sense.

I'm going to push a commit that completes the removal of this parameter, and then ask you to confirm whether you really want it putting back.  

IMO it's better to clean these things up, but I'm happy to defer if maintainer opinion is definitely otherwise.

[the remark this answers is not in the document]
abn: In an ideal world, everything has test coverage, we are working towards better test coverage - however some of the older code paths do not have coverage. This is normal for a project like this. If we go by "remove everything untested", then a lot of poetry must go ;)  And to be fair, simply because core itself or poetry does not set this value today by no means imply that this is not used by a third-party code depending on poetry-core. In particular builders are something that is expected to be extended on.

For now, i think the call to keep the logic as is, is the least worst option. Bear in mind that this PR's scope is to introduce type checking, not alter logic.

I suspect the team would be happy to entertain targetted proposals removing "dead" code. But I would appreciate if we did that in a separate change set so we can evaluate the impact rather than shoehorning it into a type fix change.

I appreciate your opinion on this, and personally I can't really think of a reason why this flag would be valuable - but I can imagine others having a different view. As maintainers, we need to balance these. But better discussed separate to this.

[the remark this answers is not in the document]
dimbleby: I put it back, this isn't important enough to argue about!  Thanks for the discussion, hopefully this is now good to merge.

src/poetry/core/masonry/builders/sdist.py
```diff
@@ -55,10 +56,18 @@ class SdistBuilder(Builder):
...
 
-    def build(self, target_dir: Path | None = None) -> Path:
+    def __init__(
+        self,
+        poetry: Poetry,
+        executable: Path | None = None,
+        target_dir: Path | None = None,
+    ) -> None:
+        super().__init__(poetry, executable=executable)
+        self._target_dir = target_dir
```
abn, line 66: Not sure if this is desired. Iirc build is intended to be called multiple times with different target directories. You can default to cwd if desired though.

dimbleby: The builders were inconsistent about whether they took a target directory on their constructor or on their `build()` method.  I have made them consistently put it on the constructor.

I think the suggestion is that we should jump the other way?  ie remove the parameter from the constructors, put it on the `build()` method?

(There's a default value either way)

abn: I suspect the inconsistency comes from the use of `make_in` in the wheel builder. Happy for it to be unified as consistency here is better, with the following funcitonality maintained.

1. There is a default target directory (which as you said is present either way).
2. the `build()` method can accept target directory.

This can also mean that make_in / make becomes redundant. Will need to check downstream see if they are being used.

dimbleby: I pushed a commit that does what I think we're agreeing - say if I've misunderstood

dimbleby: `make()` and `make_in()` do indeed look redundant, happy to remove them if you want

dimbleby:
@abn are we good to go?  

There was a question about potentially removing `WheelBuilder.make()` and WheelBuilder.make_in()`, if you want that please say so - though even if so, perhaps it would better be done in a separate MR.

src/poetry/core/masonry/builders/builder.py
```diff
@@ -378,14 +376,11 @@ def __init__(
 
         self.path = self.path.resolve()
 
-    def __eq__(self, other: BuildIncludeFile | Path) -> bool:
-        if hasattr(other, "path"):
-            return self.path == other.path
-
-        return self.path == other
+    def __eq__(self, other: object) -> bool:
```
abn, line 379: Why are we disallowing comparison with Path?

src/poetry/core/masonry/builders/sdist.py
```diff
@@ -55,10 +56,19 @@ class SdistBuilder(Builder):
 
     format = "sdist"
 
-    def build(self, target_dir: Path | None = None) -> Path:
+    def __init__(
```
abn, line 59: Is this init required?

src/poetry/core/masonry/builders/wheel.py
```diff
@@ -69,29 +67,31 @@ def make_in(
...
 
-    def build(self) -> None:
+    def build(
+        self,
+        target_dir: Path | None = None,
+    ) -> Path:
         logger.info("Building wheel")
 
-        dist_dir = self._target_dir
+        dist_dir = target_dir or (self._poetry.file.parent / "dist")
```
abn, line 94: I'd consider the fallback target directory a property of the base class.

src/poetry/core/masonry/builders/builder.py
dimbleby: Not disallowing, but always returning False.

Because a BuildIncludeFile is not a Path!

src/poetry/core/masonry/builders/sdist.py
dimbleby: probably not, will remove

src/poetry/core/masonry/builders/wheel.py
dimbleby: can't say I'm too excited either way but have introduced `default_target_dir()`

dimbleby:
@abn are we waiting on anything to get this merged?

src/poetry/core/masonry/builders/builder.py
```diff
@@ -378,14 +380,11 @@ def __init__(
...
+        if not isinstance(other, BuildIncludeFile):
```
abn, line 384: Are we sure we were not relying on this for any contains check anywhere?

abn (commented):
Minor question otherwise good.

src/poetry/core/masonry/builders/builder.py
dimbleby: Pretty sure: the tests are decent here, and use of the BuildIncludeFile is limited to `poetry.core.masonry.builders` so there aren't too many places to check.

I suppose there's always non-zero risk with a change; but for me part of the point of this initial round of typechecking is to find opportunities to clean up the code, and I think the pipelines provide a good level of confidence.

dimbleby: An example of a trap that the previous implementation left for the unwary:

```
>>> foo = Path("a")
>>> bar = BuildIncludeFile(foo, "b")
>>> foo == bar
False
```
(this one is because the `BuildIncludeFile` makes its path absolute)

And here's a (perhaps less likely) example where this `__eq__` implementation gives non-transitive equality:
```python
#!/usr/bin/env python3

from pathlib import Path
from poetry.core.masonry.builders.builder import BuildIncludeFile

class HasPath:
    def __init__(self, path):
        self.path = path

a = Path("/foo/bar")
b = BuildIncludeFile(a, "whatever")
c = HasPath(a)

print(f"a == b ? {a == b}")
print(f"b == c ? {b == c}")
print(f"a == c ? {a == c}")
```
with output
```
a == b ? True
b == c ? True
a == c ? False
```

abn (approved):

PR #276, opened 2022-01-30
Ensure that authors and maintainers are normalized.

canburak:
This is a bugfix from my point of view, but feel free to reject if you feel this is expected.

# Steps to reproduce
1. Insert a decomposed unicode string as an author in `pyproject.toml`.
2. `poetry install`.

Expected outcome is that the installation takes place but instead you get a mismatch on the `AUTHOR_REGEX`.

An alternative way to reproduce is to update `user.name` in `~/.gitconfig` and create a new project with `poetry new`. Since poetry seems to read the author name from the git config, you will have the same issue with the generated `pyproject.toml`.

While this sounds evil or an attempt to break things on purpose, I actually had such a string in my `.gitconfig` but unfortunately I don't remember how it got there, I blame one of the git UIs. 

# Explanation
Unicode can represent characters in one of two ways, combined or decomposed. My surname contains the letter Ç:
* Combined: `LATIN CAPITAL LETTER C WITH CEDILLA`
* Decomposed: `LATIN CAPITAL LETTER C` + `COMBINING CEDILLA`.

When I have the decomposed version, looks like it is a mismatch with the [`AUTHOR_REGEX`](https://github.com/python-poetry/poetry-core/blob/d9ed333ef0c98e185a79e0f98093e7c5c4d817be/src/poetry/core/packages/package.py#L27).

# Fix
I [normalized](https://docs.python.org/3/library/unicodedata.html#unicodedata.normalize) the authors and maintainers which seemed to solve the issue. Note that you will see I've decomposed the `é` in the [`sample_project`](https://github.com/python-poetry/poetry-core/blob/3bdd0cff9a3968f60457e7bea343cf42c69fb91b/tests/fixtures/sample_project/pyproject.toml#L6) but I have not changed the assert in the related test case, thus, the same test case serves my purpose.

Please note that the output is rendered same for the both versions of the strings, thus the diff seems funny.

### DIFF
--- a/src/poetry/core/factory.py
+++ b/src/poetry/core/factory.py
@@ -9,6 +9,7 @@ from typing import Optional
 from typing import Union
 from warnings import warn
 
+from poetry.core.utils.helpers import combine_unicode
 from poetry.core.utils.helpers import readme_content_type
 
 
@@ -76,10 +77,10 @@ class Factory:
         package.root_dir = root
 
         for author in config["authors"]:
-            package.authors.append(author)
+            package.authors.append(combine_unicode(author))
 
         for maintainer in config.get("maintainers", []):
-            package.maintainers.append(maintainer)
+            package.maintainers.append(combine_unicode(maintainer))
 
         package.description = config.get("description", "")
         package.homepage = config.get("homepage")
--- a/src/poetry/core/utils/helpers.py
+++ b/src/poetry/core/utils/helpers.py
@@ -3,6 +3,7 @@ import re
 import shutil
 import stat
 import tempfile
+import unicodedata
 
 from collections.abc import Mapping
 from contextlib import contextmanager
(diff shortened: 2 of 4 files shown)


dimbleby:
I'm not _entirely_ sure, but it looks as though #136 was probably a previoius try at this.  That one seems to have fixed using `regex` instead of `re`, I don't have strong feelings about whether that's a better or worse approach.

Either way it now looks to be languishing in a sea of merge conflicts.

cf https://github.com/python-poetry/poetry/issues/221, I guess.

neersighted (approved):

julienmalard:
@dimbleby Not entirely sure myself, but I think this might not fix #136 as of yet. The problem with that one was a long-standing bug in the `re` module (see [here](https://github.com/lark-parser/lark/issues/590) and [here](https://stackoverflow.com/questions/12746458/python-unicode-regular-expression-matching-failing-with-some-unicode-characters)) mostly regarding South Asian scripts. I'm not sure if normalising unicode will fix these.
Anyways I'd be happy to recreate a pull request for that one if the community is interested, though I'd need a bit of help figuring out how to add `regex` as a dependency to poetry.
Thanks!

PR #282, opened 2022-02-04
Marker simplifications

dimbleby:
Resolves: python-poetry#

Markers get complicated; I've no intention of trying to solve this in full generality but here are a couple of pieces of low-hanging fruit to improve some cases.

In particular the first commit was motivated by a real life project in which `poetry export` gave the marker `python_version >= "3.6" or python_version >= "3.6"`, which looked a bit sad.


radoering (approved):
Looks good to me. 👍 I'm currently on the same track because every marker simplification makes python-poetry/poetry#4695 better - just grasping for some higher hanging fruit. I hope I can reach some. 😉

finswimmer (approved):

PR #284, opened 2022-02-05
recalculate python markers on dependency more often

dimbleby:
Updating a marker in a way that removes python version constraints fails to remove the python version constraint

Not for merging, until there's a fix!  This is a demonstration of "bug 2" per https://github.com/python-poetry/poetry/pull/5154

### DIFF
--- a/src/poetry/core/packages/dependency.py
+++ b/src/poetry/core/packages/dependency.py
@@ -179,6 +179,8 @@ class Dependency(PackageSpecification):
                 for _, extra in or_:
                     self.in_extras.append(extra)
 
+        # Recalculate python versions.
+        self._python_versions = "*"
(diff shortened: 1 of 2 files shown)


dimbleby:
now including a proposed fix

radoering (approved):
Looks good to me. I wonder if it may break some code that exploited the current behavior ([as in the case of the pretty_constraint fix](https://github.com/python-poetry/poetry/issues/4919#issuecomment-1000416680)). Nevertheless, seems to be correct anyway and since all tests are passing, I think it could be merged.

dimbleby:
(rebased to resolve conflicts after other MRs were merged)

radoering (approved):

neersighted (approved):

PR #286, opened 2022-02-06
Marker name is a string, not a lexer token

dimbleby:
Discovered incidentally while working on https://github.com/python-poetry/poetry/pull/5156 - sometimes a marker name was set to a lexer token rather than a string which, for instance, defeats comparisons for equality.

### DIFF
--- a/src/poetry/core/version/markers.py
+++ b/src/poetry/core/version/markers.py
@@ -737,7 +737,9 @@ def _compact_markers(tree_elements: "Tree", tree_prefix: str = "") -> MarkerType
                 )
 
             value = value[1:-1]
-            groups[-1] = MultiMarker.of(groups[-1], SingleMarker(name, f"{op}{value}"))
+            groups[-1] = MultiMarker.of(
+                groups[-1], SingleMarker(str(name), f"{op}{value}")
+            )
         elif token.data == f"{tree_prefix}BOOL_OP" and token.children[0] == "or":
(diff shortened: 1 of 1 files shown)


abn (dismissed):
Can we add some regression coverage for this one?

dimbleby:
I must admit that I'm struggling now to recall why this was a problem.  Per https://lark-parser.readthedocs.io/en/latest/classes.html#lark.Token

> The Token class inherits from Python’s str, so normal string comparisons and operations will work as expected.

which seems like it should be fine, and my attempts to demonstrate otherwise in unit test are failing. 

I'd be inclined to merge this anyway - it can't possibly be wrong, can it?  But I am unable to prove that it is necessary, so if you're feeling super-cautious I guess don't.

Maybe we sometimes end up with different tokens for the same thing?  [This](https://github.com/python-poetry/poetry-core/blob/4e5207e4d41c0c4f1844b1389eef79276af47a49/src/poetry/core/version/grammars/markers.lark#L17-L18) looks like an accident, I wonder whether that could confuse things sometimes?!?

dimbleby:
I'm back on the trail... depending on where the parser is called from, you can end up with (eg) either `Token(markers__MARKER_NAME, 'extra')` or `Token(MARKER_NAME, 'extra')`; and these do not compare equal because they are not the same type of token

dimbleby:
testcase added

PR #287, opened 2022-02-09
Fix vendor patch file

abn:
- Fix jsonschema vendoring patchfile.
- Update pursistent licence file


dimbleby:
cf #275, which has some discussion

abn:
I have dropped the pre-commit dependency commit as this is being discussed in #275

neersighted (approved):

PR #288, opened 2022-02-09
Fix for git subdirectories

ashnair1:
This fix is required for correctly handling subdirectory urls. For example consider the following url

```
https://github.com/user/repo.git@master#subdirectory=mydir
```

Since the regex pattern is looking for `@rev` at the end, it will not strip out the revision in the above url since rev is not at the end of the string resulting in a url like this:

```
https://github.com/user/repo.git@master
```

The correct url should be the following

```
https://github.com/user/repo.git
```


ashnair1:
@finswimmer Any thoughts on this?

ashnair1:
Pinging @abn @finswimmer

ashnair1:
Pinging @abn @finswimmer

ashnair1:
> Hi @ashnair1 sorry about the latency. :)
> 
> I need to check a couple of things downstream before reviewing this. But in the meantime can you add some test coverage?

No problem. I've added the test case. Cheers.

ashnair1:
As mentioned in the companion PR, I've set the default language version for pre-commit to be 3.8 due to flake8-type-checking not supporting Python<3.8

src/poetry/core/vcs/git.py
```diff
@@ -226,7 +226,7 @@ def normalize_url(cls, url: str) -> GitUrl:
 
         formatted = re.sub(r"^git\+", "", url)
         if parsed.rev:
-            formatted = re.sub(rf"[#@]{parsed.rev}$", "", formatted)
+            formatted = re.sub(rf"[#@]{parsed.rev}", "", formatted)
```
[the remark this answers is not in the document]
ashnair1: Won't this only happen if rev=subdirectory? We could just add a guard for it, but I am open to suggestions.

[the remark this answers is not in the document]
ashnair1: @abn Needed a negative lookahead too to address the rev=subdirectory problem. Now it's good. 

```
formatted = re.sub(rf"[#@]{parsed.rev}(?=[#&]?)(?!\=)", "", formatted) 
```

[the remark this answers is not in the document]
ashnair1: Added some additional tests to ensure expression works as expected.

.pre-commit-config.yaml
```diff
@@ -1,3 +1,6 @@
+default_language_version:
```
neersighted, line 1: I don't think this should be necessary at all -- certainly, CI passes with the defaults. The only way I can see this being an issue is if you have `python3.7` as your default interpreter but `python3.8` is available in your environment. Is that possible/common outside of say, an old RHEL that we don't really support as a development environment?

src/poetry/core/packages/specification.py
```diff
@@ -131,6 +131,7 @@ def __hash__(self) -> int:
             ^ hash(self._source_url)
             ^ hash(self._source_reference)
             ^ hash(self._source_resolved_reference)
+            ^ hash(self._source_subdirectory)
```
neersighted, line 134: We should add a test to check that two specs with different subdirectories has differently.

neersighted (changes requested):

.pre-commit-config.yaml
ashnair1: I encountered this on my Ubuntu 20.04.

neersighted: [Ubuntu 20.04 ships with](https://wiki.ubuntu.com/FocalFossa/ReleaseNotes) Python 3.8 as the default -- according to both the release notes and a fresh VM. I'm going to chalk what you ran into as local configuration strangeness (probably some `update-alternatives` in the background), barring a reproduction in a clean VM/container of a currently supported Linux distro.

src/poetry/core/packages/specification.py
ashnair1: Added the test for package equality

.pre-commit-config.yaml
ashnair1: I'm away from my machine so I'll just remove this and open an issue if this persists.

neersighted (approved):

PR #290, opened 2022-02-11
perf: faster blobless git clones

Yajo:
Cloning a repo with git can be really slow if the repo is big.

Adding `--filter=blob:none` to the `git clone` command will make it really faster and slimmer because it will lazy-download blobs on checkout, while only getting commit metadata for whatever is not checked out.

See "Blobless Clones" in https://github.blog/2020-12-21-get-up-to-speed-with-partial-clone-and-shallow-clone/ for a good explanation.

Also interesting to see that pip 21.3 itself landed this enhancement: https://github.com/pypa/pip/pull/9086.

@moduon MT-83

### DIFF
--- a/src/poetry/core/vcs/git.py
+++ b/src/poetry/core/vcs/git.py
@@ -255,10 +255,28 @@ class Git:
     def config(self) -> GitConfig:
         return self._config
 
+    @property
+    def version(self) -> tuple[int, int, int]:
+        output = self.run("version")
+        version = re.search(r"(\d+)\.(\d+)\.(\d+)", output)
+        if not version:
+            return (0, 0, 0)
+        return int(version.group(1)), int(version.group(2)), int(version.group(3))
+
     def clone(self, repository: str, dest: Path) -> str:
         self._check_parameter(repository)
-
-        return self.run("clone", "--recurse-submodules", "--", repository, str(dest))
+        cmd = [
+            "clone",
+            "--filter=blob:none",
+            "--recurse-submodules",
+            "--",
+            repository,
+            str(dest),
+        ]
+        # Blobless clones introduced in Git 2.17
+        if self.version < (2, 17):
+            cmd.remove("--filter=blob:none")
+        return self.run(*cmd)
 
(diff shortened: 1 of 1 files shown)


radoering:
I wonder if even a shallow fetch with depth 1 would be sufficient? This could result in an even bigger performance boost. There is a follow up post with some performance measurements: https://github.blog/2020-12-22-git-clone-a-data-driven-study-on-cloning-behaviors/

In my understanding (which may be wrong), poetry always does a `clone`, followed by a `checkout` of a specific reference. `checkout` is never used several times with different references on an existing repo, is it?

Of course, using shallow fetch would be a more significant change compared to blobless clone. `clone` and `checkout` could not be considered separately anymore. The (simplified) sequence of git commands should probably look like this (in order to maximize the performance benefit):
```
git init
git remote add origin <repo-url>
git fetch --depth 1 origin <ref>
git checkout FETCH_HEAD
```
Probably, shallow fetch should not replace the existing `clone` and `checkout` methods of the `Git` class anyway but be an additional option which could be used by `poetry` but may not be used by other downstream projects. Thus, it could be considered in a separate PR.

Yajo:
It could work. And it could be better. However, that would be more drastic. For example, if poetry kept a real cache, it would be more useful with a blobless clone than with a shallow clone.

This one enhancement is good and means practically no difference in current behavior. Actually, if all tests go green, is enough prove that this won't change anything deeply important.

I think that can be considered as a separate future enhancement, but I don't see that as a blocker for this enhancement.

OTOH, neither of these fix the problem explained in https://github.com/python-poetry/poetry/issues/5188 (quadruplicated clones).

radoering:
Probably, this is not specified anywhere.

Yajo:
Version checked.

neersighted (dismissed):

radoering:
@Yajo Can you please rebase your branch in order to allow merging? Probably, you have to adapt the type hints.

Yajo:
I think it should be good now.

PR #293, opened 2022-02-12
improve detection of Any marker union

dimbleby:
Resolves: python-poetry#

Another trivial simplification of markers


radoering:
Nice. I'm just wondering if "overlapped" is the correct word. That sounds like "the intersection is not empty". However, the variable only becomes True if one marker is a subset of the other, doesn't it? After reading the [introduction about subsets at Wikipedia](https://en.wikipedia.org/wiki/Subset), I think "contained" or "included" may be better suited than "overlapped".

finswimmer (approved):

PR #294, opened 2022-02-13
merging python_version and python_full_version markers

dimbleby:
Resolves: python-poetry#

recognise that `python_version` and `python_full_version` markers say much the same thing and often can be merged

As an example of the difference this makes, the `poetry export` testcase that I have added in https://github.com/python-poetry/poetry/pull/5156 sees the following diff when using this fix:

```diff
- click-didyoumean==0.3.0 ; python_full_version >= "3.6.2" and python_full_version < "4.0.0" and python_version >= "3.7" and python_version < "4.0"
+ click-didyoumean==0.3.0 ; python_version >= "3.7" and python_full_version < "4.0.0"
- click==7.1.2 ; python_version >= "3.6" and python_full_version < "3.0.0" and python_version < "3.7" or python_version >= "3.6" and python_version < "3.7" and python_full_version >= "3.5.0"
+ click==7.1.2 ; python_version >= "3.6" and python_version < "3.7"
- click==8.0.3 ; python_version >= "3.7" and python_version < "4.0" or python_full_version >= "3.6.2" and python_full_version < "4.0.0" and python_version >= "3.7" and python_version < "4.0"
+ click==8.0.3 ; python_version >= "3.7" and python_version < "4.0" or python_version >= "3.7" and python_full_version < "4.0.0"
```

The last click marker is not quite perfect, in ideal world we'd recognise that `python_version < "4.0"` was the same as `python_full_version < "4.0.0"` and then merge things - but it's a substantial improvement.

### DIFF
--- a/src/poetry/core/version/markers.py
+++ b/src/poetry/core/version/markers.py
@@ -49,6 +49,7 @@ ALIASES = {
     "python_implementation": "platform_python_implementation",
 }
 
+PYTHON_VERSION_MARKERS = ["python_version", "python_full_version"]
 
 # Parser: PEP 508 Environment Markers
 _parser = Parser(GRAMMAR_PEP_508_MARKERS, "lalr")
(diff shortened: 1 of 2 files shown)


src/poetry/core/version/markers.py
```diff
@@ -259,21 +260,28 @@ def value(self) -> str:
...
-            new_constraint = self._constraint.intersect(other.constraint)
-            if new_constraint.is_empty():
-                return EmptyMarker()
+                if new_constraint == self.constraint:
+                    return self
 
-            if new_constraint == self._constraint or new_constraint == other.constraint:
-                return SingleMarker(self._name, new_constraint)
+                if new_constraint == other.constraint:
+                    return other
```
radoering, line 284: I've not checked in full detail, but I have the feeling that this part of code can be removed for the most part after your other PRs (#282 and #293) have been merged because it's a duplication of the code in `MultiMarker.of()`. Thus, I'd propose to wait for them, rebase and check if my feeling is right.

dimbleby: I think you're right and this section can become just

```python
if isinstance(other, SingleMarker):
    return MultiMarker.of(self, other)
```

So I agree it makes sense to do others first and perhaps this can be simplified

dimbleby:
rebased following merge of #282 and #293, including simplification per suggestion of https://github.com/python-poetry/poetry-core/pull/294#discussion_r809377375

radoering (approved):

neersighted (approved):

PR #295, opened 2022-02-14
Fix Version.next_devrelease() when starting with a prerelease.

sherbang:
Version('1.2.3-pre.0').next_devrelease() raises an exception, and Version('1.2.3-dev.0').next_devrelease() doesn't change the version.  It seems that this code was probably copied from next_prerelease(), but not completely updated from pre to dev?

The version/expected list is based on the behavior of next_prerelease().  I'm not sure if this is the most "correct" logic here?  I removed passing pre to the created version because turning "1.2.3-alpha.0" into "1.2.3-alpha.0-dev.0" didn't seem right, but I'm not sure that turning it into "1.2.3-dev.0" is right either.

- [X ] Added **tests** for changed code.


abn:
Technically `1.2.3-alpha.0.dev.0` is a pre-release to the pre-release `1.2.3-alpha.0`. But in this case I suspect the right behaviour is to bump to `1.2.3-alpha.1.dev.0`. This means the last set component gets bumped prior to initialising `dev` release. So `1.2.3 -> 1.2.4-dev0`, `1.2.3-post1 -> 1.2.3-post2.dev0` etc.

sherbang:
> Technically `1.2.3-alpha.0.dev.0` is a pre-release to the pre-release `1.2.3-alpha.0`. But in this case I suspect the right behaviour is to bump to `1.2.3-alpha.1.dev.0`. This means the last set component gets bumped prior to initialising `dev` release. So `1.2.3 -> 1.2.4-dev0`, `1.2.3-post1 -> 1.2.3-post2.dev0` etc.

That's an option for sure.  In this case I was just trying to make the logic not obviously broken.

Right now, `Version('1.2.3-dev.0').next_devrelease()` doesn't do anything and that's obviously wrong, with this patch it will return `Version('1.2.3-dev.1')` which is correct.

I consider the `Version('1.2.3-pre.0').next_devrelease() ` behavior to be more of a corner-case.  The alpha+dev type versioning will likely be controversial no matter which approach is chosen and some people will use it and others will just implement their own versioning logic (which I ended up doing anyway because I didn't like how the release tags kept getting expanded).

src/poetry/core/version/pep440/version.py
```diff
@@ -201,13 +201,11 @@ def next_postrelease(self) -> "PEP440Version":
...
     def next_devrelease(self) -> "PEP440Version":
-        if self.is_prerelease():
+        if self.is_devrelease():
             dev = self.dev.next()
         else:
             dev = ReleaseTag(RELEASE_PHASE_DEV)
-        return self.__class__(
-            epoch=self.epoch, release=self.release, pre=self.pre, dev=dev
-        )
+        return self.__class__(epoch=self.epoch, release=self.release, dev=dev)
```
abn, line 208:
```suggestion
        return self.__class__(epoch=self.epoch, release=self.release, pre=self.pre, post=self.post, dev=dev)
```

Let's retain the pre and post release for now. Handling of next pre-release, post-release when dev is set can be addressed in another PR.

tests/semver/test_version.py
```diff
@@ -215,3 +215,16 @@ def test_difference():
...
         v.difference(VersionRange(Version.parse("1.4.0"), Version.parse("3.0.0"))) == v
     )
+
+
+@pytest.mark.parametrize(
+    "version, expected",
+    [
+        ("1.2.3", "1.2.3-dev.0"),
+        ("1.2.3-alpha.0", "1.2.3-dev.0"),
+        ("1.2.3-dev.0", "1.2.3-dev.1"),
```
abn, line 225:
```suggestion
        ("1.2.3", "1.2.3-dev.0"),
        ("1.2.3-alpha.0", "1.2.3-alpha.0-dev.0"),
        ("1.2.3-dev.0", "1.2.3-dev.1"),
```

abn (changes requested):

abn (approved):
Approving this now as it solves the dev release bumps. Pre/post release dev release handling to improved in follow-up PRs.

abn:
Merged via https://github.com/python-poetry/poetry-core/commit/4781c82b2a0326fd344fbb28c721b58f8551b81a

PR #299, opened 2022-02-27
Single marker union simplifications

radoering:
Relates-to: python-poetry/poetry#5192

Allows the following new simplifications:
`sys_platform == "linux" or sys_platform != "win32"` -> `sys_platform != "win32"`
`python_version == "3.6" or python_version > "3.6"` -> `python_version >= "3.6"`
`python_version == "3.6" or python_version < "3.6"` -> `python_version <= "3.6"`
`python_version < "3.6" or python_version > "3.6"` -> `python_version != "3.6"`

### DIFF
--- a/src/poetry/core/version/markers.py
+++ b/src/poetry/core/version/markers.py
@@ -8,6 +8,7 @@ from typing import List
 from typing import Type
 from typing import Union
 
+from poetry.core.semver.version_constraint import VersionConstraint
 from poetry.core.version.grammars import GRAMMAR_PEP_508_MARKERS
 from poetry.core.version.parser import Parser
 
@@ -16,7 +17,6 @@ if TYPE_CHECKING:
     from lark import Tree
 
     from poetry.core.packages.constraints import BaseConstraint
-    from poetry.core.semver.version_constraint import VersionConstraint
 
 
 class InvalidMarker(ValueError):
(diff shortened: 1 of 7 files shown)


src/poetry/core/version/markers.py
```diff
@@ -246,7 +249,7 @@ def constraint_string(self) -> str:
         return self._constraint_string
 
     @property
-    def constraint(self) -> "VersionTypes":
+    def constraint(self) -> Union["BaseConstraint", VersionConstraint]:
```
neersighted, line 252: Just curious what the reason for migrating away form the `VersionTypes` type is -- is it merely a simple, useless typedef? In that case, should we factor it out entirely in the future?

radoering: I tend to say "yes". `VersionTypes` contains every subclass of `VersionConstraint`, but not `VersionConstraint` itself. Everywhere `VersionTypes` is used, you can only be sure to use the interface of `VersionConstraint`, otherwise an isinstance check for a subclass is required.

As for this case: At first, I only added `BaseConstraint` because it can be both (`BaseConstraint` and `VersionTypes`). Due to the fact it can be both, in `MarkerUnion.of()` I had to add `isinstance(union, VersionConstraint)` before calling a method specific to `VersionConstraint`/`VersionTypes`, afterwards calling `SingleMarker()` with this union. Then, mypy tells me that `BaseConstraint` or `VersionTypes` is required but `VersionConstraint` is given...

radoering: The future ticket has overtaken this one. 😉 (VersionTypes has been removed in #302)

neersighted (approved):

PR #300, opened 2022-03-01
Fix PEP508 formatting for subdirectory parameter

marink:
Resolves: python-poetry#[Fix for git subdirectories #288](https://github.com/python-poetry/poetry-core/pull/288)

There is a minor adjustment to fix #288. It was missing the '=' sign when specifying the Git subdirectory to load.


finswimmer (changes requested):
Thanks a lot :+1:

Can you please add a test for it?

aryarm:
any updates on this, @marink? I'm encountering this issue, as well

AKuederle:
@marink @finswimmer is there anything I could do to help getting this merged?

As @finswimmer mentioned, this is missing a test. @marink I think the right place to add one would be here: `https://github.com/python-poetry/poetry-core/blob/main/tests/packages/test_vcs_dependency.py`

And if I understand the structure correctly, a valid test could look something like this:

```
def test_to_pep_508() -> None:
    dependency = VCSDependency(
        "poetry", "git", "https://github.com/python-poetry/poetry.git", directory="subdir"
    )

    expected = "poetry @ git+https://github.com/python-poetry/poetry.git#subdirectory=subdir"

    assert dependency.to_pep_508() == expected
```

aryarm:
I think this PR can be closed, since #288 has been merged and it already implements this fix.

AKuederle:
Actually no. #288 fixed another occurrence of the missing equal sign.

The issue fixed here is still in master: https://github.com/python-poetry/poetry-core/blob/d9ea915208dceae6ad9bb09b04bcfcb82e83f561/src/poetry/core/packages/vcs_dependency.py#L120

AKuederle:
@marink Could you have a look at the example test that I posated a couple of comments up? I think that would be a sufficient test. If you add this we could move this PR forward.

radoering:
It seems this one has been duplicated by #451, which got merged without a test. (Sorry about that.)

Later I added the missing test (and some more) in #453. Thus, nothing to do here.

AKuederle:
Ah perfect! Thanks for handling that!

PR #301, opened 2022-03-01
Replace MarkerTypes by BaseMarker in type hints

radoering:
In my opinion, `MarkerTypes` is an unnecessary type definition that does not do any good.

`MarkerTypes` is a union of all sub classes of `BaseMarker`, but not `BaseMarker` itself.

If a method returns an object of type `MarkerTypes`, you can only rely on the interface of `BaseMarker` or have to do an `isinstance` check anyway.

If a method requires a parameter of type `MarkerTypes`, you can't pass an object whose type has been inferred to be `BaseMarker`, e.g. by calling `isinstance(object, BaseMarker)` before.

Thus, `MarkerTypes` should be replaced by `BaseMarker` in type hints.

### DIFF
--- a/src/poetry/core/factory.py
+++ b/src/poetry/core/factory.py
@@ -18,7 +18,7 @@ if TYPE_CHECKING:
     from poetry.core.packages.types import DependencyTypes
     from poetry.core.poetry import Poetry
     from poetry.core.spdx.license import License
-    from poetry.core.version.markers import MarkerTypes
+    from poetry.core.version.markers import BaseMarker
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


branchvincent (approved):
I agree and concluded the same when I came across it

dimbleby:
Adding my support to the principle: we have previously got rid of
```python
ConstraintTypes = Union[
    AnyConstraint, Constraint, UnionConstraint, EmptyConstraint, MultiConstraint
]
```
for much the same reasons. 

(I have not reviewed the details of this MR)

abn (approved):

PR #302, opened 2022-03-05
Replace VersionTypes by VersionConstraint in type hints

radoering:
Rationale analogous to #301


abn:
@radoering isin't `VersionConstraint` broader than current types in `VersionTypes`?

radoering:
`VersionTypes` contains all concrete sub classes of `VersionConstraint`. `VersionConstraint` is only broader by allowing `VersionConstraint` itself and `VersionRangeConstraint`, which are both abstract base classes.

abn (approved):

PR #307, opened 2022-03-12
Fix special case of get_python_constraint_from_marker() and SingleMarker.only()

radoering:

This PR fixes the util function `get_python_constraint_from_marker()` by fixing `SingleMarker.only()`.

Given a marker not containing `python_version` or `python_full_version` (e.g. `'sys_platform == "linux"'`), any python version should be allowed. Thus, `get_python_constraint_from_marker()` should return `"*"`. Analogously, `only("python_version")` on a `SingleMarker` `'sys_platform == "linux"'` should return `AnyMarker()` instead of `EmptyMarker()` because if `python_version` is not mentioned, any version is fine.

If the last point seems counterintuitive, keep in mind that `""` (empty string) **does not represent** `EmptyMarker` **but** `AnyMarker`.

Rationale: This fix enables further improvements to reduce the number of overrides in poetry's dependency resolution.

### DIFF
--- a/src/poetry/core/version/markers.py
+++ b/src/poetry/core/version/markers.py
@@ -294,9 +294,9 @@ class SingleMarker(BaseMarker):
 
         return self
 
-    def only(self, *marker_names: str) -> Union["SingleMarker", EmptyMarker]:
+    def only(self, *marker_names: str) -> Union["SingleMarker", AnyMarker]:
         if self.name not in marker_names:
-            return EmptyMarker()
+            return AnyMarker()
 
         return self
 
--- a/tests/packages/utils/test_utils.py
+++ b/tests/packages/utils/test_utils.py
@@ -35,6 +35,7 @@ def test_convert_markers():
             'python_full_version >= "3.6.1" and python_full_version < "4.0.0"',
             ">=3.6.1, <4.0.0",
         ),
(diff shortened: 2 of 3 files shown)


src/poetry/core/version/markers.py
```diff
@@ -294,9 +294,9 @@ def exclude(self, marker_name: str) -> BaseMarker:
 
         return self
 
-    def only(self, *marker_names: str) -> Union["SingleMarker", EmptyMarker]:
+    def only(self, *marker_names: str) -> Union["SingleMarker", AnyMarker]:
```
abn, line 297: The commit mentions `get_python_constraint_from_marker()` and `SingleMarker.only()`, is it missing a change?

radoering: `get_python_constraint_from_marker()` is fixed indirectly by fixing `SingleMarker.only()` which is used in `get_python_constraint_from_marker()`:
https://github.com/python-poetry/poetry-core/blob/e8031bbc2d573f94a99b0d0ab4d25f8efcef83f8/src/poetry/core/packages/utils/utils.py#L303

Without the change both tests (for `SingleMarker.only()` and `get_python_constraint_from_marker()`) will fail.

abn (approved):

PR #308, opened 2022-03-13
Improve union of multi markers

radoering:
Relates-to: https://github.com/python-poetry/poetry/issues/5192

Improves the union of multi markers with all kinds of markers. See added test cases for new simplifications that were not possible before.

Improving the union of multi markers is similar but a bit more complicated than #226. That's because both `MarkerUnion.intersect()` and `MarkerUnion.union()` as well as `MultiMarker.union()` prefer to return a `MarkerUnion` of `MultiMarkers` instead of a `MultiMarker` of `MarkerUnions`. Therefore, in addition to the standard `MultiMarker.union()` method a `MultiMarker.union_simplify() `method is introduced. This method prefers to return a `MultiMarker` instead of a `MarkerUnion`, but returns `None` if no simplification is possible in order to avoid an endless recursion.

### DIFF
--- a/src/poetry/core/version/markers.py
+++ b/src/poetry/core/version/markers.py
@@ -460,13 +460,50 @@ class MultiMarker(BaseMarker):
         return MultiMarker.of(*new_markers)
 
     def union(self, other: BaseMarker) -> BaseMarker:
-        if other in self._markers:
-            return other
-
         if isinstance(other, (SingleMarker, MultiMarker)):
             return MarkerUnion.of(self, other)
 
         return other.union(self)
+
+    def union_simplify(self, other: BaseMarker) -> BaseMarker | None:
+        """
+        In contrast to the standard union method, which prefers to return
+        a MarkerUnion of MultiMarkers, this version prefers to return
+        a MultiMarker of MarkerUnions.
+
+        The rationale behind this approach is to find additional simplifications.
+        In order to avoid endless recursions, this method returns None
+        if it cannot find a simplification.
+        """
+        if isinstance(other, SingleMarker):
+            new_markers = []
+            for marker in self._markers:
+                union = marker.union(other)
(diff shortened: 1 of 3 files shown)


radoering:
@dimbleby I wonder if you might review this change? Even if your review doesn't affect mergeability, I think it could increase confidence and speed up the process, since you have already made some changes in that direction and seem to have a good understanding of marker handling.

dimbleby:
This is all getting quite elaborate, isn't it?  

I don't see a way in which this is not correct, it clearly does achieve some new simplifications and the test coverage is reassuring.

However... I must admit that I am finding it quite hard to follow!

We are beginning to reach diminishing returns on these marker simplifications: this one is a fairly significant increase in overall complexity, for relatively modest gains.

To my taste - and probably this is a matter of taste, reasonable people might disagree - I think it's not worth it.

But that's not a hill that I would want to die on, I don't intend to argue about it.  As I say, I don't see bugs.

radoering:
I can understand your concerns and I agree that it is a significant increase in complexity but I think it is a significant gain, too. If it was just about cosmetics I might even doubt whether it is worth it. But the main reason why marker simplifications are necessary is that they lay the foundations for reducing the number of overrides (regarding python-poetry/poetry#4695) - and that is not just a performance issue.

Any omitted marker simplification can lead to an unnecessary override and any unnecessary override can lead to an unnecessary failure in dependency resolution because the solver tries to find a solution for a case that is not relevant due to markers that can never be fulfilled. Considering that users experiencing an unfounded solver failure sometimes can't do anything but reducing the number of supported python versions, I think it's worth it.

Nevertheless, thanks for the review. If someone comes up with a simpler solution, I will be happy to adopt it. If there isn't one, at least in my opinion a complex solution is better than no solution at all.

abn (approved):

PR #310, opened 2022-03-21
build whl using shutil move with copy instead of copy2

aronstef:
Resolves: I have not created an issue for this and seems like no one else has. See issue description below.

I have an issue using "poetry build" inside a pipeline container run with podman.
I get permission errors when poetry uses shutils copystat function.
The copystat function copies "extended attributes" which fails in a podman container with mounted drives due to permission issues.

When changing the shutil.move function to use copy instead of copy2 the "extended attributes" will not be transfered and the build process will succeed in podman containers without breaking existing functionality.

### DIFF
--- a/src/poetry/core/masonry/builders/wheel.py
+++ b/src/poetry/core/masonry/builders/wheel.py
@@ -122,7 +122,7 @@ class WheelBuilder(Builder):
         wheel_path = dist_dir / self.wheel_filename
         if wheel_path.exists():
             wheel_path.unlink()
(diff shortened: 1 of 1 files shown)


abn:
Doesn't this drop the extended metadata like file creation/modification time etc? I suspect that would not be desireable for the general cases as this will mean the checksum for a wheel/sdist is less deterministic. I wonder if it is better to fallback to `copy` if an error is raised so as to not change current behaviour.

And the upstream issue on podman seems to indicate that the issue needs to be fixed in `shutils`.

PR #312, opened 2022-03-22
Set GIT_DIR env when checking for git

ptrcnull:
This prevents poetry from traversing beyond project directory

Currently 2 tests are failing, because they expect to be in a git repo (in this case, `poetry-core`); TODO: make them have their own repo, which should have been the case from the start

Resolves: python-poetry/poetry#1946

### DIFF
--- a/src/poetry/core/vcs/__init__.py
+++ b/src/poetry/core/vcs/__init__.py
@@ -16,9 +16,13 @@ def get_vcs(directory: Path) -> Optional[Git]:
     try:
         from poetry.core.vcs.git import executable
 
+        env = os.environ.copy()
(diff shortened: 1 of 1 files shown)


neersighted:
I'm not fully convinced this is the right solution -- can you further explain why this is advantageous?

abn:
Note that this will be ripped out soon. We should really replace git ignore parsing with a pure python solution.

ptrcnull:
the current behaviour is in fact correct; the reason i thought this should be fixed is because when you try to run `poetry build` in a directory that's ignored by git entirely, it just outputs an empty wheel file with no python files inside it. if the gitignore parsing is getting reworked anyway, i hope this will be fixed

PR #313, opened 2022-03-26
Improve dependency group access and fix type hinting for groups/extras

abn:
- package: improve access to dependency groups
- dependencies: fix groups/extras type hinting


radoering (changes requested):
`features` in `Package.__init__()` and `Package.with_features()` can be changed from `List` to `Iterable`, too.

radoering (approved):

PR #316, opened 2022-04-01
Fix an issue where dependencies with an epoch are parsed as empty

radoering:
* remove duplicated and buggy method of first_pre_release() and use correct method first_prerelase()

Resolves: python-poetry/poetry#5374


abn (changes requested):
Minor suggestion.

abn (approved):

PR #318, opened 2022-04-01
build: do not generate setup.py by default

abn:
This is the first step in removing setup file generation from Poetry
projects. With this change, projects that require a setup.py to be
generated when a build script is used needs to explicitly set
`tool.poetry.build.generate-setup-file`, introduced in #26, to `true`
in `pyproject.toml`.

**References:**
- #26 
- https://github.com/python-poetry/poetry/issues/1338#issuecomment-657860111

**Warning** :warning: 
This is a breaking change for users relying on setuptools to run their build script without explicitly configuring the generation.

### DIFF
--- a/src/poetry/core/json/schemas/poetry-schema.json
+++ b/src/poetry/core/json/schemas/poetry-schema.json
@@ -635,7 +635,7 @@
         "generate-setup-file": {
           "type": "boolean",
           "description": "Generate and include a setup.py file in sdist.",
-          "default": true
+          "default": false
         },
         "script": {
           "$ref": "#/definitions/build-script"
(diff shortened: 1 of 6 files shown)


abn:
@python-poetry/core @python-poetry/triage Would apprecaite feedback on this. We need to disable auto setup generation, this was originally planned for Poetry 1.2 as it was deprecated in Poetry 1.1.

I am fairly certain releasing this will break for a few folks building extensions using setuptools, just not sure how large the number is going to be.

Secrus (approved):

Secrus:
@abn I updated the PR, but this required deleting some code in `test_complete.py`. Could you check if that's fine?

finswimmer (dismissed):

finswimmer (dismissed):

Secrus (approved):

PR #321, opened 2022-04-11
specification: add ignore_features flag to is_same_package_as

radoering:
... as proposed in https://github.com/python-poetry/poetry/pull/5435#discussion_r846855182

**Update:** Since `allow_similar` is too ambiguous the parameter is called `ignore_features`.

### DIFF
--- a/src/poetry/core/packages/specification.py
+++ b/src/poetry/core/packages/specification.py
@@ -71,9 +71,15 @@ class PackageSpecification:
     def features(self) -> frozenset[str]:
         return self._features
 
-    def is_same_package_as(self, other: PackageSpecification) -> bool:
-        if other.complete_name != self.complete_name:
-            return False
+    def is_same_package_as(
+        self, other: PackageSpecification, *, ignore_features: bool = False
+    ) -> bool:
+        if ignore_features:
+            if other.name != self.name:
+                return False
+        else:
+            if other.complete_name != self.complete_name:
+                return False
 
         if self._source_type:
             if self._source_type != other.source_type:
--- a/tests/packages/test_specification.py
+++ b/tests/packages/test_specification.py
@@ -1,0 +1,37 @@
+from __future__ import annotations
+
+import pytest
+
+from poetry.core.packages.specification import PackageSpecification
+
+
+@pytest.mark.parametrize(
+    "spec1, spec2, expected_exact, expected_ignore_features",
+    [
+        (PackageSpecification("a"), PackageSpecification("a"), True, True),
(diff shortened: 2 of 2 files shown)


src/poetry/core/packages/specification.py
```diff
@@ -71,9 +71,15 @@ def source_subdirectory(self) -> str | None:
     def features(self) -> frozenset[str]:
         return self._features
 
-    def is_same_package_as(self, other: PackageSpecification) -> bool:
-        if other.complete_name != self.complete_name:
-            return False
+    def is_same_package_as(
+        self, other: PackageSpecification, *, allow_similar: bool = False
+    ) -> bool:
+        if allow_similar:
```
abn, line 77: Should we also treat `foo (1.2.3 git rev)` as being similar to `foo (1.2.3)` ?

radoering: It depends:

I suppose for the current downstream use case that would not be suitable because "similar" means same package with any additional features. That can be true for `foo (1.2.3 git rev)` and `foo (1.2.3)` but is not safe.

Considering other use cases (I don't know yet), it might be suitable.

If we only want to map the current use case, maybe, the parameter should be renamed: `match_complete_name=True` or `ignore_features=False`? (I think the latter would require a slightly different implementation but is most exactly what is required by the current downstream use case.)

If we don't want to map the current use case, we can keep the parameter name and treat your example as similar.

radoering: I thought about it again and I changed it to `ignore_features`. That is better because it is clear and exactly what is required. `allow_similar` is too ambiguous and open to interpretation. I think I will revisit the downstream usage of `allow_similar` after this has been merged and included in the next (pre-)release.

abn: So there is another issue that I keep forgetting about. There is a use case of "overrides" in development.

```sh
poetry add project
poetry add --group=dev git+https://github.com/org/project.git
# alternatively
poetry add --group=dev --editable ./project
```

I am not sure how the recent changes have affected this use case.

Also, along the same lines (maybe) https://github.com/python-poetry/poetry/issues/5617.

abn: Althought, it is important to note that these are poetry specific use case and might not be interesting for core itself.

abn: I agree. Lets drop this for now.

radoering:
Superseded by #365. (Since `is_same_source_as()` plus `name` comparison fulfills the use case, introducing an additional parameter is not worth it anymore.)

PR #325, opened 2022-04-15
specification: make is_same_package_as() commutative and …

radoering:
always return False if source_types do not match

Fixes an issue where `p1.is_same_package_as(p2)` returns `True`, but `p2.is_same_package_as(p1)` return `False`.

### DIFF
--- a/src/poetry/core/packages/specification.py
+++ b/src/poetry/core/packages/specification.py
@@ -75,9 +75,10 @@ class PackageSpecification:
         if other.complete_name != self.complete_name:
             return False
 
+        if self._source_type != other.source_type:
+            return False
+
         if self._source_type:
-            if self._source_type != other.source_type:
-                return False
 
             if (
                 self._source_url or other.source_url
--- a/tests/packages/test_specification.py
+++ b/tests/packages/test_specification.py
@@ -1,0 +1,23 @@
+from __future__ import annotations
+
+import pytest
+
+from poetry.core.packages.specification import PackageSpecification
+
+
+@pytest.mark.parametrize(
+    "spec1, spec2, expected",
+    [
+        (PackageSpecification("a"), PackageSpecification("a"), True),
+        (PackageSpecification("a", "type1"), PackageSpecification("a", "type1"), True),
+        (PackageSpecification("a", "type1"), PackageSpecification("a", "type2"), False),
+        (PackageSpecification("a"), PackageSpecification("a", "type1"), False),
+        (PackageSpecification("a", "type1"), PackageSpecification("a"), False),
+    ],
+)
+def test_is_same_package_source_type(
+    spec1: PackageSpecification,
+    spec2: PackageSpecification,
+    expected: bool,
+) -> None:
+    assert spec1.is_same_package_as(spec2) == expected


radoering:
Even though the current behavior is clearly incorrect (because not commutative), this change might be too risky without further investigation.

It's very likely, that in some cases the behavior as proposed in this PR is desired, but in other cases a less strict behavior à la "package is the same if name is the same" is required (see python-poetry/poetry#5617 for example). These cases have to be identified and fixed before fixing `is_same_package_as()`.

radoering:
@abn Since you had some concerns about this change influencing the behavior of the installer, I investigated that a bit. This should be the relevant part in the code: https://github.com/python-poetry/poetry/blob/80b03ece039ec72e86d87295ea557a4282891bd9/src/poetry/puzzle/transaction.py#L45-L51

It's obvious that there is a special handling if PyPI dependencies (no source_type) and legacy dependencies are involved. Given the version is equal, I investigated the relevant combinations:

installed_package | result_package | update
---|---|---
pypi | legacy | no (because of the source_type check in the linked code)
legacy | pypi | **no (because of the bug fixed in this PR)**
pypi | direct | yes
direct | pypi | **no (because of the bug fixed in this PR)**
legacy | direct | yes
direct | legacy | yes

I assume the missing update when switching from direct to pypi is a real bug. The missing update when switching between pypi and legacy may be desired (though I'm not sure if it is sensible). If it is desired it should be handled directly in the linked spot and not partially relying on a weird behavior of `is_same_package_as()`.

By the way, due to `is_same_package_as()` not being commutative, [Package.\_\_eq\_\_()](https://github.com/python-poetry/poetry-core/blob/81551a7326f16a78009f7a945100154f1a7dca49/src/poetry/core/packages/package.py#L551) is not commutative! `__eq__` not being commutative is kind of 💩 and may provoke even stranger bugs. I kind of feel like we should fix this better sooner than later even if it's risky...

What do you think? Fix https://github.com/python-poetry/poetry/blob/80b03ece039ec72e86d87295ea557a4282891bd9/src/poetry/puzzle/transaction.py#L45-L51 one way or the other (so it does not rely on the weird `is_same_package_as()` behavior) and than merge this one prior to the next pre-release?

radoering:
I did some further investigations. It seems an installed package is never of type legacy. That's just not tracked. If you install a package with `source_type=="legacy"`, the package in the installed repository has no `source_type`. Thus, cases in the table in my previous comment with `installed_package` of type legacy are only theoretical use cases and making `is_same_package_as()` commutative should not change the behavior of the `install` command. No need for a downstream fix.

neersighted:
> I did some further investigations. It seems an installed package is never of type legacy. That's just not tracked. If you install a package with `source_type=="legacy"`, the package in the installed repository has no `source_type`. Thus, cases in the table in my previous comment with `installed_package` of type legacy are only theoretical use cases and making `is_same_package_as()` commutative should not change the behavior of the `install` command (except for the desired fix when overwriting a direct origin dependency by a PyPI dependency). No need for a downstream fix.

Thanks for looking into this! Given your investigation, I am pretty comfortable merging this, especially as we want to get a beta out. :tada:

neersighted (approved):

abn:
@radoering thanks again for digging into this. Just voicing that I am in agreement with @neersighted here. We can deal with any fall out in the lead upto poetry `1.2.0rc1`. We definitely need to add a bit more test coverage I feel around this area. Additionally, we should either add parts of your digging into either a README.md in the puzzle module or witin the code itself.

PR #326, opened 2022-04-17
Rename implicit group "default" to "main"

radoering:
This is done in order to avoid ambiguity between "default group" and "default dependencies" (the groups that are considered by a command by default)

Has to be considered together with python-poetry/poetry#5465 and python-poetry/poetry-plugin-export#52

In order to allow the downstream tests to pass, python-poetry/poetry#5465 has to be merged first.


src/poetry/core/factory.py
```diff
@@ -78,6 +78,8 @@ def _add_package_group_dependencies(
         group: str | DependencyGroup,
         dependencies: DependencyConfig,
     ) -> None:
+        from poetry.core.packages.dependency_group import MAIN_GROUP
```
abn, line 81: We should do a try catch here and on ImprotError set `MAIN_GROUP="default"` until we bump core.

radoering: Interesting idea. I don't think it will do any good here in `poetry-core`. But I can do this in the other two PRs.

abn: Ah wrong PR. I meant to comment on the poetry one.

abn (approved):

PR #327, opened 2022-04-19
semver: improve constraint parsing

abn:
This change replaces the custom regex used with `packaging.version.VERSION_PATTERN` for consistency with other parts of the code base. Additionally, this fixes previous issues with parsing  dev releases etc.

### DIFF
--- a/src/poetry/core/semver/helpers.py
+++ b/src/poetry/core/semver/helpers.py
@@ -63,9 +63,9 @@ def parse_single_constraint(constraint: str) -> VersionConstraint:
     # Tilde range
     m = TILDE_CONSTRAINT.match(constraint)
     if m:
-        version = Version.parse(m.group(1))
+        version = Version.parse(m.group("version"))
         high = version.stable.next_minor()
-        if len(m.group(1).split(".")) == 1:
+        if version.release.precision == 1:
             high = version.stable.next_major()
 
         return VersionRange(version, high, include_min=True)
@@ -73,16 +73,8 @@ def parse_single_constraint(constraint: str) -> VersionConstraint:
     # PEP 440 Tilde range (~=)
     m = TILDE_PEP440_CONSTRAINT.match(constraint)
     if m:
-        precision = 1
-        if m.group(3):
-            precision += 1
-
-            if m.group(4):
-                precision += 1
-
-        version = Version.parse(m.group(1))
-
-        if precision == 2:
+        version = Version.parse(m.group("version"))
+        if version.release.precision == 2:
             high = version.stable.next_major()
         else:
             high = version.stable.next_minor()
@@ -92,14 +84,14 @@ def parse_single_constraint(constraint: str) -> VersionConstraint:
     # Caret range
     m = CARET_CONSTRAINT.match(constraint)
     if m:
-        version = Version.parse(m.group(1))
+        version = Version.parse(m.group("version"))
 
         return VersionRange(version, version.next_breaking(), include_min=True)
 
     # X Range
     m = X_CONSTRAINT.match(constraint)
     if m:
-        op = m.group(1)
+        op = m.group("op")
         major = int(m.group(2))
         minor = m.group(3)
 
@@ -124,15 +116,8 @@ def parse_single_constraint(constraint: str) -> VersionConstraint:
     # Basic comparator
     m = BASIC_CONSTRAINT.match(constraint)
     if m:
-        op = m.group(1)
-        version_string = m.group(2)
-
-        # Technically invalid constraints like `>= 3.*` will appear
-        # here as `3.`.
-        # Pip currently supports these and to avoid breaking existing
-        # users workflows we need to support them as well. To do so,
-        # we just remove the inconsequential part.
-        version_string = version_string.rstrip(".")
+        op = m.group("op")
+        version_string = m.group("version")
 
         if version_string == "dev":
             version_string = "0.0-dev"
--- a/src/poetry/core/semver/patterns.py
+++ b/src/poetry/core/semver/patterns.py
@@ -2,21 +2,27 @@ from __future__ import annotations
 
 import re
 
+from packaging.version import VERSION_PATTERN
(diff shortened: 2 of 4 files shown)


radoering:
Shouldn't the regex be fixed? This seems like doing workarounds.

abn:
Probably yes. But that requires a bit more testing I reckon as I am unsure what the impact of that is going to be else where and in poetry.

Will take a look.

abn:
@radoering ended up doing a more comprehensive fix. Thanks for the nudge :)

src/poetry/core/semver/helpers.py
```diff
@@ -63,26 +63,19 @@ def parse_single_constraint(constraint: str) -> VersionConstraint:
...
-            precision += 1
-
-            if m.group(4):
-                precision += 1
-
-        version = Version.parse(m.group(1))
-
-        if precision == 2:
+        version = Version.parse(m.group("version"))
+        print(version)
```
radoering, line 77:
```suggestion
```

src/poetry/core/semver/patterns.py
```diff
@@ -2,21 +2,27 @@
...
 
-COMPLETE_VERSION = re.compile("(?i)" + _COMPLETE_VERSION)
+CARET_CONSTRAINT = re.compile(
+    rf"^\^(?P<version>{VERSION_PATTERN})$", re.VERBOSE | re.IGNORECASE
+)
+TILDE_CONSTRAINT = re.compile(
+    rf"^~(?!=)\s*(?P<version>{VERSION_PATTERN})$", re.VERBOSE | re.IGNORECASE
+)
+TILDE_PEP440_CONSTRAINT = re.compile(
+    rf"^~=\s*(?P<version>{VERSION_PATTERN})\s*$", re.VERBOSE | re.IGNORECASE
```
radoering, line 17:
```suggestion
    rf"^~=\s*(?P<version>{VERSION_PATTERN})$", re.VERBOSE | re.IGNORECASE
```

We don't need this, do we?

src/poetry/core/semver/patterns.py
```diff
@@ -2,21 +2,27 @@
...
 
-CARET_CONSTRAINT = re.compile(rf"(?i)^\^({_COMPLETE_VERSION})$")
-TILDE_CONSTRAINT = re.compile(rf"(?i)^~(?!=)\s*({_COMPLETE_VERSION})$")
-TILDE_PEP440_CONSTRAINT = re.compile(rf"(?i)^~=\s*({_COMPLETE_VERSION})$")
-X_CONSTRAINT = re.compile(r"^(!=|==)?\s*v?(\d+)(?:\.(\d+))?(?:\.(\d+))?(?:\.[xX*])+$")
-BASIC_CONSTRAINT = re.compile(rf"(?i)^(<>|!=|>=?|<=?|==?)?\s*({_COMPLETE_VERSION}|dev)")
+# note that we also allow technically incorrect version patterns with astrix (eg: 3.5.*)
+# as this is supported by pip and appears in metadata within python packages
+BASIC_CONSTRAINT = re.compile(
+    rf"^(?P<op><>|!=|>=?|<=?|==?)?\s*(?P<version>{VERSION_PATTERN}|dev)(\.\*)?\s*$",
```
radoering, line 26:
```suggestion
    rf"^(?P<op><>|!=|>=?|<=?|==?)?\s*(?P<version>{VERSION_PATTERN}|dev)(\.\*)?$",
```
We don't need this, do we?

radoering (changes requested):
Much better than before. 👍 Just some minor nitpicks.

src/poetry/core/semver/patterns.py
abn: We do, because this was sort of already done before. And also handles some bad data issues. Not common enough to be tested.

src/poetry/core/semver/patterns.py
abn: Similar to above. Also see how we handle the PEP440 regex.

abn:
Let me know if you feel strongly about any of the comments I responded to.

radoering:
At the moment, we allow trailing white space for 2 out of 5 constraint patterns. If we allow it, shouldn't we do for all?

I only feel strongly against _"Not common enough to be tested."_. A test seems quite easy and makes clear that this part of the regex is not a mistake.

abn:
I agree. Will allow it in all of them as it makes sense. Tests too ;)

src/poetry/core/semver/patterns.py
```diff
@@ -2,21 +2,27 @@
...
-)
 
-_COMPLETE_VERSION = (
-    rf"v?(?:\d+!)?(\d+)(?:\.(\d+))?(?:\.(\d+))?(?:\.(\d+))?{MODIFIERS}(?:\+[^\s]+)?"
-)
+COMPLETE_VERSION = re.compile(VERSION_PATTERN, re.VERBOSE | re.IGNORECASE)
 
-COMPLETE_VERSION = re.compile("(?i)" + _COMPLETE_VERSION)
+CARET_CONSTRAINT = re.compile(
+    rf"^\^(?P<version>{VERSION_PATTERN})$", re.VERBOSE | re.IGNORECASE
```
radoering, line 11: The `\s*` is missing here what triggered me again to question if it is required at all. I removed it in the other patterns and the tests still pass. That seems plausible because leading and trailing whitespace is handled here:
https://github.com/python-poetry/poetry-core/blob/16cb1592200694e2ea2f2dfd4931b7843a1567da/src/poetry/core/semver/helpers.py#L18

Thus, I think we can safely remove the `\s*` at the end of the other patterns.

radoering (changes requested):

src/poetry/core/semver/patterns.py
abn: Sure, I'm not really fussed about it either way. Let me do that.

radoering (approved):

PR #328, opened 2022-04-21
canonicalize package name correctly

dimbleby:
Resolves: https://github.com/python-poetry/poetry/issues/5474, https://github.com/python-poetry/poetry/issues/3132, https://github.com/python-poetry/poetry/issues/3200

looks like this regex was probably just badly copied from https://github.com/pypa/packaging/blob/ba124324e866e518ebfe0378a25b6ba4816e5880/packaging/utils.py#L27

### DIFF
--- a/src/poetry/core/utils/helpers.py
+++ b/src/poetry/core/utils/helpers.py
@@ -17,7 +17,7 @@ from typing import no_type_check
 from poetry.core.version.pep440 import PEP440Version
 
 
-_canonicalize_regex = re.compile(r"[-_]+")
+_canonicalize_regex = re.compile(r"[-_.]+")
 
 
 def combine_unicode(string: str) -> str:
--- a/tests/utils/test_helpers.py
+++ b/tests/utils/test_helpers.py
@@ -73,7 +73,7 @@ isort@ git+git://github.com/timothycrosley/isort.git@e63ae06ec7d70b06df9e5283576
     assert result == expected
 
 
-@pytest.mark.parametrize("raw", ["a-b-c", "a_b-c", "a_b_c", "a-b_c"])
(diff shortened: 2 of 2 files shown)


dimbleby:
downstream tests fixed in https://github.com/python-poetry/poetry/pull/5475

dimbleby:
https://github.com/python-poetry/poetry/pull/5475

dimbleby:
I have just spotted that there is history here at #100 and #132.

#132 doesn't say what the "unintended side effects" were that caused this to be reverted. I doubt that they were worse than the infinite loops that this fixes.  

That makes my guess that this was mis-copied from `packaging.utils` look less likely; but in that case it might be reassuring to see that they have the same solution as in this MR.

radoering:
> #132 doesn't say what the "unintended side effects" were that caused this to be reverted. I doubt that they were worse than the infinite loops that this fixes.

At least it says that the canonicalized names were propagated in too many places and that the original names should be kept longer. Thus, I suppose we would have to evaluate each call of `canonicalize_name`, remove some of them and maybe add some new ones if we want to be thorough about this. 🙁

@abn Do you remember why the original PR was reverted? I assume it wasn't just for cosmetic reasons?

abn:
I suspect there were a few different issues that piled up which may have already been resolved. Based on talking with sdispater, seems the issue was dependency resolution was breaking in some cases. Like, for example `A` depends on `A-X` and `A.X` is provided on PyPI. The downstream test failures might be an indicator, but curosory look seems to imply the issue might be more test data than actual resolution failures. We should probably get the poetry tests fixed in a forwards compatible manner and add some test cases around this - hopefully that will reveal any real issues.

abn (approved):

PR #329, opened 2022-04-22
Support build script in sub directory

brandon-leapyear:
Noticed that specifying a path like `build = "foo/bar.py"` would result in `from foo/bar import *` which is invalid syntax


radoering:
Can you add a unit test to `test_sdist.py`, please?

brandon-leapyear:
@radoering done!

radoering (approved):
Great. 👍

PR #330, opened 2022-04-23
add set_version() to PackageProject

dimbleby:
Let's get rid of https://github.com/python-poetry/poetry/blob/master/src/poetry/packages/project_package.py.  If set_version() is useful, put it here.

### DIFF
--- a/src/poetry/core/packages/project_package.py
+++ b/src/poetry/core/packages/project_package.py
@@ -4,12 +4,12 @@ from typing import TYPE_CHECKING
 from typing import Any
 
 from poetry.core.semver.helpers import parse_constraint
+from poetry.core.semver.version import Version
 from poetry.core.version.markers import parse_marker
 
 
 if TYPE_CHECKING:
     from poetry.core.packages.types import DependencyTypes
-    from poetry.core.semver.version_constraint import VersionConstraint
 
 from poetry.core.packages.package import Package
 from poetry.core.packages.utils.utils import create_nested_marker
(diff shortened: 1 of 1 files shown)


radoering (commented):
Making `self._version` mutable does not comply with the inherited `__hash__` method. What do you think about making `ProjectPackage` unhashable by adding `__hash__ = None` (and a comment why it's done) to the class definition? (I don't think it has to be hashable, but of course I'm not sure. If the downstream tests still pass, that would probably be sufficient for me.)

dimbleby:
> Making self._version mutable does not comply with the inherited __hash__ method. 

That's not the contract.  What it would mean is that callers must be careful about updating the version if they are relying on the hash (eg because they have the package in a set, or dictionary).

FWIW I'm not sure that `set_version()` is useful at all.  The only place I find it sort-of-used is in the plugin example at https://github.com/python-poetry/poetry/blob/dca0c562c42cb79a494eff982fc8a52c632100f4/docs/plugins.md?plain=1#L70-L72.  

It's not clear to me that this is an example of something that anyone might actually want to do.

So perhaps we should just delete this from `poetry` altogether, and not bother replacing it here.

abn:
Don't think it would be desirable to make it unhashable as, iirc, these show up during key lookups somewhere in the code.

dimbleby:
I think I'm wrong about "that's not the contract", https://docs.python.org/3/glossary.html#term-hashable has

> An object is hashable if it has a hash value which never changes during its lifetime ...

so in that case `set_version()` currently represents a bug in `poetry`, and this MR is transferring that bug to `poetry-core`.

Views on removing `set_version()` altogether?

abn:
I agree that the current `set_version` is problematic, and this really should be handled using either allowing `clone()` to accept a version or introducing a `with_version()` method (similar to `Package.with_features()`) that clones the package with a new version. 

The plugin example is I think inspired from community requests for a plugin similar to `setuptools-scm`. So, I would think being able to update the version is a handy functionality.

dimbleby:
https://github.com/python-poetry/poetry/pull/5486 removes the problematic code from poetry.  

If setting versions is desired then an improved MR can be submitted over here.

radoering:
> Don't think it would be desirable to make it unhashable as, iirc, these show up during key lookups somewhere in the code.

`Package` does for sure, but `ProjectPackage`? I thought the root package is the only `ProjectPackage`.

abn:
> `Package` does for sure, but `ProjectPackage`? I thought the root package is the only `ProjectPackage`.

I need to dig up the specifics. But this came up during the repository refactor where there were cases when a project package ended up being returned even where one was not expected. I didn't dig too deep into that one then.

PR #334, opened 2022-04-28
typecheck poetry.core.version

dimbleby:
typechecking poetry.core.version

### DIFF
--- a/src/poetry/core/version/pep440/segments.py
+++ b/src/poetry/core/version/pep440/segments.py
@@ -33,17 +33,16 @@ class Release:
     patch: int | None = dataclasses.field(default=None, compare=False)
     # some projects use non-semver versioning schemes, eg: 1.2.3.4
     extra: int | tuple[int, ...] | None = dataclasses.field(default=None, compare=False)
-    precision: int = dataclasses.field(default=None, init=False, compare=False)
-    text: str = dataclasses.field(default=None, init=False, compare=False)
-    _compare_key: tuple[int, ...] = dataclasses.field(
-        default=None, init=False, compare=True
-    )
+    precision: int = dataclasses.field(init=False, compare=False)
+    text: str = dataclasses.field(init=False, compare=False)
+    _compare_key: tuple[int, ...] = dataclasses.field(init=False, compare=True)
 
     def __post_init__(self) -> None:
         if self.extra is None:
             object.__setattr__(self, "extra", ())
         elif not isinstance(self.extra, tuple):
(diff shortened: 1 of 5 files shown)


src/poetry/core/version/pep440/segments.py
```diff
@@ -74,6 +73,7 @@ def to_string(self) -> str:
         return self.text
 
     def next_major(self) -> Release:
+        assert isinstance(self.extra, tuple)
```
abn, line 76: More out of curiosity, any paritcular reason why an assert is preferred in these cases over a cast?

src/poetry/core/version/pep440/version.py
```diff
@@ -18,12 +18,32 @@
...
+
+class NegativeInfinity(AlwaysSmaller, int):
+    pass
+
+
 T = TypeVar("T", bound="PEP440Version")
 
 # we use the phase "z" to ensure we always sort this after other phases
-_INF_TAG = ReleaseTag("z", math.inf)
+_INF_TAG = ReleaseTag("z", Infinity())
```
abn, line 44: Any reason why this is required?

dimbleby: `math.inf` is not an `int`

src/poetry/core/version/pep440/segments.py
dimbleby: I probably try to go with assert where I think there's any chance that I might be wrong, and cast only when I'm really 100% sure.  Because if I cast and am wrong, then things will break in possibly confusing ways, whereas if I assert and am wrong then things will break in a clearer way.

But I don't promise that I'm 100% consistent about this and, if this were more than curiosity, I'd likely be willing to change some or more of the choices that I've made.

src/poetry/core/version/pep440/version.py
```diff
@@ -188,13 +216,15 @@ def next_patch(self: T) -> T:
...
         if self.is_prerelease():
+            assert self.pre is not None
             pre = self.pre.next_phase() if next_phase else self.pre.next()
         else:
             pre = ReleaseTag(RELEASE_PHASE_ALPHA)
         return self.__class__(epoch=self.epoch, release=self.release, pre=pre)
 
     def next_postrelease(self: T) -> T:
-        if self.is_prerelease():
+        if self.is_postrelease():
```
branchvincent, line 226: 😅

branchvincent (approved):

PR #335, opened 2022-04-28
Fix editable installations using `setup.py`.

matthewwardrop:
Hi! I have really come to love poetry, but cannot currently use it as is for my project that uses compiled modules. The issues I've encountered (and fixed in this PR) are detailed below. Hopefully these small patches are helpful to others also.

When you have configured poetry to generate a setup file, for example by using:
```
[tool.poetry.build]
script = "build.py"
generate-setup-file = true
```
and then attempt an editable install, `poetry` will include the module in the generated wheel. This shouldn't happen according to [PEP 660](https://peps.python.org/pep-0660/) (you should only include the files which are necessary for the editable installation to work), and it gets in the way of the editable module being imported from the "editable" path. This is fixed in the commit labelled: "fix: Do not include library in wheel when performing an editable install".

Additionally, when your project uses extension modules, to be usable in an editable installation they need to be built in-place during the editable installation. This is added in the commit labelled: "enh: Build extension modules inplace during editable install." PEP 660 also mentions that this is an optional thing to do, but for modules with compiled dependencies, it is required to achieve an editable installation.

- [ ] Added **tests** for changed code. (I've tested things locally, but will wait for your blessing before worrying about unit tests).
- [ ] Updated **documentation** for changed code. (Based on comments in https://github.com/python-poetry/poetry/issues/2740, I don't think you are documenting this functionality yet).

### DIFF
--- a/src/poetry/core/masonry/builders/wheel.py
+++ b/src/poetry/core/masonry/builders/wheel.py
@@ -169,7 +169,7 @@ class WheelBuilder(Builder):
 
                     build_dir = self._path / "build"
                     lib = list(build_dir.glob("lib.*"))
-                    if not lib:
+                    if self._editable or not lib:
                         # The result of building the extensions
                         # does not exist, this may due to conditional
                         # builds, so we assume that it's okay
--- a/src/poetry/core/masonry/builders/wheel.py
+++ b/src/poetry/core/masonry/builders/wheel.py
@@ -201,6 +201,15 @@ class WheelBuilder(Builder):
             )
 
     def _run_build_command(self, setup: Path) -> None:
+        if self._editable:
+            subprocess.check_call(
+                [
+                    self.executable.as_posix(),
+                    str(setup),
+                    "build_ext",
+                    "--inplace",
+                ]
+            )
         subprocess.check_call(
             [
                 self.executable.as_posix(),


matthewwardrop:
@python-poetry/triage

matthewwardrop:
@dimbleby @branchvincent @radoering Sorry to ping directly, but wanted to see whether this is something that you would be interested in merging, or whether I should commit to maintaining my own fork for my repositories?

neersighted:
If I'm not mistaken, python-poetry/poetry#5590 should have fixed this in a more proper way. Could you please test against the latest master and report back?

matthewwardrop:
@neersighted Nope. Unfortunately the `master` branch of `poetry` has the same pathologies that I tried to fix here. It doesn't compile the modules in the local directory *and* it installs all the package files into `site-packages`. :(.

abn:
I'm a bit unclear on what is being solved here. Poetry's editable builder has nothing to do with poetry-core's wheel builder. No wheel is built during that process. A temporary setup file is created and `pip install -e .` is called when setup file generation is enabled. 

Can you provide an example project with the issue you are detailing. 

Also do note that PEP 662 is rejected.

matthewwardrop:
Hi @abn,

Thanks for your reply, and yes... apologies. I should have referred to PEP 660 (will update in the PR description).

I've put together a minimum working example [here](https://github.com/matthewwardrop/poetry-editable-failure).

Let me know if you need more context/etc.

matthewwardrop:
@abn Any additional thoughts given the example code?

robbotorigami:
Is it possible to get a response on this thread? Unfortunately, especially with the changes made in https://github.com/python-poetry/poetry/pull/7975, this is now a significant issue for doing development in any project that uses compiled extension modules. There's no way I've found to get an editable install to work properly, either with `poetry install` or `pip install -e .`.

radoering:
I'll take a closer look if the conflicts are resolved and a unit test is added.

robbotorigami:
@matthewwardrop If you'd like to work on this let me know, otherwise I'm happy to make a fork and work on @radoering 's requests.

PR #337, opened 2022-04-30
use tempfile.TemporaryDirectory

dimbleby:
cf https://github.com/python-poetry/poetry/pull/5522

but note that there's a testcase in this repository: https://github.com/python-poetry/poetry-core/blob/8c75c1ae53996be8c4c88ccc2c04dfd6940e0290/tests/utils/test_helpers.py#L90.  So if the pipeline passes that should give us confidence that this is good,


dimbleby:
Looks like Windows and python3.7 are not happy with this, https://bugs.python.org/issue26660 presumably.

if and when we eventually get to do this, should probably remove the unit test case also - it wouldn't make much sense to have a unit test for tempfile function.

neersighted:
Re: implementing this now, one option is https://github.com/python-poetry/poetry/pull/5522#issuecomment-1120488761

dimbleby:
rebased in #566 so that we no longer need to care about Windows python 3.7 bugs

PR #338, opened 2022-05-01
chore(tests): remove type errors

branchvincent:
Removes `tests` from mypy's ignored list

The first commit 9f8ffc2 fixes type issues, while the second 81bfebd just adds explicit return types `-> None`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -82,6 +82,7 @@ module = [
   'lark.*',
   'setuptools.*',
   'tomlkit.*',
+  'virtualenv.*',
 ]
 ignore_missing_imports = true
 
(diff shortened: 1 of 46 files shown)


tests/packages/test_dependency.py
```diff
@@ -294,7 +294,7 @@ def test_with_constraint():
 def test_marker_properly_sets_python_constraint():
     dependency = Dependency("foo", "^1.2.3")
 
-    dependency.marker = 'python_version >= "3.6" and python_version < "4.0"'
+    dependency.marker = 'python_version >= "3.6" and python_version < "4.0"'  # type: ignore[assignment]
```
branchvincent, line 299: this is an issue with `marker`'s setter accepting the union `str | BaseMarker` and mypy not being able to understand: https://github.com/python/mypy/issues/3004

pyproject.toml
```diff
@@ -111,6 +111,11 @@ module = [
 ]
 ignore_errors = true
 
+[[tool.mypy.overrides]]
+module = ['tests.*']
+disallow_untyped_decorators = false
```
branchvincent, line 116: this is only disabled due to `pytest` not currently being in `mypy`'s environment

needs https://github.com/python-poetry/poetry/pull/5279

radoering (approved):

PR #339, opened 2022-05-01
chore(packages): remove type errors

branchvincent:
Removes `poetry.core.packages.*` from the ignored list

Relates-to: https://github.com/python-poetry/poetry/issues/5017


src/poetry/core/packages/directory_dependency.py
```diff
@@ -101,9 +101,7 @@ def with_constraint(
             extras=list(self._extras),
         )
 
-        new._constraint = constraint
-        new._pretty_constraint = str(constraint)
-
+        new.set_constraint(constraint)
```
branchvincent, line 104: before, this allowed `constraint` to be a `str`, although it seems the intention was for `set_constraint` to handle normalizing to `VersionConstraint` (as well as setting `_pretty_constraint`)

https://github.com/python-poetry/poetry-core/blob/f26e2361f0d643826d31366887dcb425fc078214/src/poetry/core/packages/dependency.py#L102

src/poetry/core/packages/utils/utils.py
```diff
@@ -138,20 +142,20 @@ def splitext(path: str) -> tuple[str, str]:
...
+    groups: list[list[tuple[str, str, str]]] = [[]]
 
     for marker in markers:
         if or_:
             groups.append([])
 
         if isinstance(marker, (MultiMarker, MarkerUnion)):
             groups[-1].append(
-                group_markers(marker.markers, isinstance(marker, MarkerUnion))
+                group_markers(marker.markers, isinstance(marker, MarkerUnion))  # type: ignore[arg-type]
```
radoering, line 158: I got stuck here (in combination with the changed return type), but maybe not worth thinking about it since `group_markers` is not required anymore with #347. (The output of it seems to be in a weird form for representing `or` anyway.)

radoering:
@branchvincent Can you rebase, please? (Sorry for the merge conflict.) 🙏

src/poetry/core/packages/dependency.py
```diff
@@ -538,6 +533,7 @@ def create_from_pep_508(
                 link = Link(path_to_url(p))
 
         # it's a local file, dir, or url
+        dep: Dependency | None
```
dimbleby, line 536: suggestion: initialize the value to `None` here, rather than later.

src/poetry/core/packages/dependency.py
```diff
@@ -558,7 +554,7 @@ def create_from_pep_508(
                 name = m.group("name")
                 version = m.group("ver")
 
-            name = req.name or link.egg_fragment
+            name = cast(str, req.name or link.egg_fragment)
```
dimbleby, line 557: it's not clear what the intention of the (original) code is here: the name extracted only three lines previously is discarded here.  So something has gone wrong and it would be good to take the opportunity to fix it.

My guess is that this whole fragment can be discarded (anyway that doesn't break any tests).

src/poetry/core/packages/package.py
```diff
@@ -83,18 +85,18 @@ def __init__(
...
         self.repository_url: str | None = None
         self.documentation_url: str | None = None
         self.keywords: list[str] = []
-        self._license = None
+        self._license: License | None = None
         self.readmes: tuple[Path, ...] = ()
 
-        self.extras = {}
-        self.requires_extras = []
+        self.extras: dict[str, list[str | Dependency]] = {}
```
dimbleby, line 98: pretty sure this should be `dict[str, list[Dependency]]`

src/poetry/core/packages/utils/utils.py
```diff
@@ -164,10 +168,10 @@ def group_markers(
 def convert_markers(marker: BaseMarker) -> dict[str, list[list[tuple[str, str]]]]:
     groups = group_markers([marker])
 
-    requirements = {}
+    requirements: dict[str, list[list[tuple[str, str]]]] = {}
```
dimbleby, line 171: this is `ConvertedMarkers` isn't it?

src/poetry/core/packages/dependency.py
```diff
@@ -412,7 +407,7 @@ def _create_nested_marker(
                 if not constraint.include_min:
                     op = ">"
 
-                version = constraint.min.text
+                version: Version | str = constraint.min.text
```
dimbleby, line 410: or I think this can always be a string (without needing annotation) if you set `version = constraint.max.text` a few lines later

src/poetry/core/packages/dependency.py
dimbleby: (and if that's right then there are no users of `egg_fragment` left so that can be discarded too)

neersighted:
Closing this in favor of #354.

PR #343, opened 2022-05-02
semver/pep508: handle wildcard exclusions

abn:
Prior to this change, when exporting PEP 508 strings for dependencies, wildcard exclusion constraints like `!=2.0.*` were incorrectly serialised as invalid PEP 508 due to how version unions were serialsed.

This change allows for determining if a version union is a single wildcard range exclusion, and if so serialise it appropriately.

Resolves: python-poetry/poetry#5470

PS: Not entirely sure if this is the best fix, but it does the trick for now. Not entirely happy with the determination logic, if anyone has a better approach please do share.

Additionally, during testing it was identified that the `X_CONSTRAINT` pattern mattaching when parsing constraints ignores any non semver versions (eg: calver). While the parser is in the "semver" package, since Poetry needs to support PEP440 based constraints, this should be improved. Will propose a fix in another PR.

### DIFF
--- a/src/poetry/core/packages/dependency.py
+++ b/src/poetry/core/packages/dependency.py
@@ -260,7 +260,13 @@ class Dependency(PackageSpecification):
 
         constraint = self.constraint
         if isinstance(constraint, VersionUnion):
-            if constraint.excludes_single_version():
+            if (
+                constraint.excludes_single_version()
+                or constraint.excludes_single_wildcard_range()
+            ):
+                # This branch is a short-circuit logic for special cases and
+                # avoids having to split and parse constraint again. This has
+                # no functional difference with the logic in the else branch.
                 requirement += f" ({str(constraint)})"
             else:
                 constraints = self.pretty_constraint.split(",")
--- a/src/poetry/core/semver/version_union.py
+++ b/src/poetry/core/semver/version_union.py
@@ -238,6 +238,157 @@ class VersionUnion(VersionConstraint):
 
         raise ValueError(f"Unknown VersionConstraint type {constraint}")
 
+    def _exclude_single_wildcard_range_string(self) -> str:
+        """
+        Helper method to convert this instance into a wild card range
+        string.
+        """
+        if not self.excludes_single_wildcard_range():
+            raise ValueError("Not a valid wildcard range")
+
+        # we assume here that since it is a single exclusion range
+        # that it is one of "< 2.0.0 || >= 2.1.0" or ">= 2.1.0 || < 2.0.0"
+        # and the one with the max is the first part
+        idx_order = (0, 1) if self._ranges[0].max else (1, 0)
+        one = self._ranges[idx_order[0]].max
+        two = self._ranges[idx_order[1]].min
+
+        # versions can have both semver and non semver parts
+        parts_one = [
+            one.major,
+            one.minor or 0,
+            one.patch or 0,
+            *list(one.non_semver_parts or []),
+        ]
+        parts_two = [
+            two.major,
+            two.minor or 0,
+            two.patch or 0,
+            *list(two.non_semver_parts or []),
+        ]
+
+        # we assume here that a wildcard range implies that the part following the
+        # first part that is different in the second range is the wildcard, this means
+        # that multiple wildcards are not supported right now.
+        parts = []
+
+        for idx, part in enumerate(parts_one):
+            parts.append(str(part))
+            if parts_two[idx] != part:
+                # since this part is different the next one is the wildcard
+                # for example, "< 2.0.0 || >= 2.1.0" gets us a wildcard range
+                # 2.0.*
+                parts.append("*")
+                break
+        else:
+            # we should not ever get here, however it is likely that poorly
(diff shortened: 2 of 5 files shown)


abn:
@radoering since you looked at the last semver fix; can you take a peek here too?

src/poetry/core/packages/dependency.py
```diff
@@ -260,7 +260,10 @@ def base_pep_508_name(self) -> str:
 
         constraint = self.constraint
         if isinstance(constraint, VersionUnion):
-            if constraint.excludes_single_version():
+            if (
+                constraint.excludes_single_version()
+                or constraint.excludes_single_wildcard_range()
```
radoering, line 265: It seems that tests succeed if I remove this line. This should be tested, too.

src/poetry/core/semver/version_union.py
```diff
@@ -238,6 +238,62 @@ def _ranges_for(
 
         raise ValueError(f"Unknown VersionConstraint type {constraint}")
 
+    def _exclude_single_wildcard_range_string(self) -> str:
+        """
+        Helper method to convert
+        """
+        assert self.excludes_single_wildcard_range()
```
radoering, line 245: We should raise an exception here. Relying on an `assert` seems to be a no-go to me, since asserts are not executed when running python in optimized mode.

tests/packages/test_dependency.py
```diff
@@ -245,6 +245,17 @@ def test_complete_name() -> None:
             ["x"],
             "A[x] (>=1.6.5,!=1.8.0,<3.1.0)",
         ),
+        # test single version range exclusions
+        ("A", ">=1.8,!=2.0.*", None, "A (>=1.8,!=2.0.*)"),
+        ("A", ">=1.8,!=2.*", None, "A (>=1.8,!=2.*)"),
+        ("A", ">=1.8,!=2.*.*", None, "A (>=1.8,!=2.*)"),
+        # we verify that the range exclusion logic is not too eager
+        ("A", ">=1.8,<2.0 || >=2.1.0", None, "A (>=1.8,<2.0 || >=2.1.0)"),
```
radoering, line 253: I removed the `assert` (see previous comment) and this test failed. However, the result `A (>=1.8,!=2.0.*)` seemed to be fine - even better than the expected result. Doesn't it? It seems that the assert fails because the precision differs. Maybe, this can be improved. (See next comment.)

However, since this is the only test that fails without the `assert`, I think a more clear one that fails without the assertion should be added at least, e.g. `("A", ">=1.8,<2.0 || >=2.2.0", None, "A (>=1.8,<2.0 || >=2.2.0)")`.

src/poetry/core/semver/version_union.py
```diff
@@ -238,6 +238,62 @@ def _ranges_for(
...
+        idx_order = (0, 1) if self._ranges[0].max else (1, 0)
+        one = self._ranges[idx_order[0]]
+        two = self._ranges[idx_order[1]]
+
+        is_range_exclusion = (
+            one.max and not one.include_max and two.min and two.include_min
+        )
+        return (
+            is_range_exclusion
+            and one.max.release.precision == two.min.release.precision
```
radoering, line 293: As indicated in the previous comment, using an "alternative precision" may lead to better results. Maybe, something like make lists of the parts like in `[one.major, one.minor, one.patch, *list(one.extra or [])]`, remove trailing zeros, compare the lengths of these lists?

radoering (changes requested):
Seems a bit cumbersome indeed, but since I have no better approach I think it's fine for now with some minor changes.

src/poetry/core/packages/dependency.py
abn: This whole block is more a short-circuit, even removing the if condition will pass the test suite. I will explicity add tests for version range exclusion, but it won't fail even if we remove this block.

src/poetry/core/semver/version_union.py
abn: I have ... added a solution but turned out it was not as simple as that to cover additional edge cases. It does make things more involved. Let me know what you think.

radoering (changes requested):
Wow, definitively more extensive than I would have expected. But since it is encapsulated in its own method and you already know the goal of the long else branch after reading the short if branch it is not that difficult to follow. And since the test coverage seems fine, I'm fine with it for now.

Just two minor nitpicks (see comments).

radoering (approved):

src/poetry/core/semver/version_union.py
```diff
@@ -238,6 +238,157 @@ def _ranges_for(
...
+                    *_extra_one[: idx - 1],
+                    (_extra_one[idx] + 1),
+                    *_extra_one[idx + 1 :],
+                ]
+                _check_versions.add(
+                    from_parts(
+                        _padded_version_one.major,
+                        _padded_version_one.minor,
+                        _padded_version_one.patch,
+                        *_extra,
```
dimbleby, line 364: mypy points out that this is not right - see #353

PR #344, opened 2022-05-04
version: fix normalize_version() according to PEP440

radoering:
Resolves: python-poetry/poetry#5534

`normalize_version()` seems quite buggy supposing that it should do normalizations according to PEP 440. This PR fixes this method and adds tests derived from PEP 440 examples.

Since I wasn't sure if the `to_string(short=False)` method has some other usage (seems to be more semver than PEP440), I introduced a new property `norm_string`.

### DIFF
--- a/src/poetry/core/version/pep440/segments.py
+++ b/src/poetry/core/version/pep440/segments.py
@@ -7,32 +7,23 @@ from typing import Tuple
 from typing import Union
 
 
-RELEASE_PHASE_ALPHA = "alpha"
-RELEASE_PHASE_BETA = "beta"
-RELEASE_PHASE_RC = "rc"
-RELEASE_PHASE_PREVIEW = "preview"
-RELEASE_PHASE_POST = "post"
-RELEASE_PHASE_REV = "rev"
-RELEASE_PHASE_DEV = "dev"
-RELEASE_PHASES = {
-    RELEASE_PHASE_ALPHA: "a",
-    RELEASE_PHASE_BETA: "b",
-    RELEASE_PHASE_RC: "c",
-    RELEASE_PHASE_PREVIEW: "pre",
-    RELEASE_PHASE_POST: "-",  # shorthand of 1.2.3-post1 is 1.2.3-1
-    RELEASE_PHASE_REV: "r",
-    RELEASE_PHASE_DEV: "dev",
+# Release phase IDs according to PEP440
+RELEASE_PHASE_ID_ALPHA = "a"
+RELEASE_PHASE_ID_BETA = "b"
+RELEASE_PHASE_ID_RC = "rc"
+RELEASE_PHASE_ID_POST = "post"
+RELEASE_PHASE_ID_DEV = "dev"
+
+RELEASE_PHASE_SPELLINGS = {
+    RELEASE_PHASE_ID_ALPHA: {RELEASE_PHASE_ID_ALPHA, "alpha"},
+    RELEASE_PHASE_ID_BETA: {RELEASE_PHASE_ID_BETA, "beta"},
+    RELEASE_PHASE_ID_RC: {RELEASE_PHASE_ID_RC, "c", "pre", "preview"},
+    RELEASE_PHASE_ID_POST: {RELEASE_PHASE_ID_POST, "r", "rev", "-"},
+    RELEASE_PHASE_ID_DEV: {RELEASE_PHASE_ID_DEV},
 }
-RELEASE_PHASES_NORMALIZED = {
-    RELEASE_PHASE_ALPHA: "a",
-    RELEASE_PHASE_BETA: "b",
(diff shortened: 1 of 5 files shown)


abn:
My 2c, I would say we return PEP440 compliant string when short is true rather than introducing another property. I'd like to avoid adding more cognitive load here and standardise one way.

radoering:
That was my first thought, too. But it seemed a bit confusing later.

It would be a bit weird especially for `ReleaseTag` because there is a `shorten()` and an `expand()` method. At the moment each long name has its short name (`rev` <-> `r`, `post` <-> `-`). But for PEP440 normalization `rev` as well as `post` are normalized to `post`. So if you call `shorten()` and `expand()` on `rev` you don't get `rev` back if we make the proposed change. That may be a bit confusing and even increase the cognitive load.

I'm not sure whether the two different variants (short and long) are useful at all. I tend to say that `to_string()` should always return a normalized version string (no `short` parameter). If that would be too much a breaking change, I'd say it may be better to introduce `norm_string` and maybe deprecate `to_string()`.

Further, I'm not happy with `short` being a bad parameter name for `normalized`.

src/poetry/core/version/pep440/version.py
```diff
@@ -142,6 +142,29 @@ def non_semver_parts(self) -> tuple[int, ...]:
         assert isinstance(self.release.extra, tuple)
         return self.release.extra
 
+    @property
+    def norm_string(self) -> str:
```
abn, line 146:
```suggestion
    def normalize(self) -> str:
```

This might be better served as a method.

abn: Although I would suggest we deprecate the `short` parameter in favour of a new parameter `normalize`, and implement the normalisation logic there. We should really just conform to PEP440 as this is the standard we adhere to. 

Since `rev` exists mostly to support legacy package versions, I'd say we support reading `-` but we always write to PEP440 format in core.

tests/utils/test_helpers.py
```diff
@@ -9,11 +9,80 @@
 
 from poetry.core.utils.helpers import canonicalize_name
 from poetry.core.utils.helpers import combine_unicode
+from poetry.core.utils.helpers import normalize_version
 from poetry.core.utils.helpers import parse_requires
 from poetry.core.utils.helpers import readme_content_type
 from poetry.core.utils.helpers import temporary_directory
 
 
+@pytest.mark.parametrize(
```
abn, line 18: :+1:

src/poetry/core/version/pep440/version.py
radoering: I changed the property to a method.

I'm not convinced of introducing a new parameter because it makes `to_string()` more complicated and I think everything else than normalization is not really useful.

In a second commit, I changed `to_string()` to always return a normalized string according to PEP 440. I deprecated `short `because it has no effect anymore.

I think we can either go with both commits or only with the first one. What do you think about it?

abn: I think we are going in the right direction. The caveat here seems that some projects might loose readability when poetry cli is used to update version. For example, in a case where the project had `version = "1.0.0.alpha1"`, will be bumped to `1.0.0.a1`. Personally, I would have liked to retain the expanded version if it was used as the original here is more human friendly.

radoering: On the one hand it may be a bit more readable, but on the other hand it is a bit confusing that `version` is not PEP 440 normalized. I have wondered about this myself.

In long term, we will change it anyway, won't we? [PEP 621](https://peps.python.org/pep-0621/#version) says _"Users SHOULD prefer to specify already-normalized versions."_ It's only "should" not "must", but I see an advantage in sticking to a form that people are used to see everywhere else. So if we change it anyway, I'd question if we have to put effort in keeping two variants for some time?

By the way, it would be `1.0.0a2`. (Dots are only for `post` and `dev`, not for pre releases.)

src/poetry/core/version/pep440/segments.py
```diff
@@ -7,23 +7,23 @@
...
-    RELEASE_PHASE_DEV: "dev",
+# Release phase IDs according to PEP440
+RELEASE_PHASE_ID_ALPHA = "a"
+RELEASE_PHASE_ID_BETA = "b"
+RELEASE_PHASE_ID_RC = "rc"
+RELEASE_PHASE_ID_POST = "post"
+RELEASE_PHASE_ID_DEV = "dev"
+
+RELEASE_PHASE_SPELLINGS = {
+    RELEASE_PHASE_ID_ALPHA: {"a", "alpha"},
```
abn, line 18: Might be good to reuse the vars here for "a" here and similar for the rest. Nitpick.

abn (commented):
Very minor comment. Otherwise lgtm.

abn (approved):

PR #345, opened 2022-05-06
fix parsing PEP 508 url requirements with extras

abn:
Discovered this issue while implementing https://github.com/python-poetry/poetry/pull/5554.


src/poetry/core/packages/dependency.py
```diff
@@ -576,7 +576,7 @@ def create_from_pep_508(
                     name, "git", link.url_without_fragment, extras=req.extras
                 )
             elif link.scheme in ["http", "https"]:
-                dep = URLDependency(name, link.url)
+                dep = URLDependency(name, link.url, extras=req.extras)
```
neersighted, line 579: Whoops!

neersighted (approved):

PR #347, opened 2022-05-08
Fix get_python_constraint_from_marker

radoering:

Fixes `get_python_constraint_from_marker()` for markers which are not in disjunctive normal form (DNF).

E.g. without the fix:

`(python_version < "2.7" or python_full_version >= "3.0.0") and python_full_version < "3.6.0"` results in python_constraint `*`
whereas `<2.7 || >=3.0,<3.6` would be correct.

By the way, increased the test coverage for `convert_markers()` and `get_python_constraint_from_marker()`.

Considering the commits separately may facilitate the review.

**Update:**

There was another bug in `get_python_constraint_from_marker()` that is fixed by the last commit.
E.g. for `python_version >= "3.9" or sys_platform == "linux"` the function returned `>=3.9` whereas `*` would be correct (because for `linux` all python versions are allowed).

Further, relying on the marker being in DNF before conversion, allows to simplify `convert_markers()` significantly.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "poetry-core"
-version = "1.1.0-alpha.7"
+version = "1.1.0-alpha.8"
 description = "Poetry PEP 517 Build Backend"
 authors = ["Sébastien Eustace <sebastien@eustace.io>"]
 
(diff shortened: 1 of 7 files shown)


abn:
@radoering https://github.com/python-poetry/poetry/pull/5579

abn (changes requested):

abn (approved):

PR #353, opened 2022-05-12
typechecking semver.version_union

dimbleby:
I believe that if we can get the MRs merged, this will complete the typechecking campaign in poetry-core 🎉 .

Specifically:
- #274
- #339
- #340
- this

(though it does look as though #339 may have some non-trivial conflicts at the moment).

It would be a stretch to call this a blocker for the next release; but it would be awfully nice to get this mini-project completed.

### DIFF
--- a/src/poetry/core/semver/version_union.py
+++ b/src/poetry/core/semver/version_union.py
@@ -68,7 +68,9 @@ class VersionUnion(VersionConstraint):
             ):
                 merged.append(constraint)
             else:
-                merged[-1] = merged[-1].union(constraint)
+                new_constraint = merged[-1].union(constraint)
+                assert isinstance(new_constraint, VersionRangeConstraint)
+                merged[-1] = new_constraint
 
         if len(merged) == 1:
             return merged[0]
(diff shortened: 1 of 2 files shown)


src/poetry/core/semver/version_union.py
```diff
@@ -302,26 +310,31 @@ def _excludes_single_wildcard_range_check_is_valid_range(
         wildcard range.
         """
 
-        max_precision = max(one.max.precision, two.min.precision)
+        one_max = one.max
+        assert one_max is not None
+        two_min = two.min
+        assert two_min is not None
+
```
abn, line 317: Can't we just replace this and avoid the reassignment?

```suggestion
        assert one.max is not None
        assert two.min is not None

```

dimbleby: huh, I'm slightly surprised to find that you can.  Not sure how mypy is deciding that what is true of `one.max` at some point is necessarily also true at some later point, but it doesn't complain.

dimbleby:
@abn are we waiting on anything to get this merged?

neersighted (approved):

PR #354, opened 2022-05-15
Mypy/packages rebased

dimbleby:
I got excited about completing the typechecking in poetry-core, and took the liberty of rebasing #339.  This MR now removes the allow-mypy-errors list!  

I also made a handful of tweaks per my review of that MR.  I have tried to keep @branchvincent's work and my own separate, though I did have to decide what to do about conflicts in the rebase.

The change relating to https://github.com/python-poetry/poetry-core/pull/339#discussion_r873153018 might be worth some thought.  Per that comment, the original code looks confused so I have tidied it up in a way that looks sensible and doesn't break any tests.  But if anyone has a clearer idea than I do about what was intended, I could believe that a different tidying might be better.

### DIFF
--- a/src/poetry/core/packages/dependency.py
+++ b/src/poetry/core/packages/dependency.py
@@ -8,7 +8,6 @@ from pathlib import Path
 from typing import TYPE_CHECKING
 from typing import Any
 from typing import Iterable
-from typing import cast
 
 from poetry.core.packages.constraints import (
     parse_constraint as parse_generic_constraint,
@@ -408,7 +407,7 @@ class Dependency(PackageSpecification):
                 if not constraint.include_min:
                     op = ">"
 
-                version: Version | str = constraint.min.text
+                version = constraint.min.text
                 if constraint.max is not None:
                     max_name = name
                     if constraint.max.precision >= 3 and name == "python_version":
@@ -420,7 +419,7 @@ class Dependency(PackageSpecification):
                     if not constraint.include_max:
                         op = "<"
 
-                    version = constraint.max
+                    version = constraint.max.text
 
                     text += f' and {max_name} {op} "{version}"'
 
@@ -433,7 +432,7 @@ class Dependency(PackageSpecification):
                 if not constraint.include_max:
                     op = "<"
 
-                version = constraint.max
+                version = constraint.max.text
             else:
                 return ""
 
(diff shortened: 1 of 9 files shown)


src/poetry/core/packages/utils/utils.py
```diff
@@ -172,9 +178,10 @@ def add_constraint(
     for group_name in requirements:
         # remove duplicates
         seen = []
-        requirements[group_name] = [
-            r for r in requirements[group_name] if not (r in seen or seen.append(r))
-        ]
+        for r in requirements[group_name]:
```
neersighted, line 181: Is there a typing reason we can't use the list comprehension here?

src/poetry/core/packages/dependency.py
```diff
@@ -559,8 +553,7 @@ def create_from_pep_508(
                 name = m.group("name")
                 version = m.group("ver")
 
-            name = req.name or link.egg_fragment
```
neersighted, line 562: I am pretty sure this is just dead code -- more investigation and a possible removal of the field would be welcome.

neersighted (commented):
This looks really good! I have a couple questions, but overall this looks good to merge once those are addressed.

src/poetry/core/packages/utils/utils.py
dimbleby: yeah, it's a bit annoying but it complains that `seen.append(r)` does not return a value

src/poetry/core/packages/dependency.py
dimbleby: I reached the same conclusion - and removed it! are you asking me to do something more?

dimbleby: oh, you mean the `egg_fragment`?  I decided that removing that might make it harder to get this MR merged, so left it for another day.

src/poetry/core/packages/utils/utils.py
neersighted: I'm not thrilled about that as I find it less readable, but hopefully mypy matures and a future refactor cleans this up. I for one am always rewriting loops as comprehensions where it adds readability, because it's (very moderately) more performant as well.

neersighted (approved):

neersighted:
We are now fully typed and mypy clean! Now that we have a py.typed, consumers should be able to fully typecheck against the poetry-core API with the next release! ✨

PR #355, opened 2022-05-17
build: update dev deps

branchvincent:
Just some miscellaneous cleanup to:

- Update dev dependencies
- Simplify vendoring config and re-sync
- Address our own [deprecation warning](https://github.com/python-poetry/poetry-core/runs/6442139532?check_suite_focus=true#step:12:22) when running tests


src/poetry/core/_vendor/jsonschema/LICENSE
```diff
@@ -1,19 +0,0 @@
-Copyright (c) 2012 Julian Berman
```
branchvincent, line 1: `vendoring` removed this, but note there is already a [`COPYING`](https://github.com/python-poetry/poetry-core/blob/master/src/poetry/core/_vendor/jsonschema/COPYING)

branchvincent:
Downstream test failure is due to https://github.com/pypa/setuptools/issues/3320

abravalheri:
Sorry for the disruption. Soon the new setuptools version will be available with the fix for the unexpected error (it should only be a warning).
Please let me know if it does not work.

tests/version/test_version_pep440.py
```diff
@@ -25,6 +25,7 @@ def test_pep440_release_segment_from_parts(
     assert Release.from_parts(*parts) == result
 
 
+@pytest.mark.filterwarnings("ignore::DeprecationWarning")
```
abn, line 28: Better to remove the `short` parameter.

abn (changes requested):

tests/version/test_version_pep440.py
branchvincent: oops, of course 😅

abn (approved):

PR #357, opened 2022-05-17
Update lark and jsonschema

dimbleby:
More vendoring updates

- set the `SHELL` in the Makefile.  Else, at least for me, it runs `/bin/sh` and doesn't know anything about `pushd`
- update `lark`
- don't need to explicitly require pyrsistent in pyproject.toml, it's an unconditional dependency of jsonschema

I started trying to update jsonschema too but it looks more complicated:

- we have a patch, which might or might not need updating to apply to the latest code.  (the jsonschema code has certainly changed, but I didn't investigate)
- recent jsonschema has dependencies that vary according to python version, I don't know how that's supposed to be handled in vendoring

so I took the cowardly option and left jsonschema alone.

To be honest I don't fully understand why vendoring is required here at all, though I'm ready to believe that there's a reason.  For my education, is this written down somewhere?


dimbleby:
Figured out what to do about jsonschema after all

(updated our patch so that it applies cleanly to the latest version, it looks rather as though the point of the patch is to avoid the extra dependencies that I wasn't sure what to do about)

neersighted:
@abn or @sdispater can probably authoritatively surmise why we currently vendor deps.

I am under the impression it is to speed up builds when used as a PEP 517 backend, as well as to ensure that the core is robust and not broken by users updating their site-packages or similar. I believe PEP 517 backends are installed in a clean venv by pip, so the latter is more a feature for installs that may not be managed by Poetry itself/pipx.

dimbleby:
Ah, but there's a new python3.7-only dependency on typing-extensions... hmm

dimbleby:
crikey, what is pre-commit trying to do to the patchfile...?

dimbleby:
I think I'm done, the commit history tells the tale

abn:
> I am under the impression it is to speed up builds when used as a PEP 517 backend, as well as to ensure that the core is robust and not broken by users updating their site-packages or similar. I believe PEP 517 backends are installed in a clean venv by pip, so the latter is more a feature for installs that may not be managed by Poetry itself/pipx.

This covers most of the story. Backends essentially are better served to be standalone and cannot rely on the frontend resolving the dependencies when the build environment is created. Additionally, one cannot rely on external network access in build envionrments either.

That said, I personally think except for lark - the rest of the vendored packages should eventually be removed. Because imho, schema validation should be done during development or as an additional "extra" step rather than a default feature.

As for the patch for jsonschema, the intent mentioned above is accurate. We did not want to add more vendors just for unused compatibility code.

abn:
> Ah, but there's a new python3.7-only dependency on typing-extensions... hmm

At somepoint we made the decision that we will discard compat depenencies from vendors to not overly bloat the package. That said, this is not idea.

The way we vendor today is also a source of headaches and annoying bugs. See https://github.com/python-poetry/poetry/issues/5225#issuecomment-1128865908 for example.

Makefile
```diff
@@ -1,3 +1,4 @@
+SHELL                           := /bin/bash
```
abn, line 1:
```suggestion
SHELL                           := $(shell which bash) -e
```

dimbleby:
re typing-extensions, jsonschema really does rely on it - there's not a trivial patch to disentangle this.  I'd hoped to put behind `if TYPECHECKING:` or something, but see use of `runtime_checkable` in https://github.com/python-jsonschema/jsonschema/blob/main/jsonschema/protocols.py#L23.

As for bloat, I figure typing_extensions is pretty small in the context of all this vendoring - it's just one file, albeit 2000 lines

neersighted (approved):

abn:
Thanks for doing the heavy lifting on this one @dimbleby.

PR #361, opened 2022-05-21
More type fixes

dimbleby:
Companion to https://github.com/python-poetry/poetry/pull/5659: some of the imports that we have been ignoring have type stubs, let's use them.

Also picked off a couple missed in #359

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -506,6 +506,14 @@ optional = false
 python-versions = "*"
 
 [[package]]
+name = "types-setuptools"
+version = "57.4.14"
+description = "Typing stubs for setuptools"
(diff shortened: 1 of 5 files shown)


branchvincent (approved):

branchvincent:
Looks like I can't merge this myself since the downstream test is `Required`

dimbleby:
> Looks like I can't merge this myself since the downstream test is Required

Yeah, I believe https://github.com/python-poetry/poetry/pull/5654 is expected to resolve this.

Perhaps `poetry` pipelines should also have a "downstream tests" section in which they run against the latest `poetry-core`...

PR #365, opened 2022-05-23
packages: add helper methods for spec, package and dependency

abn:
This change introduces various helper methods to improve usablity downstream in poetry.

### DIFF
--- a/src/poetry/core/packages/specification.py
+++ b/src/poetry/core/packages/specification.py
@@ -71,6 +71,14 @@ class PackageSpecification:
     def features(self) -> frozenset[str]:
         return self._features
 
+    def is_direct_origin(self) -> bool:
+        return self._source_type in [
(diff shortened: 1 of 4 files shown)


src/poetry/core/packages/specification.py
```diff
@@ -71,62 +71,90 @@ def source_subdirectory(self) -> str | None:
...
+        return bool(
+            self._source_type
+            and self._source_type
+            in [
+                "directory",
+                "file",
+                "url",
+                "git",
+            ]
+        )
```
radoering, line 84: simplification:

```suggestion
        return self._source_type in [
            "directory",
            "file",
            "url",
            "git",
        ]
```

radoering (changes requested):

radoering (approved):

PR #366, opened 2022-05-23
update version

dimbleby:
It would have been more helpful if I'd noticed this yesterday, wouldn't it?

### DIFF
--- a/src/poetry/core/__init__.py
+++ b/src/poetry/core/__init__.py
@@ -1,11 +1,23 @@
 from __future__ import annotations
 
+import contextlib
 import sys
 
 from pathlib import Path
 
 
-__version__ = "1.1.0a7"
+__version__ = "0.0.0+in-tree"
+
+with contextlib.suppress(ImportError):
+    if sys.version_info < (3, 8):
+        # compatibility for python <3.8
+        import importlib_metadata as metadata
+    else:
+        from importlib import metadata
+
+    with contextlib.suppress(metadata.PackageNotFoundError):
+        __version__ = metadata.version("poetry-core")  # type: ignore[no-untyped-call]
+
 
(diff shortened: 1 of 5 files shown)


dimbleby:
(perhaps we should add something to the pipeline to verify that this version matches the version as defined elsewhere)

abn:
We should really drop `__version__`.

src/poetry/core/__init__.py
```diff
@@ -5,7 +5,7 @@
 from pathlib import Path
 
 
-__version__ = "1.1.0a7"
+__version__ = "1.1.0b1"
```
abn, line 8:
```suggestion
```

We have been promoting the use of importlib elsewhere for this use case. We should really dogfood our own advice on the matter and either remove this or use importlib/pkgutil.

dimbleby: I'm not fussed either way - feel free either to update this MR or to reject it in favour of something else

abn: Let's see what the rest of the team thinks. My vote is to just remove it.

neersighted: Let's just drop it.

src/poetry/core/__init__.py
```diff
@@ -5,7 +5,7 @@
...
```
abn, line 9:
```suggestion
```

src/poetry/core/__init__.py
abn: Hmm, apparently not. We imprint the metadata for wheels with the version it seems.

neersighted (approved):

abn:
@dimbleby sorry for the noise.

PR #367, opened 2022-05-24
bump poetry.core.__version__ with test coverage

abn:
As discussed with @neersighted on discord, going with the simple bump as `importlib.metadata.version` is a bit more tricky to do here. And if we want to drop `__version__` we need to drop it's use in wheel builder and/or accept that version will not be available for in-tree builds unless we parse `pyproject.toml` file.
 
 Closes: #366


finswimmer (approved):
LGTM :+1:

PR #370, opened 2022-05-24
packages: make hash() consistent and incl features

abn:


src/poetry/core/packages/package.py
```diff
@@ -559,7 +559,11 @@ def clone(self: T) -> T:
         return clone
 
     def __hash__(self) -> int:
-        return super().__hash__() ^ hash(self._version)
+        return (
+            super().__hash__()
+            ^ hash(self._version)
+            ^ hash(frozenset(self.extras.keys()))
```
radoering, line 565: If we use `extras` in `__hash__`, we have to use it in `__eq__`, too. Actually, `extras` should probably not be used in hash because they can change and thus the hash...

src/poetry/core/packages/package.py
```diff
@@ -559,7 +559,11 @@ def clone(self: T) -> T:
...
+        )
 
     def __eq__(self, other: object) -> bool:
```
radoering, line 568: _"Hashable objects which compare equal must have the same hash value."_ ([source](https://docs.python.org/3/glossary.html#term-hashable)) That's not fulfilled here. Since `__eq__` relies on `is_same_package_as()` and that can be fulfilled even if `_source_reference` attributes are not equal and `_source_reference` is used in hash.

radoering (changes requested):

src/poetry/core/packages/specification.py
```diff
@@ -152,18 +152,18 @@ def is_same_package_as(self, other: PackageSpecification) -> bool:
         return self.is_same_source_as(other)
 
     def __hash__(self) -> int:
```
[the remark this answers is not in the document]
abn: Ack. Compeltely forgot about `__eq__`. Unsure if we really need it, but we should probably just do it anyway.

src/poetry/core/packages/package.py
radoering: I think, there are two options:

1. Don't use `is_same_package_as()` in `__eq__` and just compare everything that is used in `__hash__`.
2. Only use attributes in `__hash__` that have to be equal according to `is_same_package_as()`. That probably means, we can't use `_source_reference` and `_resolved_source_reference` in hash.

I suppose 2. might be ok. 1. seems to be a breaking change.

radoering:
Just a quick thought when deciding what to put into the hash for each class. In my understanding, it's not required to consider each attribute in the hash even if it is considered for equality. E.g., consider the following slightly exaggerated example:

```
class A:
    def __init__(self, a):
        self.a = a

    def __hash__(self):
        return 0

    def __eq__(self, other):
        return self.a == other.a

    def __repr__(self):
        return f'A({self.a})'


s = {A(0), A(1), A(2)}
print(s)
```
Output: `{A(0), A(1), A(2)}`

It might be inperformant, but it seems to be valid even though every instance has the same hash.

Thus, I tend to say we have to decide for each attribute:

- relevant for equality? if yes -> use in `__eq__`
- if used in `__eq__`, also immutable? if yes -> use in `__hash__`

In other words, two (simplified) rules that must not be broken:

- attributes used in `__hash__` must be used in `__eq__`
- attributes used in `__hash__` must be immutable

abn:
Split out the attribute change to https://github.com/python-poetry/poetry-core/pull/375

src/poetry/core/packages/dependency.py
```diff
@@ -111,6 +113,15 @@ def set_constraint(self, constraint: str | VersionConstraint) -> None:
             self._constraint = parse_constraint("*")
         self._pretty_constraint = str(constraint)
 
+    def set_constraint(self, constraint: str | VersionConstraint) -> None:
```
abn, line 116: :+1:

src/poetry/core/packages/specification.py
```diff
@@ -151,19 +151,24 @@ def is_same_package_as(self, other: PackageSpecification) -> bool:
...
-            ^ hash(self._source_url)
-            ^ hash(self._source_reference)
-            ^ hash(self._source_resolved_reference)
-            ^ hash(self._source_subdirectory)
-            ^ hash(self._features)
-        )
+        result = hash(self.complete_name)  # complete_name includes features
+
+        if self._source_type:
+            # don't include _source_reference and _source_resolved_reference in hash
```
abn, line 163: I do not think a spec with reference "master" and "v1.0.0" should be considered equal as they are inherently different. I wonder if resovled reference is what we do not need.

radoering: Your example is not considered equal. 😉 See https://github.com/python-poetry/poetry-core/blob/6b45d6c3977d1963d7a247ebbc64d13e7fc2879e/src/poetry/core/packages/specification.py#L112-L146

However, this logic allows references to be different within certain limitations:

- If `source_reference` is not set, two specs with arbitrary `resolved_source_reference` can be equal. (Might be a theoretical use case.)
- If `source_reference` is set, it is allowed that one starts with the other if `resolved_source_reference` is equal.

See the test added in `test_specification.py` for example. (I just added a sentence for clarity in the comment and another test for the theoretical use case.)

radoering: Wait a minute, this logic is partially redundant and can be further simplified.

radoering: OK, nothing redundant in the logic. The test now checks each variant that is considered equal. Not sure if the logic for references in `is_same_source_as()` completely makes sense, but that may be deferred to another PR. I suppose for now, it may be "good enough" (since the focus of this PR is consistency between `__eq__` and `__hash__`).

If the logic is changed later, we can still consider to include additional attributes into the hash.

abn: I am thinking of the caches we have in poetry at present. These might resolve incorrectly. For example if provider searches for `(A git rev master)` and `(A git rev v1.0.0)` the cache would incorrectly return the package for the first one if I am groking the above correctly as the dict is using `hash(dep)`.

radoering: Nope, `set` and `dict` are safe considering hash collisions. A set can hold different objects with same hashes.

See extreme example:

```
class A:
    def __init__(self, a):
        self.a = a

    def __hash__(self):
        return 0

    def __eq__(self, other):
        return self.a == other.a

    def __repr__(self):
        return f'A({self.a})'


s = {A(0), A(1), A(2)}
d = {A(0): 0, A(1): 1, A(2): 2}
print(s)
print(d[A(0)])
print(d[A(2)])
```

Output:
```
{A(0), A(1), A(2)}
0
2
```

abn (commented):
Changes look good to me. Someone else might need to approve.

radoering (approved):

PR #373, opened 2022-05-25
add Dependency.set_extras()

dimbleby:
Will allow tidying of https://github.com/python-poetry/poetry/pull/5688, which currently is manipulating `_extras` directly.

Again the split into multiple repositories makes it complicated to figure out a merge sequence that will work...

### DIFF
--- a/src/poetry/core/packages/dependency.py
+++ b/src/poetry/core/packages/dependency.py
@@ -244,6 +244,9 @@ class Dependency(PackageSpecification):
     def extras(self) -> frozenset[str]:
         return self._extras
 
+    def set_extras(self, extras: Iterable[str]) -> None:
+        self._extras = frozenset(extras)
+
     @property
     def in_extras(self) -> list[str]:
         return self._in_extras
--- a/tests/packages/test_dependency.py
+++ b/tests/packages/test_dependency.py
@@ -285,6 +285,13 @@ def test_dependency_string_representation(
     assert str(dependency) == expected
 
 
+def test_set_extras_sets_extras() -> None:
+    dependency = Dependency("A", "^1.0")
+    assert not dependency.extras
+    dependency.set_extras(["foo", "bar"])
+    assert dependency.extras == frozenset(["foo", "bar"])
+
+
 def test_set_constraint_sets_pretty_constraint() -> None:
     dependency = Dependency("A", "^1.0")
     assert dependency.pretty_constraint == "^1.0"


src/poetry/core/packages/dependency.py
```diff
@@ -244,6 +244,9 @@ def transitive_python_constraint(self) -> VersionConstraint:
     def extras(self) -> frozenset[str]:
         return self._extras
 
+    def set_extras(self, extras: Iterable[str]) -> None:
```
abn, line 247: I wonder if it would be better to follow the `with_extras()` pattern similar to `package.with_features()`. The whole extras / features thing is a bit confusing at present. 

Because you are doing this anyway, I would suggest you roll in the removal of the `Dependency._extra` attribute in favour of the `PacakgeSpecification._feature` change from https://github.com/python-poetry/poetry-core/pull/370/files#diff-93744eaedc27b829eefbde40b5d45732da68e0716af35c30bf010e66ea72b1f7R84-R245 here too.

dimbleby: I wondered about `with_extras()`, but then we ought to do `with_constraint()` too and it all seemed like a lot of trouble for no very clear benefit.

#370 looks a bit hairy, happy to wait for the dust to settle on it before pushing forward on this one but I don't think I want to get into doing both at once...

radoering: We should really be careful that we don't tear down right away what will be cleaned up in #370. If we decide to use an attribute in the hash, it should not be changed anywhere. Thus, the `with_attribute()` pattern is required.

abn: The attribute change can be cleanly split out of #370. I can do it in another PR if you do not want to add it to this one, easy enough.

We can handle `with_constraint()` implementation in another PR as the need arises or once we settle #370.

dimbleby: You say required, but we already don't follow that in `set_constraint()`, and the single place that calls that (and would call this new setter) is safe.

So I only sort of agree...

abn: Re: attribute removal https://github.com/python-poetry/poetry-core/pull/375

dimbleby:
Actually I might be changing my mind about the best way to fix this...

The current approach is vulnerable to `foo[bar]` being required on Windows but only `foo` elsewhere.  Then the exporter would show `foo[bar]` everywhere. 

I previously rejected a simpler fix that resulted in an export like:
```
localstack-ext==1.0.0 ; python_version >= "3.6" and python_version < "4.0"
localstack-ext[bar]==1.0.0 ; python_version >= "3.6" and python_version < "4.0"
```
but actually while that's cosmetically awkward, it seems like it should give the right answers and be more reliably right.

Rework coming, and in that case https://github.com/python-poetry/poetry-core/pull/373 will be redundant anyway

abn:
That resembles the original fix I was playing with that led me down the hash rabbit hole. But yeah I think that is more accurate anyway. It is upto whoevever consumes the export to decide how to handle it I reckon.

src/poetry/core/packages/dependency.py
radoering: "Required" for clean code I assume.

That we don't follow the pattern somewhere does not mean that this is alright. Of course, we could allow setting attributes that are used in hash and evaluate each place that sets the attribute. But you don't have to be a prophet to see that eventually someone will forget that it's dangerous to set this attribute... Thus, in my opinion it may be better not to provide a setter but just a `with_attribute()` (or don't use the attribute in `__hash__()`).

PR #374, opened 2022-05-25
remove Package.requires_extras

dimbleby:
If and when https://github.com/python-poetry/poetry/pull/5688 is merged, there'll be no use for this field. 

(Even today there is no _good_ use for this field)


abn:
I suspect the usage in Poetry at https://github.com/python-poetry/poetry/blob/a13ccf887057445560c8302f3c0668d1a49a524e/src/poetry/packages/locker.py#L373 should be replaced prior to merging this.

abn (approved):

PR #376, opened 2022-05-26
Enable more `mypy` optional error codes and bump `mypy` to `0.960`

mkniewallner:
- [ ] ~Added **tests** for changed code.~ Not applicable

Similarly to https://github.com/python-poetry/poetry/pull/5665, enable [truthy-bool](https://mypy.readthedocs.io/en/stable/error_code_list2.html#check-that-expression-is-not-implicitly-true-in-boolean-context-truthy-bool) and [redudant-expr](https://mypy.readthedocs.io/en/stable/error_code_list2.html#check-that-expression-is-redundant-redundant-expr).
Also update `mypy` to `0.960`.

_Note: I also needed to update the tests CI workflow to use PyPy 3.8 instead of PyPy 3.7, which is what we used to have in the past. See comments for the details._

### DIFF
--- a/src/poetry/core/semver/version.py
+++ b/src/poetry/core/semver/version.py
@@ -92,8 +92,6 @@ class Version(PEP440Version, VersionRangeConstraint):
         # allow weak equality to allow `3.0.0-1` for `3.0.0`
         if not _this.is_postrelease() and _other.is_postrelease():
             _other = _other.without_postrelease()
-        elif _this.without_postrelease() and not _other.without_postrelease():
-            _this = _this.without_postrelease()
 
         return _this == _other
 
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -18,7 +18,7 @@ jobs:
         bootstrap-args: [""]
         include:
           - os: Ubuntu
-            python-version: pypy-3.7
+            python-version: pypy-3.8
             experimental: false
           - os: Ubuntu
             python-version: "3.10"
--- a/poetry.lock
+++ b/poetry.lock
@@ -101,14 +101,14 @@ test = ["flake8 (==3.7.8)", "hypothesis (==3.55.3)"]
 
 [[package]]
 name = "coverage"
-version = "6.3.3"
+version = "6.4"
 description = "Code coverage measurement for Python"
 category = "dev"
 optional = false
 python-versions = ">=3.7"
 
 [package.dependencies]
-tomli = {version = "*", optional = true, markers = "extra == \"toml\""}
+tomli = {version = "*", optional = true, markers = "python_version < \"3.11\" and extra == \"toml\""}
 
 [package.extras]
 toml = ["tomli"]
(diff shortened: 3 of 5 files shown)


src/poetry/core/semver/version.py
```diff
@@ -92,8 +92,6 @@ def allows(self, version: Version | None) -> bool:
         # allow weak equality to allow `3.0.0-1` for `3.0.0`
         if not _this.is_postrelease() and _other.is_postrelease():
             _other = _other.without_postrelease()
-        elif _this.without_postrelease() and not _other.without_postrelease():
-            _this = _this.without_postrelease()
```
mkniewallner, line 96: Since `Version` doesn't provide a `__bool__` nor `__len__` method, the condition would always evaluate to `True`, as reported by `mypy`:
```shell
"without_postrelease" returns "Version" which does not implement __bool__ or __len__ so it could always be true in boolean context  [truthy-bool]
```

So we never entered this branch, since `not _other.without_postrelease()` would always evaluate to False, meaning this part of the code is either unnecessary or incorrect. My guess is that this is the former, but I'm definitely not sure.

mkniewallner:
It looks like PyPy 3.7 has troubles installing [typed-ast](https://pypi.org/project/typed-ast/).

It seems to only fail on this pull request because we are updating dependencies here, which invalidates the cache.
Potentially, any PR updating dependencies may end up with failing PyPy tests.

Will try to bump `typed-ast` to see if it fixes the problem.

src/poetry/core/semver/version.py
```diff
@@ -92,8 +92,6 @@ def allows(self, version: Version | None) -> bool:
         # allow weak equality to allow `3.0.0-1` for `3.0.0`
         if not _this.is_postrelease() and _other.is_postrelease():
             _other = _other.without_postrelease()
-        elif _this.without_postrelease() and not _other.without_postrelease():
```
abn, line 95: Think this should be

```py
elif _this.is_postrelease() and not _other.is_postrelease():
```

mkniewallner: This is what I originally assumed, but this made the following 2 tests fail in [test_allows_with_post](https://github.com/mkniewallner/poetry-core/blob/f26e2361f0d643826d31366887dcb425fc078214/tests/semver/test_version.py#L134-L141):
```python
assert not v.allows(Version.parse("1.2.3"))
assert not v.allows(Version.parse("1.2.3+build.2"))
```
Which led me to believe that the current tests are correct, and that this code branch may well be unnecessary.

abn: I wonder if that is exposing https://github.com/python-poetry/poetry/issues/4729#issuecomment-1110930059

abn: But that said `1.2.3-1` shouldn't in fact allow `1.2.3` or `1.2.3+local`, so you might be correct.

abn (approved):

PR #368, opened 2022-05-24
deps: support valid PEP 517 projects as dir deps

abn:
Resolves: python-poetry/poetry#5655
Resolves: https://github.com/python-poetry/poetry/issues/5692

```sh
pipx install --force --suffix=@5655 'poetry @ git+https://github.com/python-poetry/poetry.git'
pipx inject --force --pip-args="--force --no-deps" poetry@5655 'poetry-core @ git+https://github.com/python-poetry/poetry-core.git@refs/pull/368/head'
poetry@5655 --version
```

PR #377, opened 2022-05-26
packages: unify python project check

abn:
Following up from #368 in relation to https://github.com/python-poetry/poetry/issues/5692

Seems we were only checking for `setup.py` when determining if a relative path dependency was installable.

### DIFF
--- a/src/poetry/core/packages/utils/utils.py
+++ b/src/poetry/core/packages/utils/utils.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-import os
+import functools
 import posixpath
 import re
 import sys
@@ -14,6 +14,7 @@ from urllib.parse import unquote
 from urllib.parse import urlsplit
 from urllib.request import url2pathname
 
+from poetry.core.pyproject.toml import PyProjectTOML
 from poetry.core.semver.version_range import VersionRange
 from poetry.core.version.markers import dnf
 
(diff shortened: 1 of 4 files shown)


tests/packages/utils/test_utils.py
```diff
@@ -132,3 +135,19 @@ def test_get_python_constraint_from_marker(marker: str, constraint: str) -> None
...
+
+@pytest.mark.parametrize(
+    ("fixture", "result"),
+    [
+        ("simple_project", True),
+        ("project_with_setup_cfg_only", True),
+        ("project_with_setup", True),
+        ("project_with_pep517_non_poetry", True),
+        ("project_without_pep517", False),
+        ("does_not_exists", False),
```
neersighted, line 148:
```suggestion
        ("does_not_exist", False),
```

src/poetry/core/packages/utils/utils.py
```diff
@@ -119,14 +120,22 @@ def strip_extras(path: str) -> tuple[str, str | None]:
     return path_no_extras, extras
 
 
-def is_installable_dir(path: str) -> bool:
-    """Return True if `path` is a directory containing a setup.py file."""
-    if not os.path.isdir(path):
+@functools.lru_cache(maxsize=None)
```
neersighted, line 123: I'm curious how much of a hot path this is/if caching is a significant benefit here?

neersighted (changes requested):

src/poetry/core/packages/utils/utils.py
abn: My thinking is this is called every time a dir dependency gets cloned, this happens a few times during resolution. And since this requires IO figured it can't hurt.

neersighted (approved):

PR #379, opened 2022-05-26
fix(pep440): do not ignore post for inclusive ordering

abn:
## Summary by Sourcery

Bug Fixes:
- Fixed an issue where inclusive ordering with post releases were inconsistent with PEP 440.


tests/semver/test_version.py
```diff
@@ -127,7 +127,8 @@ def test_allows_with_local() -> None:
     assert not v.allows(Version.parse("1.3.3"))
     assert not v.allows(Version.parse("1.2.3-dev"))
     assert not v.allows(Version.parse("1.2.3+build.2"))
-    assert v.allows(Version.parse("1.2.3-1"))
+    assert not v.allows(Version.parse("1.2.3-1"))
+    assert v.allows(Version.parse("1.2.3-1+build.1"))
     assert v.allows(Version.parse("1.2.3-1+build.1"))
```
radoering, line 132: Duplicated line.

Maybe, it's worth adding `assert not v.allows(Version.parse("1.2.3"))` since that's new.

src/poetry/core/semver/version_range.py
```diff
@@ -69,6 +69,11 @@ def is_simple(self) -> bool:
 
     def allows(self, other: Version) -> bool:
         if self._min is not None:
+            if self._min.is_local() and (
+                not other.is_local() or self._min.local != other.local
```
radoering, line 73: That's not completely correct. It actually has to be something like  `self._min.local < other.local`.

src/poetry/core/semver/version_range.py
```diff
@@ -81,6 +86,10 @@ def allows(self, other: Version) -> bool:
             if not _this.is_local() and _other.is_local():
                 # allow weak equality to allow `3.0.0+local.1` for `<=3.0.0`
                 _other = _other.without_local()
+            elif _this.is_local() and (
+                not _other.is_local() or _this.local != _other.local
```
radoering, line 90: Same as above.

radoering (dismissed):
I don't think that's completely correct yet. [PEP440](https://peps.python.org/pep-0440/#local-version-identifiers) says:

_"Comparison and ordering of local versions considers each segment of the local version (divided by a .) separately. If a segment consists entirely of ASCII digits then that section should be considered an integer for comparison purposes and if a segment contains any ASCII letters then that segment is compared lexicographically with case insensitivity. When comparing a numeric and lexicographic segment, the numeric section always compares as greater than the lexicographic segment. Additionally a local version with a great number of segments will always compare as greater than a local version with fewer segments, as long as the shorter local version’s segments match the beginning of the longer local version’s segments exactly."_

It seems that comparing local version segments is already prepared here: https://github.com/python-poetry/poetry-core/blob/dfd1b39ffec8bfad027a6dda0c31bab24b4c6019/src/poetry/core/version/pep440/version.py#L108-L125

src/poetry/core/semver/version_range.py
abn: IIRC, local labels are to be treated as strings, so they should not be ordered when comparing.

abn: This means `>=1+local.1` includes `1+local.1` and `2+local.1` but not `1+local.2`.

radoering: Yeah, it might not be suitable for stuff like `1.10.0+cu113` and `1.10.0+cpu` but that's what PEP440 says.

dimbleby:
https://github.com/python-poetry/poetry/issues/5791#issuecomment-1148333841

abn:
@sourcery-ai review

radoering:
Actually, I am not sure if this fixes anything regarding compliance with PEP 440. I even think this will be less compliant with PEP 440 than before.

If feels like switching from one use case to another. According to PEP 440, local version labels are not allowed in constraints with `>`, `<`, `>=`, `<=`. However, since we are building version ranges, we have to deal with it anyway. Although they are not allowed in constraints, it is defined how they are ordered and compared, see https://github.com/python-poetry/poetry-core/pull/379#pullrequestreview-988611044

If I understand correctly, this PR tries to make something like this work:
`>= 1.0+cpu` does not allow `1.0+cuda` or `1.1+cuda` (even though they are both greater!) but `1.1+cpu` (which is also greater)

However, to make this work, it breaks the use case where `>=1.0+internal.1` allows `1.0+internal.2`.

Honestly, the second use case feels more correct (i.e. more compliant) than the first one. On the other side, I understand that the first one might be the more common one (at least when neglecting private packages inside of companies).

I think this change will be breaking for some users and make us less compliant with PEP 440. I feel like we should carefully weigh if it is worth it. Is there an open issue about this? The issue that is linked has already been resolved.

abn:
When through the spec again, been a while. And I agree that the original changes I had is swapping one "interpretation" for another. I also agree that `>=1.0+internal.1 allows 1.0+internal.2` is better than `>= 1.0+cpu does not allow 1.0+cuda` as the latter does not comply with [_if a segment contains any ASCII letters then that segment is compared lexicographically with case insensitivity_](https://peps.python.org/pep-0440/#local-version-identifiers).

Taking all that into considering, I have gone ahead and repurposed this PR to only include the following:
1. Test cleanups and new test cases.
2. Fixing inclusive ordering.

As for what to do with `cpu` vs `cuda`, I am more convinced that solving this at the version range level is not the appropriate fix if we want to ensure that `VersionRange.allows` adheres to PEP440 inclusive and exclusive ordering. We do bend the rules (see quote below) a bit already because we have to consider local versions.

> Except where specifically noted below, local version identifiers MUST NOT be permitted in version specifiers, and local version labels MUST be ignored entirely when checking if candidate versions match a given version specifier.

abn:
@sourcery-ai review

src/poetry/core/constraints/version/version_range.py
```diff
@@ -58,13 +58,25 @@ def is_any(self) -> bool:
     def is_simple(self) -> bool:
         return self._min is None or self._max is None
 
+    @property
+    def is_inclusive(self) -> bool:
+        return self._include_min or self.include_max
```
radoering, line 63: Isn't the naming a little ambiguous? Why `or` and not `and`? Further, the `or self.include_max` part seems not to be required by any test.

src/poetry/core/constraints/version/version_range.py
```diff
@@ -58,13 +58,25 @@ def is_any(self) -> bool:
...
+
     def allows(self, other: Version) -> bool:
         if self._min is not None:
             _this, _other = self.allowed_min, other
 
             assert _this is not None
 
-            if not _this.is_postrelease() and _other.is_postrelease():
+            if (
+                self.is_exclusive
```
radoering, line 76:
```suggestion
                not self._include_min
```

That should be sufficient concerning the comment below, shouldn't it? (`<V` will not allow a post-release anyway.)

tests/constraints/version/test_version_range.py
```diff
@@ -79,135 +79,68 @@ def v300b1() -> Version:
...
 
 @pytest.mark.parametrize(
-    "base,other",
+    ("constraint", "check_version", "allowed"),
     [
-        pytest.param(Version.parse("3.0.0-1"), Version.parse("3.0.0-1"), id="post"),
-        pytest.param(
-            Version.parse("3.0.0"), Version.parse("3.0.0+local.1"), id="local"
-        ),
+        # Inclusive ordering
```
radoering, line 84: This is really an improvement. 👍  One minor thing: Can we add all cases where `check_version` is the version in the constraint, e.g. `<=3.0.0+local1` with `3.0.0+local1`, `>=3.0.0-1` with `3.0.0-1` and so on?

radoering (commented):
Looks like a valid fix now. 😄 Just some nitpicks.

src/poetry/core/constraints/version/version_range.py
abn: Suggestions welcome. 

I just followed the names inclusive/exclusive ordering from pep440 really. But I guess in the context of ranges it can be ambiguous.

I could just drop the property altogether.

radoering: I would just drop it. (The naming is clear as long as there is only a lower limit or only an upper limit, but not if there are both.)

radoering (approved):

PR #382, opened 2022-05-29
don't merge python_version and python_full_version

dimbleby:
These markers are not quite equivalent; I've updated the relevant testcases to give examples where the previous code got it wrong.

cf https://github.com/python-poetry/poetry/issues/5717

I do not claim that this is the only code that confuses `python_version` and `python_full_version`, I just remember this particular example because I wrote it myself!

### DIFF
--- a/src/poetry/core/version/markers.py
+++ b/src/poetry/core/version/markers.py
@@ -47,8 +47,6 @@ ALIASES = {
     "python_implementation": "platform_python_implementation",
 }
 
-PYTHON_VERSION_MARKERS = ["python_version", "python_full_version"]
-
 # Parser: PEP 508 Environment Markers
 _parser = Parser(GRAMMAR_PEP_508_MARKERS, "lalr")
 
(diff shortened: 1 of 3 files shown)


neersighted (approved):
LGTM, but I'm curious how we got here -- did everyone just conflate the two of these together and not realize the subtle difference (that python_version is basically always .0, if I understand correctly)? Can we add some documentation somewhere that explicitly defines the intention for each of these (or link to the upstream docs)?

dimbleby:
> python_version is basically always .0 ...

that's not wrong exactly, but thinking that way is pretty much the misconception that I had when introducing the code that this MR is now removing.

`python_version x.y` and `python_full_version x.y.0` are not interchangeable.  Eg `python_version > 3.6` is more like `python_full_version >= 3.7.0` than it is like `python_full_version > 3.6.0`.

neersighted:
> > python_version is basically always .0 ...
> 
> that's not wrong exactly, but thinking that way is pretty much the misconception that I had when introducing the code that this MR is now removing.
> 
> `python_version x.y` and `python_full_version x.y.0` are not interchangeable. Eg `python_version > 3.6` is more like `python_full_version >= 3.7.0` than it is like `python_full_version > 3.6.0`.

See, that's definitely needed somewhere in this codebase... Any thoughts on a good place to document it or link to docs?

PR #385, opened 2022-06-01
fix and make common python version normalization

dimbleby:
relating to https://github.com/python-poetry/poetry/issues/5717

there were two similar but diverging bits of code.  They really ought to be the same; and the version that I've gone for is actually not quite the same as either of the originals.

Specifically, in the handling of python_version when specified with either one or two digits...

First, the single digit case.  
- I believe that a constraint like `python_version > 3` is strictly not valid, PEP 508 says that the `python_version` should be equivalent to `'.'.join(platform.python_version_tuple()[:2])`.  
- However experiment with `pip` and a `requirements.txt` teaches me that in my 3.8.10 environment, `python_version > 3` _does_ allow a dependency to be installed
- `python_version <= "3"` does not allow installation at 3.8.10
- So I conclude that the interpretation of `>3` as `> 3.0.0` per #155 is reasonable, and have removed code that treats `>3` as `>=4.0.0`

For the two digit case though, experiment shows that `python_version > "3.8"` does indeed cause pip not to install a dependency in my 3.8.10 environment, that I think _should_ be reinterpreted as `>=3.9.0`.  So I have included that code.

### DIFF
--- a/src/poetry/core/packages/dependency.py
+++ b/src/poetry/core/packages/dependency.py
@@ -16,6 +16,7 @@ from poetry.core.packages.constraints import (
 from poetry.core.packages.dependency_group import MAIN_GROUP
 from poetry.core.packages.specification import PackageSpecification
 from poetry.core.packages.utils.utils import contains_group_without_marker
+from poetry.core.packages.utils.utils import normalize_python_version_markers
 from poetry.core.semver.helpers import parse_constraint
 from poetry.core.semver.version_range_constraint import VersionRangeConstraint
 from poetry.core.version.markers import parse_marker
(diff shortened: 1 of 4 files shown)


neersighted (changes requested):
I think that being bug/implementation compatible with Pip here is, unfortunately, the correct move. That being said, I would like to put a streamlined version of your testing/research in the docstring or comments as appropriate. Looks good to merge after that.

dimbleby:
I added a comment.

I also went to the trouble of installing python 3.0.0, to see how single digit precision  really works at the edge cases.

Turns out that `python_version <= "3"` is _not_ satisfied, according to pip, at 3.0.0.  I reckon that probably what's happening is that the strings `"3"` (from the marker) and `"3.0"` (from the actual python version) are compared, as strings.

So I've added a comment saying that, and updated the code accordingly (not that there's anyone out there running 2.0.0 or 3.0.0 who will ever care about the difference)

neersighted:
> Hmm, not sure I appreciate sonarcloud telling me that it doesn't like this code, let's try a small rearrangement...

Honestly I find the SonarCloud-driven version to be more readable and cleaner :)

LGTM if suppressing Sonar is all we can do, despite the mild refactor.

PR #386, opened 2022-06-02
Typechecking tidying

dimbleby:
I was just going to tidy up some `Any` annotations, but in doing so I noticed that nearly all of the code in the `Link` object is unused.  So I removed it.

Did these in separate commits in case it's controversial

### DIFF
--- a/src/poetry/core/packages/constraints/constraint.py
+++ b/src/poetry/core/packages/constraints/constraint.py
@@ -2,8 +2,6 @@ from __future__ import annotations
 
 import operator
 
-from typing import Any
-
 from poetry.core.packages.constraints import AnyConstraint
 from poetry.core.packages.constraints.base_constraint import BaseConstraint
 from poetry.core.packages.constraints.empty_constraint import EmptyConstraint
@@ -128,7 +126,7 @@ class Constraint(BaseConstraint):
     def is_empty(self) -> bool:
         return False
 
-    def __eq__(self, other: Any) -> bool:
+    def __eq__(self, other: object) -> bool:
         if not isinstance(other, Constraint):
             return NotImplemented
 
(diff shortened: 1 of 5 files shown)


abn:
Metadata code should not be removed; this is used to support PEP 658.

see: https://github.com/python-poetry/poetry/pull/5509

dimbleby:
reinstated metadata

neersighted:
> reinstated metadata

https://github.com/python-poetry/poetry/pull/5509 also requires `requires_python` -- consensus here is that removing interfaces from core is a bad idea with no deprecation period unless they were ephemeral (e.g. only existed in alphas/betas before removal).

That being said, I think we can remove `comes_from` as we have no plans to use that field as best I can tell, and the types it was originally intended to be used with have been removed or heavily changed.

dimbleby:
> I'm curious if Poetry 1.1.x still uses this code and if a newer core may break stable Poetry

well I'd be all in favour of this project respecting semantic versioning: the next release should be 2.0.0.

But poetry 1.1.13 requires "~1.0.7", so it'll be ok.

Will put `requires_python` back too: `comes_from` was always the worst of these...

neersighted:
> > I'm curious if Poetry 1.1.x still uses this code and if a newer core may break stable Poetry
> 
> well I'd be all in favour of this project respecting semantic versioning: the next release should be 2.0.0.
> 
> But poetry 1.1.13 requires "~1.0.7", so it'll be ok.
> 
> Will put `requires_python` back too: `comes_from` was always the worst of these...

Don't worry, we're talking about that too and trying to come up with a better scheme for both upstream and end-users. I'd still love for you to join Discord and be part of these discussions, since we tend to flesh them out there before posting a design or implementation here 😄

neersighted (approved):

PR #388, opened 2022-06-02
Merge python_version and python_full_version markers

radoering:
Relates-to: python-poetry/poetry#5717

Reintroduces merging of `python_version` and `python_full_version`, which was removed in #382 because of python-poetry/poetry#5717

The `constraint` of the `python_marker` is converted via `get_python_constraint_from_marker()` and thus should be consistent with the latest findings described in #385.

### DIFF
--- a/src/poetry/core/version/markers.py
+++ b/src/poetry/core/version/markers.py
@@ -47,6 +47,8 @@ ALIASES = {
     "python_implementation": "platform_python_implementation",
 }
 
+PYTHON_VERSION_MARKERS = {"python_version", "python_full_version"}
+
 # Parser: PEP 508 Environment Markers
 _parser = Parser(GRAMMAR_PEP_508_MARKERS, "lalr")
 
@@ -414,11 +416,10 @@ class MultiMarker(BaseMarker):
                     for i, mark in enumerate(new_markers):
                         if isinstance(mark, SingleMarker) and (
                             mark.name == marker.name
+                            or {mark.name, marker.name} == PYTHON_VERSION_MARKERS
                         ):
                             new_marker = _merge_single_markers(mark, marker, cls)
                             if new_marker is not None:
-                                if new_marker.is_empty():
-                                    return new_marker
                                 new_markers[i] = new_marker
                                 intersected = True
 
@@ -617,11 +618,10 @@ class MarkerUnion(BaseMarker):
                     for i, mark in enumerate(new_markers):
                         if isinstance(mark, SingleMarker) and (
                             mark.name == marker.name
+                            or {mark.name, marker.name} == PYTHON_VERSION_MARKERS
                         ):
                             new_marker = _merge_single_markers(mark, marker, cls)
                             if new_marker is not None:
-                                if new_marker.is_any():
-                                    return new_marker
                                 new_markers[i] = new_marker
                                 included = True
                                 break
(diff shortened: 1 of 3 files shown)


neersighted (changes requested):
Given how brittle this has been I'm a little nervous about your change as I'm not quite familiar enough with the subtleties here -- however the set of test cases look quite robust and I can easily reason about that part of the change.

@dimbleby could you PTAL?

src/poetry/core/version/markers.py
```diff
@@ -869,3 +850,72 @@ def dnf(marker: BaseMarker) -> BaseMarker:
...
+        # We have to fix markers like 'python_full_version == "3.6"'
+        # to receive 'python_full_version == "3.6.0"'.
+        # It seems a bit hacky to convert to string and back to marker,
+        # but it's probably much simpler than to consider the different constraint
+        # classes (mostly VersonRangeConstraint, but VersionUnion for "!=") and
+        # since this conversion is only required for python_full_version markers
+        # it may be sufficient to handle it here.
+        marker_string = str(merged_marker)
+        precision = marker_string.count(".") + 1
+        if precision != 3:
```
dimbleby, line 918: I suppose the parser has already assured us that there can never be more than two dots: still `if precision < 3` feels both safer and more exactly expresses what we mean

src/poetry/core/version/markers.py
```diff
@@ -869,3 +850,72 @@ def dnf(marker: BaseMarker) -> BaseMarker:
...
+        marker = marker2
+        full_marker = marker1
+
+    normalized_constraint = get_python_constraint_from_marker(marker)
+    normalized_marker = SingleMarker("python_full_version", normalized_constraint)
+    merged_marker = _merge_single_markers(normalized_marker, full_marker, merge_class)
+    if merged_marker == normalized_marker:
+        # prefer original marker to avoid unnecessary changes
+        return marker
+    elif merged_marker and isinstance(merged_marker, SingleMarker):
```
dimbleby, line 908: no need to `elif` since we returned from the `if` branch: just `if` will do

neersighted:
> I think this looks correct, I'll leave a couple of style quibbles.
> 
> I continue to have a slight feeling of unease that this code has become so elaborate: I like a marker simplification as much as the next person, but there's so much of it!

I don't necessarily disagree, but I think ultimately we get higher quality results if we try to make sure the markers are as consistent and clean as possible. I think our best tool for maintainability is robust and through testing, and incremental refactoring as better patterns make themselves apparent. That being said, yes there's getting to be a lot of it and if I don't pay attention for a few weeks I forget how it works again 😆

src/poetry/core/version/markers.py
```diff
@@ -630,28 +623,16 @@ def of(cls, *markers: BaseMarker) -> BaseMarker:
...
-                                isinstance(constraint_union, VersionConstraint)
-                                and constraint_union.is_simple()
-                            ):
-                                new_markers[i] = SingleMarker(
-                                    mark.name, constraint_union
-                                )
+                        ):
+                            new_marker = _merge_single_markers(mark, marker, cls)
+                            if new_marker is not None:
+                                if new_marker.is_any():
```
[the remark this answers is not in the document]
radoering: It can save some loop cycles and function calls, but in the big picture, I think marker simplifications don't add up to a significant part of the overall time. So let's remove it for now until someone does some real world measurements.

radoering:
> I continue to have a slight feeling of unease that this code has become so elaborate: I like a marker simplification as much as the next person, but there's so much of it!

It's just necessary for the solver to not try solving some bogus overrides. Solving not just for one environment is complex, but that's one of poetry's USPs.

neersighted (approved):

PR #390, opened 2022-06-03
union of adjacent ranges should collapse

dimbleby:
Picking off some more cases in marker simplification.


dimbleby:
NB some of these testcases - and also some existing ones - should really use `python_full_version` when testing 3-digit precision.  #388 would allow that.

radoering (approved):

PR #394, opened 2022-06-06
Follow most recent packaging specification for wheel name normalization

mkniewallner:
Highlighted by https://github.com/python-poetry/poetry/issues/5782 and looking into PyPA [packaging specification](https://packaging.python.org/en/latest/specifications/binary-distribution-format/#escaping-and-unicode) that supersedes [PEP 427](https://peps.python.org/pep-0427/):
> In distribution names, any run of `-_.` characters (HYPHEN-MINUS, LOW LINE and FULL STOP) should be replaced with `_` (LOW LINE), and uppercase characters should be replaced with corresponding lowercase ones. This is equivalent to [PEP 503](https://www.python.org/dev/peps/pep-0503) normalisation followed by replacing `-` with `_`. Tools consuming wheels must be prepared to accept `.` (FULL STOP) and uppercase letters, however, as these were allowed by an earlier version of this specification.

So it looks like the regex we use to escape wheel name should be updated to what [PEP 503 defines](https://peps.python.org/pep-0503/#normalized-names), with the minor change of using `_` for the replacement character instead of `-`.

### DIFF
--- a/src/poetry/core/masonry/utils/helpers.py
+++ b/src/poetry/core/masonry/utils/helpers.py
@@ -29,5 +29,8 @@ def escape_version(version: str) -> str:
 
 
 def escape_name(name: str) -> str:
-    """Escaped wheel name as specified in https://packaging.python.org/en/latest/specifications/binary-distribution-format/#escaping-and-unicode."""
+    """
+    Escaped wheel name as specified in https://packaging.python.org/en/latest/specifications/binary-distribution-format/#escaping-and-unicode.
+    This method should only be used for the generation of wheels, and not for the artifact package names.
+    """
     return re.sub(r"[-_.]+", "_", name, flags=re.UNICODE).lower()
--- a/src/poetry/core/masonry/utils/helpers.py
+++ b/src/poetry/core/masonry/utils/helpers.py
@@ -29,5 +29,5 @@ def escape_version(version: str) -> str:
 
 
 def escape_name(name: str) -> str:
-    """Escaped wheel name as specified in :pep:`427#escaping-and-unicode`."""
-    return re.sub(r"[^\w\d.]+", "_", name, flags=re.UNICODE)
+    """Escaped wheel name as specified in https://packaging.python.org/en/latest/specifications/binary-distribution-format/#escaping-and-unicode."""
+    return re.sub(r"[-_.]+", "_", name, flags=re.UNICODE).lower()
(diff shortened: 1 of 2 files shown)


neersighted (dismissed):
This looks good to me as this only affects generated artifacts (bdists) and the spec explicitly calls out that tools must continue to support wheel names generated using the old logic -- given this is a subset of possible names before the spec update, it should be fully compatible. Still, going to wait for @abn to concur.

abn:
As this is only affecting generated bdist files, I agree with @neersighted.

However (outside the scope of this PR), I am now curious if we do the right thing in cases where PEP 503 normalisation does not match wheel name escapes - ie. how do we tell if `foo-bar` in a package source is the same as `foo_bar.whl` direct reference dependency?

mkniewallner:
> As this is only affecting generated bdist files, I agree with @neersighted.

Thanks to both of you for confirming that. I also think that since this only touches the generation of the wheel name (not it's retrieval locally nor on a distant remote, AFAIK), we should be safe here.

> However (outside the scope of this PR), I am now curious if we do the right thing in cases where PEP 503 normalisation does not match wheel name escapes - ie. how do we tell if `foo-bar` in a package source is the same as `foo_bar.whl` direct reference dependency?

That's a valid concern. I can take a look at what we currently do later on, since the normalisation of wheel names may occur in various different situations (local cache, distant remote, etc.), so definitely worth a look.

src/poetry/core/masonry/utils/helpers.py
```diff
@@ -29,5 +29,5 @@ def escape_version(version: str) -> str:
 
 
 def escape_name(name: str) -> str:
-    """Escaped wheel name as specified in :pep:`427#escaping-and-unicode`."""
-    return re.sub(r"[^\w\d.]+", "_", name, flags=re.UNICODE)
+    """Escaped wheel name as specified in https://packaging.python.org/en/latest/specifications/binary-distribution-format/#escaping-and-unicode."""
```
abn, line 32: Maybe expand this doc string to indicate that this should only be used for generation of wheels never to canonicalize artifact package names.

src/poetry/core/masonry/utils/helpers.py
```diff
@@ -29,5 +29,8 @@ def escape_version(version: str) -> str:
 
 
 def escape_name(name: str) -> str:
-    """Escaped wheel name as specified in :pep:`427#escaping-and-unicode`."""
-    return re.sub(r"[^\w\d.]+", "_", name, flags=re.UNICODE)
+    """
+    Escaped wheel name as specified in https://packaging.python.org/en/latest/specifications/binary-distribution-format/#escaping-and-unicode.
+    This method should only be used for the generation of wheels, and not for the artifact package names.
```
neersighted, line 34:
```suggestion
    This function should only be used for the generation of artifact names, and not to normalize or filter existing artifact names.
```

neersighted (approved):

PR #395, opened 2022-06-07
make intersection of two Versions symmetrical

dimbleby:
Fixes https://github.com/python-poetry/poetry/issues/5791


radoering (approved):
`intersect` should definitively be symmetrical 👍

PR #396, opened 2022-06-07
version: fix allows for local and post releases according to PEP 440

radoering:
Resolves: python-poetry/poetry#4729

Partially overlaps with #379 but only considers simple version (not ranges).

According to PEP 440:

> For example, given the version 1.1.post1, the following clauses would match or not as shown:
> 
> == 1.1        # Not equal, so 1.1.post1 does not match clause
> == 1.1.post1  # Equal, so 1.1.post1 matches clause
> == 1.1.*      # Same prefix, so 1.1.post1 matches clause

and

> If the specified version identifier is a public version identifier (no local version label), then the local version label of any candidate versions MUST be ignored when matching versions.
> 
> If the specified version identifier is a local version identifier, then the local version labels of candidate versions MUST be considered when matching versions, with the public version identifier being matched as described above, and the local version label being checked for equivalence using a strict string equality comparison.

### DIFF
--- a/src/poetry/core/semver/version.py
+++ b/src/poetry/core/semver/version.py
@@ -86,12 +86,6 @@ class Version(PEP440Version, VersionRangeConstraint):
         # allow weak equality to allow `3.0.0+local.1` for `3.0.0`
         if not _this.is_local() and _other.is_local():
             _other = _other.without_local()
-        elif _this.is_local() and not _other.is_local():
-            _this = _this.without_local()
-
-        # allow weak equality to allow `3.0.0-1` for `3.0.0`
-        if not _this.is_postrelease() and _other.is_postrelease():
-            _other = _other.without_postrelease()
 
         return _this == _other
 
--- a/tests/semver/test_version.py
+++ b/tests/semver/test_version.py
@@ -122,25 +122,30 @@ def test_allows() -> None:
     assert not v.allows(Version.parse("1.3.3"))
     assert not v.allows(Version.parse("1.2.4"))
     assert not v.allows(Version.parse("1.2.3-dev"))
+    assert not v.allows(Version.parse("1.2.3-1"))
+    assert not v.allows(Version.parse("1.2.3-1+build"))
     assert v.allows(Version.parse("1.2.3+build"))
-    assert v.allows(Version.parse("1.2.3-1"))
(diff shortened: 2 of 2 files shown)


src/poetry/core/semver/version.py
```diff
@@ -86,12 +86,6 @@ def allows(self, version: Version | None) -> bool:
         # allow weak equality to allow `3.0.0+local.1` for `3.0.0`
         if not _this.is_local() and _other.is_local():
             _other = _other.without_local()
-        elif _this.is_local() and not _other.is_local():
-            _this = _this.without_local()
-
-        # allow weak equality to allow `3.0.0-1` for `3.0.0`
-        if not _this.is_postrelease() and _other.is_postrelease():
```
abn, line 93: Doesn't the pep only talk about when `_this.is_postrelease()`?

radoering: I interpret the example

> == 1.1 # Not equal, so 1.1.post1 does not match clause

from pep as follows:

`_this` is 1.1 (no post release)
`_other` is 1.1.post1 (post release)

`_this` should not allow `_other`

abn: > For example, given the version 1.1.post1, the following clauses would match or not as shown

Seems to imply `_this` is `1.1.post1` and `_other` is `1.1`.

radoering: I disagree. In my understanding, the "given version" is the version to be allowed or not (`_other`). The "clause" can be any version constraint (exact version, version range, version union). The clause allows or doesn't allow a given version (clause = `_this`, given version = `_other`).

By the way, considering the third example, it wouldn't make sense to check if `1.1.post1` allows `1.1.*`. Thus, it has to be the other way around, hasn't it?

dimbleby: In this particular case it doesn't much matter which reading is right, because the other is also given ten lines later.  ie we have both

> == 1.1        # Not equal, so 1.1.post1 does not match clause

and 

> == 1.1.post1  # Not equal (post-release), so 1.1 does not match clause

abn: I stand corrected.

abn (approved):

PR #399, opened 2022-06-12
versions: preserve any local label when bumping

abn:
Split out of #379

### DIFF
--- a/src/poetry/core/version/pep440/version.py
+++ b/src/poetry/core/version/pep440/version.py
@@ -205,7 +205,7 @@ class PEP440Version:
         release = self.release
         if self.is_stable() or Release(self.release.major, 0, 0) < self.release:
             release = self.release.next_major()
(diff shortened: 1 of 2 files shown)


dimbleby:
What is this fixing?

I don't think it's obvious what the "right" behaviour is, or even that there is a single right behaviour

- eg users of `poetry version` might - or might not - want this updated behaviour
- there are internal places that call `next_minor()` and `next_major()` as part of version comparison, does this change help or hurt them?

radoering:
Further, according to PEP 440 local version identifiers are not permitted in [inclusive](https://peps.python.org/pep-0440/#inclusive-ordered-comparison) and [exclusive](https://peps.python.org/pep-0440/#exclusive-ordered-comparison) ordered comparisons. Thus, the tested constraints are invalid according to PEP 440. If this behavior is required for some reason, we should at least add a comment explaining why it's required.

PR #402, opened 2022-06-15
pypa packaging compliance

abn:
Resolves: python-poetry/poetry#7050

@radoering @neersighted stashing pandora's box here; thoughts welcome


tests/semver/test_version_range.py
```diff
@@ -446,15 +447,207 @@ def test_union(
...
+            ("2!1.0", "~=1.0"),
+            ("2!1.0", "==1.0"),
+            ("1.0", "==2!1.0"),
+            ("2!1.0", "==1.*"),
+            ("1.0", "==2!1.*"),
+            ("2!1.0", "!=2!1.0"),
+        ]
+    ],
+)
+def test_specifiers(version: str, spec: str, expected: bool) -> None:
```
abn, line 633: This case was copied over from https://github.com/pypa/packaging/blob/8b86d85797b9f26d98ecfbe0271ce4dc9495d98c/tests/test_specifiers.py#L469

radoering: Do we to add a comment in the code to keep track?

src/poetry/core/semver/helpers.py
```diff
@@ -6,10 +6,21 @@
...
 def parse_constraint(constraints: str) -> VersionConstraint:
+    return _parse_constraint(constraints=constraints)
+
+
+def parse_marker_version_constraint(constraints: str) -> VersionConstraint:
+    return _parse_constraint(constraints=constraints, is_marker_constraint=True)
+
+
+def _parse_constraint(
+    constraints: str, is_marker_constraint: bool = False
```
radoering, line 22:
```suggestion
    constraints: str, *, is_marker_constraint: bool = False
```

src/poetry/core/semver/helpers.py
```diff
@@ -49,7 +64,9 @@ def parse_constraint(constraints: str) -> VersionConstraint:
         return VersionUnion.of(*or_groups)
 
 
-def parse_single_constraint(constraint: str) -> VersionConstraint:
+def parse_single_constraint(
+    constraint: str, is_marker_constraint: bool = False
```
radoering, line 68:
```suggestion
    constraint: str, *, is_marker_constraint: bool = False
```

src/poetry/core/semver/helpers.py
```diff
@@ -138,10 +144,55 @@ def parse_single_constraint(constraint: str) -> VersionConstraint:
...
+
         return version
 
     from poetry.core.semver.exceptions import ParseConstraintError
 
     raise ParseConstraintError(f"Could not parse version constraint: {constraint}")
+
+
+def _make_x_constraint_range(
+    version: Version, invert: bool = False, is_marker_constraint: bool = False
```
radoering, line 166:
```suggestion
    version: Version, *, invert: bool = False, is_marker_constraint: bool = False
```

src/poetry/core/semver/helpers.py
```diff
@@ -138,10 +144,55 @@ def parse_single_constraint(constraint: str) -> VersionConstraint:
...
+        _is_zero_major = False
+    elif version.is_stable():
+        _next = version.next_stable()
+    elif version.is_prerelease():
+        _next = version.next_prerelease()
+        _is_zero_major = False
+    elif version.is_devrelease():
+        _next = version.next_devrelease()
+        _is_zero_major = False
+
```
radoering, line 183: We should raise some error in the else, shouldn't we?

src/poetry/core/semver/helpers.py
```diff
@@ -138,10 +144,55 @@ def parse_single_constraint(constraint: str) -> VersionConstraint:
...
+    if _is_zero_major:
+        result = VersionRange(max=_next.with_precision(3))
+    else:
+        _min = version.with_precision(max(version.precision, 3))
+
+        if not is_marker_constraint and not _next.is_unstable():
+            _min = _min.next_devrelease()
```
radoering, line 190: Actually, our implementation of `next_devrelease()` is kind of confusing. The next dev release of `1.2.3` can't be `1.2.3dev0` because `1.2.3dev0` was before `1.2.3`. `first_devrelease()` might be a more suitable name.

radoering (changes requested):
Some first comments. (I did not review all of it yet.) The changed and new tests look quite good.

Can we drop the first commit here please, because I think that's a separate topic and doesn't seem to influence the other changes?

src/poetry/core/version/pep440/segments.py
```diff
@@ -102,6 +103,53 @@ def next_patch(self) -> Release:
             extra=tuple(0 for _ in self.extra),
         )
 
+    def next(self) -> Release:
```
radoering, line 106: Same here, I'd like some explicit tests. Seems like tests for Release are in `test_version_pep440.py`.

src/poetry/core/semver/version_range_constraint.py
```diff
@@ -36,44 +31,78 @@ def include_min(self) -> bool:
...
+            return None
+
+        if self.include_max or self.max.is_unstable():
+            return self.max
+
+        if self.min == self.max and (self.include_min or self.include_max):
+            # this is an equality range
+            return self.max
+
+        return self.max.next_devrelease()
```
radoering, line 60: As already mentioned before in another place, this is absolutely confusing. Should really be `first_devrelease()` or similar.

src/poetry/core/semver/version_range_constraint.py
```diff
@@ -36,44 +31,78 @@ def include_min(self) -> bool:
...
         if self.min is None:
-            return other.min is not None
+            return None
+
+        if not self.include_min or self.min.is_unstable():
+            return self.min
+
+        if self.min == self.max and (self.include_min or self.include_max):
+            # this is an equality range
+            return self.min
```
radoering, line 44:
```suggestion
```

Pointless since we return `self.min` in any case.

src/poetry/core/semver/version_range.py
```diff
@@ -69,27 +52,35 @@ def is_simple(self) -> bool:
 
     def allows(self, other: Version) -> bool:
         if self._min is not None:
-            if other < self._min:
+            _this, _other = self.allowed_min, other
+
+            assert _this is not None
+
+            if not _this.is_postrelease() and _other.is_postrelease():
+                _other = _other.without_postrelease()
```
radoering, line 60: Seems unintuitive at first glance. Maybe, we should add a comment like `# required for "2.0.post1" does not match ">2"`

src/poetry/core/semver/version_range.py
```diff
@@ -69,27 +52,35 @@ def is_simple(self) -> bool:
...
+
+            if not _this.is_local() and _other.is_local():
+                _other = other.without_local()
```
radoering, line 63: Same here: `# required for "2.0+local.version" does not match ">2"`

radoering (changes requested):
Next part of review. Some more todos in terms of code qualitiy, but functionally a good piece towards full PEP 440 compliance. 👍

radoering:
Rebased on current main and dropped first commit because it was controversial, not related to pypa packaging compliance and tracked in #399 anyway.

radoering (changes requested):
There might be an issue with pre-release-handling:

Without this PR, something like `"==3.8.*"` is parsed to `">=3.8,<3.9"`.
With this PR, it is parsed to `">=3.8.dev0,<3.9"`.

Although considered on its own, the latter is more correct, it implicitly sets `allow_prereleases`: https://github.com/python-poetry/poetry-core/blob/0759db61f01f008edc3ffe2b790ac9ecc892da77/src/poetry/core/packages/dependency.py#L79-L81

That's probably not desired behavior.

Update: fixed via #543

PR #403, opened 2022-06-18
Fix for tomlkit regression resulting in inconsistent line endings

radoering:
Related to python-poetry/poetry#5675

Only required until a new tomlkit version with sdispater/tomlkit#201 (or another fix for sdispater/tomlkit#200) is available.


radoering:
superseded by #412

PR #40, opened 2020-06-29
Added script file feature

peterdeme:
Resolves: [python-poetry#241](https://github.com/python-poetry/poetry/issues/241)

- [x] Added **tests** for changed code.
- [x] Updated **documentation** for changed code.

# Feature: ability to add script files to distribution packages

This one is a long overdue, see: https://github.com/sdispater/poetry/issues/241 (June 2018)

The implementation is based on the official specification by @abn https://github.com/python-poetry/poetry/issues/2310

This is a very basic feature of setup.py See here: [docs](https://python-packaging.readthedocs.io/en/latest/command-line-scripts.html#the-scripts-keyword-argument).

Note: this is **not the same as** `entry_point`! This setuptools feature basically just copies over the given files to the virtualenv's `bin` folder so that it'll be available on the `$PATH` and you can simply invoke it.

### Example

```toml
[tools.poetry]
package = "my_package"

[tools.poetry.scripts]
migration = { source = "bin/run_db_migrations.sh", type = "file" }
setup_script = { source = "bin/run_setup_script.sh" } # type can be omitted!
```

Then:

```bash
poetry install
poetry run sh run_db_migrations.sh
poetry run sh run_setup_script.sh
```

### Testing

- I added a couple of automated tests
- I tested this **manually** with one of our repositories and it worked

PR #404, opened 2022-06-24
fix(factory): safe get `extras` in scripts and config

mkniewallner:
Resolves: python-poetry/poetry#4665

According to:
https://github.com/python-poetry/poetry-core/blob/3bf7ad0f1dc2b1a5cabe4bb58cf7e1c6d832870d/src/poetry/core/json/schemas/poetry-schema.json#L6-L10
 https://github.com/python-poetry/poetry-core/blob/3bf7ad0f1dc2b1a5cabe4bb58cf7e1c6d832870d/src/poetry/core/json/schemas/poetry-schema.json#L597-L600

- `extras` is not required in the main config (under `poetry.tools.extras`)
- `extras` is not required when defining a script (under `poetry.tools.scripts`) using the syntax introduced in https://github.com/python-poetry/poetry-core/pull/40

Despite that, `Factory.validate` assumes that `extras` key is always present for both cases by hard accessing it, making the script definition fail for 2 possible reasons:
- `poetry.tool.extras` is not defined at all
- a script defined as a dict doesn't define an `extras` key

This PR fixes those 2 use cases by soft accessing the key and defaulting to an empty list/dict.
It also adds tests to validate the errors/warnings that could happen when using the `strict` flag, since it was lacking some coverage.

### DIFF
--- a/src/poetry/core/factory.py
+++ b/src/poetry/core/factory.py
@@ -407,13 +407,15 @@ class Factory:
             # Checking for scripts with extras
             if "scripts" in config:
                 scripts = config["scripts"]
+                config_extras = config.get("extras", {})
+
                 for name, script in scripts.items():
                     if not isinstance(script, dict):
                         continue
 
-                    extras = script["extras"]
+                    extras = script.get("extras", [])
                     for extra in extras:
-                        if extra not in config["extras"]:
+                        if extra not in config_extras:
                             result["errors"].append(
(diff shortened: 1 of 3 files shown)


TBBle (dismissed):

mkniewallner:
Rebased from `main` to pick up https://github.com/python-poetry/poetry-core/pull/406 so that tests can pass.

neersighted (approved):

PR #405, opened 2022-06-24
Do not consider constraints when checking for equality of direct origi…

Jinior:
…n dependencies

Resolves: python-poetry/poetry#5902 and partially python-poetry/poetry#2415

- [n.a.] Added **tests** for changed code.
- [n.a.] Updated **documentation** for changed code.

An alternative solution to python-poetry/poetry#5902 as proposed in python-poetry/poetry#5903.

Currently poetry re-downloads direct origin packages multiple times during locking because the deferred cache used to store downloaded packages is indexed using a dependency object with an explicit constraint while the lookup is done with a dependency object with an unbounded constraint which do not test equal.

This PR solves that issue by changing the `__eq__` equality check for direct origin dependencies to not consider the constraint. This should pose no problems since direct origin dependencies pointing to the same origin implicitly have the same constraint.

Should I add a test for this or do we consider this an implementation detail for the sake of our current caching mechanism?

### DIFF
--- a/src/poetry/core/packages/dependency.py
+++ b/src/poetry/core/packages/dependency.py
@@ -574,7 +574,9 @@ class Dependency(PackageSpecification):
         if not isinstance(other, Dependency):
             return NotImplemented
 
-        return super().__eq__(other) and self._constraint == other.constraint
+        return super().__eq__(other) and (
+            self._constraint == other.constraint or self.is_direct_origin()
+        )
 
     def __hash__(self) -> int:
         # don't include _constraint in hash because it is mutable!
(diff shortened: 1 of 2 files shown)


Jinior:
Can somebody re-run the pipeline on this PR? I suspect it failed due to #406

dimbleby:
@Jinior simply re-running the tests won't help, because your branch doesn't have the fix for #406 in it.

You should rebase on the latest `main` code which will pick up the fix; and then push which will rerun the tests

radoering (approved):

PR #406, opened 2022-06-25
tests: xfail test_pip_install_no_binary on Python 3.7

radoering:
The test case fails due to the new release of `setuptools_scm`, which now depends on `importlib-metadata` for `python_version < "3.8"`. Due to the fact that importlib-metadata itself depends on setuptools_scm, it's not possible anymore to `pip install --no-binary importlib-metadata` on Python 3.7.

Related to python/importlib_metadata#392

### DIFF
--- a/tests/integration/test_pep517.py
+++ b/tests/integration/test_pep517.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+import sys
+
 from pathlib import Path
 from typing import TYPE_CHECKING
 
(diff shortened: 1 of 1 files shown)


mkniewallner (approved):

tests/integration/test_pep517.py
```diff
@@ -75,6 +77,11 @@ def test_pip_wheel_build(temporary_directory: Path, project_source_root: Path) -
     assert len(wheels) == 1
 
 
+@pytest.mark.xfail(
+    tuple(int(part) for part in platform.python_version_tuple()[:2]) < (3, 8),
```
mkniewallner, line 81: What about using `sys.version_info < (3, 8)` like in [pytest documentation](https://docs.pytest.org/en/stable/how-to/skipping.html#id1)?

radoering: I knew there was a simpler way to do that but couldn't remember. 👍

mkniewallner (approved):

PR #409, opened 2022-07-01
Partially handle extra version for versions constraint

### LINKED ISSUE #3186: "Compatible Release" (~=) operator violates PEP 440 with long (4-part+) versions
- [X] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [X] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [X] **(N/A)** If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

<details>
<summary>poetry debug output</summary>

```
Poetry
Version: 1.1.2
Python:  3.6.11

Virtualenv
Python:         3.6.11
Implementation: CPython
Path:           /tmp/tmp.1eGGnOnKny/.venv
Valid:          True

System
Platform: linux
OS:       posix
Python:   /home/tobyh/.pyenv/versions/3.6.11
```

</details>

- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://gist.github.com/tharradine/9e5b7ea9c165a749e581697ee92fc54f (poetry.lock also included)

## Issue
It seems as though when specifying a dependency with a 4-part (or possibly higher) version number, that the `~=` operator violates PEP 440. An example is included in my pyproject.toml and poetry.lock.

Specifically, when locking:

Input:

- `anchor-exp = "~=0.0.0.5"`

Expected locked version:

- `anchor-exp = "==0.0.0.6"`

Actual locked version:

- `anchor-exp = "==0.0.2.0"`

This does not match the example shown in [PEP 440](https://www.python.org/dev/peps/pep-0440/#compatible-release):

>  The padding rules for release segment comparisons means that the assumed degree of forward compatibility in a compatible release clause can be controlled by appending additional zeros to the version specifier:
```
~= 2.2.0
>= 2.2.0, == 2.2.*

~= 1.4.5.0
>= 1.4.5.0, == 1.4.5.*
```

### Current workaround
~~Specify the correctly resolved version constraints manually (e.g. `anchor-exp~=0.0.0.5` -> `anchor-exp>=0.0.0.5,==0.0.0.*`). This works fine for primary dependencies, but may be problematic with transitive dependencies.~~ *Edit: It seems as though doing this results in the same version, which tells me the issue could actually be in the version-matching with `*`. Actual workaround can be `anchor-exp~=0.0.0.5` -> `anchor-exp>=0.0.0.5,<0.0.1`*

mkniewallner:
Resolves: python-poetry/poetry#3186

Poetry seems to assume that all packages follow SemVer specification for versioning. However, some packages do not, as python-poetry/poetry#3186 highlights with [anchor-exp](https://pypi.org/project/anchor-exp/).

Per [PEP 440 specification](https://peps.python.org/pep-0440/#compatible-release), when packages define more than 3 parts for the versioning, using `~= 1.4.5.0` is equivalent to `>= 1.4.5.0, == 1.4.5.*`.

Poetry doesn't handle that correctly today, since defining `~= 1.4.5.0` means that we accept `>= 1.4.5.0, == 1.4.*.*`.

This PR is an attempt at solving this case for both `~=` and wildcard operator. It is in draft for multiple reasons:
- It only handles one more level, which is useful for dependencies using a 4-parts versioning, but it still doesn't handle versioning using more than 4 parts (which I don't think is common, but this could still happen in theory).
- it only handles `~=` and wildcard operators, but what about other ones? Per the documentation, `^` is only for SemVer, but this is less clear for `~`, though given how it works, it also doesn't seem to be a good fit with something else than SemVer.
- Finally, since this would be a breaking change in the constraints handling, I'm not sure we want to introduce this change.

Side note: python-poetry/poetry#3186 also highlights the fact that `~=` is undocumented in Poetry. Is there a particular reason to not document it, or would a PR documenting it would be accepted?


radoering:
superseded by #973

PR #411, opened 2022-07-05
Allow : in Author name.

Amourspirit:
Suport for names such as :Barry-Thomas-Paul: Moss
See: [FOR THE CLAIM OF THE LIFE.](https://www.fortheclaimofthelife.com/)

Resolves: python-poetry#


neersighted:
I'm really unclear on the purpose or value of this (and the linked site seems to be a combo of a sovereign citizen/world citizen/law of attraction sort of thing) -- but setuptools allows it, so I suppose we might as well.

It looks like we don't have any test coverage here -- could you please write a test with good and bad names?

neersighted (changes requested):
Requires tests.

Amourspirit:
> I'm really unclear on the purpose or value of this (and the linked site seems to be a combo of a sovereign citizen/world citizen/law of attraction sort of thing) -- but setuptools allows it, so I suppose we might as well.
> 
> It looks like we don't have any test coverage here -- could you please write a test with good and bad names?

Short version Poetry is blocking my Lawful name.

Amourspirit:
> Requires tests.

All current test are passing. I will add specific test and update PR.

neersighted (approved):

PR #418, opened 2022-07-17
use packaging.utils.canonicalize_name()

dimbleby:
see https://github.com/python-poetry/poetry/pull/6022


dimbleby:
Hmm, the "poetry downstream" tests should surely have been broken by this - it removes a method that is used by poetry.

edgarrmondragon:
@dimbleby I guess `canonicalize_name` is still importable from the core's module because it has the same name.

neersighted (approved):

PR #442, opened 2022-08-14
Improve type hints for Package.name etc.

radoering:
When comparing names it's important to know if a name of a package/dependency is normalized or not. So let's tighten up type hints.

### DIFF
--- a/src/poetry/core/masonry/metadata.py
+++ b/src/poetry/core/masonry/metadata.py
@@ -6,13 +6,15 @@ from poetry.core.utils.helpers import readme_content_type
 
 
 if TYPE_CHECKING:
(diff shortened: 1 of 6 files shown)


dimbleby:
Also here https://github.com/python-poetry/poetry-core/blob/971f82488bfc082fb0065d04eaa6c6cf355ad3d3/src/poetry/core/masonry/metadata.py#L15?

Also I think this MR show that the calls to `canonicalize_name` here are redundant https://github.com/python-poetry/poetry-core/blob/971f82488bfc082fb0065d04eaa6c6cf355ad3d3/src/poetry/core/pyproject/tables.py#L46-L53

mkniewallner (approved):

PR #445, opened 2022-08-20
[WIP] fix(factory): unwrap config to not end up with `tomlkit` instances

mkniewallner:
Resolves python-poetry/poetry/issues/6201.

WIP for now, since I need to measure the impacts of this change, and this is missing tests.


Secrus:
Closing, since #483 was merged and we no longer use tomlkit in poetry-core

PR #446, opened 2022-08-24
deprecate `Package.python_marker`

dimbleby:
remove with_python_versions() on the grounds that

-  it has had no callers since more than two years now https://github.com/python-poetry/poetry/commit/2433e0e7007edddf6c74635aae5fc4b3dfdcc617
- the naming convention `with_` (or `without_`) on the package is nowadays being used to mean "a clone of the package with (or without) this thing", this collides with and confuses that
- it's bizarre anyway

### DIFF
--- a/src/poetry/core/packages/package.py
+++ b/src/poetry/core/packages/package.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import re
+import warnings
 
 from typing import TYPE_CHECKING
 from typing import ClassVar
@@ -272,6 +273,21 @@ class Package(PackageSpecification):
     @property
     def python_constraint(self) -> VersionConstraint:
         return self._python_constraint
+
+    @property
+    def python_marker(self) -> BaseMarker:
+        from poetry.core.packages.utils.utils import create_nested_marker
+        from poetry.core.version.markers import parse_marker
+
+        warnings.warn(
+            "`python_marker` is deprecated and will be removed in a future release.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+
+        return parse_marker(
+            create_nested_marker("python_version", self._python_constraint)
+        )
 
     @property
     def license(self) -> License | None:
--- a/src/poetry/core/packages/package.py
+++ b/src/poetry/core/packages/package.py
@@ -12,9 +12,7 @@ from poetry.core.constraints.version import parse_constraint
 from poetry.core.constraints.version.exceptions import ParseConstraintError
 from poetry.core.packages.dependency_group import MAIN_GROUP
 from poetry.core.packages.specification import PackageSpecification
-from poetry.core.packages.utils.utils import create_nested_marker
 from poetry.core.version.exceptions import InvalidVersionError
-from poetry.core.version.markers import parse_marker
 
 
 if TYPE_CHECKING:
@@ -113,7 +111,6 @@ class Package(PackageSpecification):
 
         self._python_versions = "*"
         self._python_constraint = parse_constraint("*")
-        self._python_marker: BaseMarker = AnyMarker()
 
         self.marker: BaseMarker = AnyMarker()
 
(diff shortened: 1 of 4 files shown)


radoering (commented):
I'm a little torn whether we should deprecate it before removing because there might be some plugin out there using it.

dimbleby:
There's lots of crud in poetry-core that it's desirable to remove (I have similar queued up that I'll submit when things quieten down / I have a large enough collection).  Going through deprecation periods for everything is going to get tedious.

IMO: clean up at will, note it in the changelog, do major version bumps if we think we've made possibly-breaking changes.  I see nothing to be afraid of in this package going 2.0 or 10.0 or even 50.0.

dimbleby:
Added a couple more removals to this MR, probably might as well discuss / make a decision for them all in the same place

- `EmptyConstraint.matches()` should be uncontroversial, who knows what this was intended for but anyone who was relying on it can replace with `True`
- `python_marker` on packages may be more interesting.  
  - If any plugin was relying on this they can always reconstruct it using the removed code `parse_marker(create_nested_marker(...))`
  - from the point of view of poetry / poetry-core this is a great candidate for removal: that's quite a lot of work to do for every package, in code that's had more than it's share of bugs, so if it's not needed then that's super news

dimbleby:
Actually the export plugin is an example of a project that relies on the python_marker, but only for the root package.

I'm pretty neutral about whether to leave it on the root package (reasoning about cost can't matter for a single package), or have the export plugin work it out itself from the root python_constraint.  Or we could just keep python_marker` everywhere.  Views welcome.

radoering:
> I'm pretty neutral about whether to leave it on the root package (reasoning about cost can't matter for a single package), or have the export plugin work it out itself from the root python_constraint. Or we could just keep `python_marker` everywhere. Views welcome.

We could make it a `cached_property`. Unfortunately, `cached_property` is not available for Python 3.7. We [added a backport package in poetry](https://github.com/python-poetry/poetry/blob/33b8b5a480de739e29938503209e404e0fa2a2a8/pyproject.toml#L52). We probably had to vendor it here (or just copy https://github.com/penguinolog/backports.cached_property/blob/1.0.2/backports/cached_property/__init__.py into `utils/_compat.py`)?

dimbleby:
Unless profiling shows otherwise, I doubt that there's enough of a performance issue to justify _adding_ complication.

My goal in deleting code is quite the opposite of that: I want to _remove_ complication.  

- This is an area that has historically been buggy.  No code = no bugs!
- But mostly: it's a burden for the human reader.  Whenever I get into one of those bugs that requires me to figure out what the solver is doing, I could do without having to worry about whether the `python_marker` is a thing that even matters

Secrus:
Now that we dropped Python 3.7 and `cached_property` availability is not an issue, are there any more blockers here?

dimbleby:
python 3.7 and cached_property are irrelevant here: per my last I have no intention of making code _more_ complicated, the goal is to delete it.

this is blocked only by a maintainer agreeing that that is a good idea

Secrus:
@dimbleby would you rebase? With 2.0 coming we are introducing breaking changes anyway, so we might as well break this too

radoering (changes requested):
> With 2.0 coming we are introducing breaking changes anyway, so we might as well break this too

In general, I agree...

> Actually the export plugin is an example of a project that relies on the python_marker, but only for the root package.

...but we should at least check our plugins and prepare them for the removal if we remove something they are still using.

src/poetry/core/packages/package.py
```diff
@@ -283,18 +280,11 @@ def python_versions(self, value: str) -> None:
...
-        )
 
     @property
     def python_constraint(self) -> VersionConstraint:
         return self._python_constraint
 
-    @property
-    def python_marker(self) -> BaseMarker:
-        return self._python_marker
-
```
radoering, line 297: I think I am still going to deprecate it instead of removing it because we already have released a `poetry-plugin-export` version, which claims to be compatible with Poetry 2.0 - and really is so far if I have not missed anything - but would not if we removed this property.

radoering:
By the way, I renamed the PR because `Package.with_python_versions()` has already been removed in https://github.com/python-poetry/poetry-core/commit/01b58ea0cc2e501c5fad70acbc356930975c6ad9 so only `Package.python_marker` is left.

radoering (approved):

PR #447, opened 2022-08-24
add setter for version on ProjectPackage

bcavagnolo:
This enables support for dynamically setting the package version
(e.g., from a plugin)

Resolves: python-poetry/poetry#5493

Resolves: python-poetry#


radoering:
I just added a test that the hash does not change and removed the todo. Iirc, we've added `type ignore` in such cases before.

radoering (approved):

PR #460, opened 2022-09-02
Fix problem with deleting temporary folders on Windows

eblis:
This is another attempt for pull request https://github.com/python-poetry/poetry/pull/5171

From what I can tell the relevant code has been moved from `poetry` to `poetry-core`, I've created a new pull request targeting this repository.

Resolves: https://github.com/python-poetry/poetry/issues/1031


da1910:
This issue still exists (albeit intermittently) and is again blocking our CI process, is there anything we as users can do to get this fix expedited?

Andy-Grigg:
I'm seeing this issue as well. From what I can tell, this fix will resolve the issue and stop our intermittent Windows build failures.

HenriqueAJNB:
Hi, I'm waiting for this PR to solve. Can anyone give us some feedback on how is it going? =)

DudeNr33:
I would also really appreciate this fix, as I am once again sitting in front of my computer, restarting poetry commands over and over for simply adding a new dependency.

src/poetry/core/utils/helpers.py
```diff
@@ -91,6 +92,27 @@ def safe_rmtree(path: str | Path) -> None:
...
+        try:
+            shutil.rmtree(path)
+            return  # Only hits this on success
+        except OSError:
+            # Increase the timeout and try again
+            time.sleep(timeout)
+            timeout *= 2
+
+    # Final attempt, pass any Exceptions up to caller.
+    safe_rmtree(path)
```
cdgriffith, line 113: Wouldn't we want to start with the is_symlink logic from `safe_rmtree` itself?

If we do that, we can remove the `safe_rmtree` function entirely as this is the only place it's used. 

```suggestion
    if Path(path).is_symlink():
        return os.unlink(str(path))

    timeout = 0.001
    while timeout < max_timeout:
        try:
            shutil.rmtree(path)
            return  # Only hits this on success
        except OSError:
            # Increase the timeout and try again
            time.sleep(timeout)
            timeout *= 2

    # Final attempt, pass any Exceptions up to caller.
     shutil.rmtree(path, onerror=_on_rm_error)
```

eblis: I'm happy with any solution as long as it's working on Windows.

eblis: I've taken the code from @cdgriffith, but with a small change.

From what I can see `os.unlink` can also throw exceptions on Windows if the resource is in use, so I added the check and call for this in the big while loop to catch this scenario as well.

da1910 (commented):
I'm not a maintainer on poetry or poetry-core, but it looks like the changes you've made are the ones requested in the previous review. Additionally the changes in general worked for another project I'm working on when deleting a large number of directories.

eblis:
@da1910 sorry, github suggested you as a reviewer so I just went ahead and accepted.
I hope it doesn't take too long for this to get reviewed again, or else we'll have the same problem with stale code yet again.

xmnlab:
that would be great to see this merged soon <3 this problem is affecting me in a CI workflow for windows
thanks @eblis for working on that

radoering (changes requested):
LGTM in general. I think we want to change the interface so that `path` must be a `Path` and consequently use `Path` methods.

Further, I think we can improve the cleanup in the test.

dimbleby:
https://github.com/python/cpython/pull/24793 adds `ignore_cleanup_errors` to the parameters of `tempfile.TemproraryDirectory()`, from python 3.10 onwards - which seems to be aimed at exactly this.

So perhaps the implementation of `temporary_directory()` should go: "if python >= 3.10 then use `TemporaryDirectory(ignore_cleanup_errors=True)`, else do this much more complicated thing"

Or perhaps - given how longstanding this must be - poetry should simply use `TemporaryDirectory` directly, with `ignore_cleanup_errors=True` where available.  Windows users on older pythons must be an ever-smaller minority.

src/poetry/core/utils/helpers.py
```diff
@@ -90,10 +91,29 @@ def _on_rm_error(func: Any, path: str | Path, exc_info: Any) -> None:
     func(path)
 
 
-def safe_rmtree(path: str | Path) -> None:
-    if Path(path).is_symlink():
-        return os.unlink(str(path))
-
+def robust_rmtree(path: str, max_timeout: float = 1) -> None:
```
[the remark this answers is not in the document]
eblis: I see that `_on_rm_error` accepts `str | Path`, maybe we should accept that instead too ?
And the first thing we do in the function we just go `path = Path(path)` to make sure it's a `Path` object.
How does that sound ?

src/poetry/core/utils/helpers.py
```diff
@@ -42,7 +43,7 @@ def normalize_version(version: str) -> str:
 def temporary_directory(*args: Any, **kwargs: Any) -> Iterator[str]:
     name = tempfile.mkdtemp(*args, **kwargs)
     yield name
-    safe_rmtree(name)
+    robust_rmtree(name)
```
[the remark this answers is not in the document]
eblis: I changed `robust_rmtree` to accept both `str` and `Path`, like `_on_rm_error` does so I left this call as is

src/poetry/core/utils/helpers.py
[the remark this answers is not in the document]
eblis: I modified `robust_rmtree` to accept both `str` and `Path`, just like `_on_rm_error` does.

src/poetry/core/utils/helpers.py
```diff
@@ -90,10 +91,29 @@ def _on_rm_error(func: Any, path: str | Path, exc_info: Any) -> None:
...
+    If the final attempt fails, the Exception is propagated
+    to the caller.
+    """
+    timeout = 0.001
+    while timeout < max_timeout:
+        try:
+            # both os.unlink and shutil.rmtree can throw exceptions on Windows
+            # if the files are in use when called
+            if Path(path).is_symlink():
+                os.unlink(str(path))
```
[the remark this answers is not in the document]
eblis: Changed per request

da1910:
> [python/cpython#24793](https://github.com/python/cpython/pull/24793) adds `ignore_cleanup_errors` to the parameters of `tempfile.TemporaryDirectory()`, from python 3.10 onwards - which seems to be aimed at exactly this.
> 
> So perhaps the implementation of `temporary_directory()` should go: "if python >= 3.10 then use `TemporaryDirectory(ignore_cleanup_errors=True)`, else do this much more complicated thing"
> 
> Or perhaps - given how longstanding this must be ([python-poetry/poetry#1031](https://github.com/python-poetry/poetry/issues/1031)) - poetry should simply use `TemporaryDirectory` always, with `ignore_cleanup_errors=True` where available. Windows users on older pythons must be an ever-smaller minority.

I would strongly favour making some effort to fix this for python 3.8 and 3.9, the fact that it's taken this long to address isn't really a reason to wash our hands of support, Python 3.9 is supported for another two and a half years after all!

da1910:
> > Python 3.9 is supported for another two and a half years after all!
> 
> Sure. However, the current situation has been tolerated for all of that and then some; and for the huge majority who are not using older Pythons on Windows, I'd have a _much_ higher degree of confidence in the standard library's temporary-directory implementation than in a home-grown implementation .
> 
> So I do think it would be preferable in general to use `TemporaryDirectory()`. I don't have very strong opinions about whether Windows with python 3.8 and 3.9 are important; but IMO it would at least be not absurd to choose to keep poetry code simple and decline to fix this for them.

I would agree about simplicity for certain, but I'm going to disagree with you about the scope of the problem. I've personally wasted several days of my life rerunning CI because of this issue, and there are plenty of other people on the original thread who have had the same experience for the last four years (I've only been using poetry 'in anger' for the last year or so).

Windows may be a minority for open-source developers, but for casual python users it's vastly the most common OS, and to have a project like poetry deprioritise fixes for windows specific bugs with in-support python versions seems slightly unfriendly.

Secrus:
> So perhaps the implementation of `temporary_directory()` should go: "if python >= 3.10 then use `TemporaryDirectory(ignore_cleanup_errors=True)`, else do this much more complicated thing"
> 

This is the solution I would go for and which is the most likely going to be accepted.

dimbleby:
"unfriendly" isn't the aim, more like "pragmatic"!

Windows makes up less than 3% of poetry's downloads according to https://pypistats.org/packages/poetry - bottom of page.  This is surely a very imperfect measure but, so far as I know, the best we have.

I don't know how to split this up by python version, and I certainly don't know how to split this up by "open-source developers" vs "casual users".  Either way I suppose it's true that 3% of poetry's users today is a larger absolute number than 3% of poetry's users was in 2019.

Anyway: if maintainers prefer to take on the additional complication, I've no fight to pick with that.

dimbleby:
re the typing errors in the pipeline: so far as I can see no-one ever passes any arguments to `temporary_directory()`, so I'd just remove the `args` and `kwargs` altogether.

eblis:
> re the typing errors in the pipeline: so far as I can see no-one ever passes any arguments to `temporary_directory()`, so I'd just remove the `args` and `kwargs` altogether.

I saw this message too late, spent a bit of time fighting with `mypy` and in the end just added the argument to kwargs before calling the function, seemed to please `mypy`.

xmnlab:
> Windows makes up less than 3% of poetry's downloads according to https://pypistats.org/packages/poetry


but maybe this analysis about this number could be a bit biased ... because poetry would be used more by devs and on CI ... 
the experience programming on windows is not that great  ... so probably most python dev doesn't use windows (IMHO) ... and if we it is not working on CI using windows, no one is  testing their packages with poetry for windows. (that is my case)

so I guess that this number makes sense. but maybe if the poetry works well for windows .. that number will increase (at least a bit)

xmnlab:
and just as an extra info, on conda/conda-forge the downloads for windows is almost the same number for osx (for the latest version there): 
https://anaconda.org/conda-forge/poetry/files

tests/utils/test_helpers.py
```diff
@@ -118,3 +126,60 @@ def test_utils_helpers_readme_content_type(
...
+    mocked_rmtree.side_effect = [
+        OSError(
+            "Couldn't delete file yet, waiting for references to clear", "mocked path"
+        ),
+        None,
+    ]
+    robust_rmtree(name)
+
+    # this should give up after retrying multiple times
+    name = tempfile.mkdtemp()
```
radoering, line 175:
```suggestion
```

Since we've mocked `rmtree`, the first temporary directory has not been deleted, so we do not have to create a new one here.

tests/utils/test_helpers.py
```diff
@@ -118,3 +126,60 @@ def test_utils_helpers_readme_content_type(
...
+    mocked_rmtree.side_effect = OSError(
+        "Couldn't delete file yet, this error won't go away after first attempt"
+    )
+    with pytest.raises(OSError):
+        robust_rmtree(name, max_timeout=0.04)
+
+    # clear the side effect (breaks the tear-down otherwise)
+    mocker.stop(mocked_rmtree)
+    # use the real method to remove the temp folder we created for this test
+    robust_rmtree(name)
```
radoering, line 185:
```suggestion
    robust_rmtree(name)
    assert not Path(name).exists()
```

For completeness, we can check the result of the lucky path without the mock. 😄

radoering (changes requested):
I think we now have a compromise that each side can live with.

Just some minor nitpicks regarding the test before merging it.

radoering (approved):

PR #469, opened 2022-09-11
Don't normalize or escape versions

dimbleby:
Per https://github.com/python-poetry/poetry/pull/6476

- `normalize_version()` is only ever used as an over-elaborate way of calling `Version.to_string()`
- `escape_version()` is actually counter-productive

I've aggressively removed the functions altogether, so would need to merge https://github.com/python-poetry/poetry/pull/6476 before this one (so that poetry no longer uses them).  

It would be possible to leave them around unused for a bit, I favour cleaning such things up.


dimbleby:
(don't understand why the downstream tests didn't fail, this MR surely removes things that are not yet removed downstream)

radoering:
> It would be possible to leave them around unused for a bit, I favour cleaning such things up.

I'd prefer a deprecation warning as in https://github.com/python-poetry/poetry-core/blob/37cee90a5dd4c7ee2c5ee836216ba813242b3ade/src/poetry/core/version/pep440/version.py#L234-L240

When removing these functions the next core release will break poetry 1.2.0 unless we bump the major version of core. We could do this, but I don't think there is consensus among maintainers to release one major version after the other.

dimbleby:
I'm going to not do any deprecating for a bit in the hope that y'all reach a consensus that I like better...  

If for some reason you decide that merging this is urgent, feel free to reintroduce deprecated versions of the functions.

dimbleby:
> When removing these functions the next core release will break poetry 1.2.0 unless we bump the major version of core

I just noticed that this is not true, because poetry 1.2.0 pins to the exact version 1.1.0 of poetry-core: https://github.com/python-poetry/poetry/blob/85993df4fa7b67dbfe573cce7b2a397846b9c320/pyproject.toml#L47

which in some ways doesn't make a lot of difference - I'm in favour of a major version bump here anyway.  But does mean that it would be _possible_ to make breaking changes here and not signal that in the version number if for some reason that was deemed desirable.

radoering:
I was just surprised to find a fix version in pyproject.toml. 😅

I agree a major bump would be more appropriate if we remove something although I still think deprecation warnings don't hurt and would allow faster progress...

dimbleby:
> I still think deprecation warnings don't hurt and would allow faster progress...

maybe so, but I don't expect ever to need this fix myself so I'm content to be stubborn about it!

If anyone does turn up who actually cares about the fix in here - eg epoch users per https://github.com/python-poetry/poetry/issues/6466 - then they are welcome to take this and rework it so that it only deprecates `escape_version` and `normalize_version`.

radoering (changes requested):
FYI: Since it's companion has been merged, I'll probably add the deprecation warning myself later.

radoering (approved):

PR #476, opened 2022-09-16
normalize extras per PEP685

dimbleby:
https://peps.python.org/pep-0685/#specification says both that all comparisons should be done with normalized extra names, and that normalized extra names should be written to metadata.

Simplest and surest approach seems to be just to deal with normalized names everywhere.

Fixes https://github.com/python-poetry/poetry/issues/6321.

will cause CI failures in the downstream tests because poetry now records normalized extra names in lockfiles.  I think that's correct and will follow up with a poetry MR to fix the unit tests against this.  

However this presents some poetry / poetry-core challenges re merge order and version numbering.  As in other MRs, I'd advocate for allowing such things and being comfortable going 2.0 here - though even then the cross-repository pipelines want some disentangling.

### DIFF
--- a/src/poetry/core/factory.py
+++ b/src/poetry/core/factory.py
@@ -11,6 +11,8 @@ from typing import Mapping
 from typing import Union
 from warnings import warn
 
+from packaging.utils import canonicalize_name
+
 from poetry.core.utils.helpers import combine_unicode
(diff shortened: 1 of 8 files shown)


radoering (approved):

radoering:
@dimbleby Please don't delete your branch until we don't depend on it in poetry anymore.

PR #477, opened 2022-09-19
fix: added 3.11 to list of available pythons

finswimmer:
Python 3.11 is just around the corner, so we have to update this list.

`Package.AVAILABLE_PYTHONS` is needed in two places:

1. Adding a classifier when building a wheel (https://github.com/python-poetry/poetry-core/blob/4535724acb356a9da2fa976a243e5f63bef892f8/src/poetry/core/packages/package.py#L309)
2. Finding a suitable interpreter when creating a venv (https://github.com/python-poetry/poetry/blob/2d09741ad67d4fabd94ce6ffd6257c5fec441c12/src/poetry/utils/env.py#L937)

I would prefer a more generic way. So if anyone has a suggestion ...

Downstream tests are fixed in https://github.com/python-poetry/poetry/pull/6561


mkniewallner:
Shouldn't this list be updated to? https://github.com/python-poetry/poetry-core/blob/d6d33de68e48cd2da4730dd01b69b8d32f51ae41/src/poetry/core/version/helpers.py#L13-L26

finswimmer:
> Shouldn't this list be updated to?
> 
> https://github.com/python-poetry/poetry-core/blob/d6d33de68e48cd2da4730dd01b69b8d32f51ae41/src/poetry/core/version/helpers.py#L13-L26

Good catch :+1:

mkniewallner (approved):

PR #482, opened 2022-09-23
Refactor constraints structure

radoering:
No functional changes, just moved two packages:

- `poetry.core.packages.constraints` -> `poetry.core.constraints.generic`
- `poetry.core.semver` -> `poetry.core.constraints.version`

I've always found it very confusing that the two types of constraints are distributed between `poetry.core.packages.constraints` and `poetry.core.semver`. It's not even clear anymore why the first one is below `packages` and why the second one is called `semver`. Of course, there is some semver related stuff in between the version constraints but the constraints are far from being limited to semver.

In my opinion moving these packages results in a clearer structure that makes it easier to navigate the code for new contributors.

In order to avoid breakage, I only deprecated and forwarded the old packages.

The commits should probably **NOT** be squashed so that the rename can be recognized in git history.


src/poetry/core/constraints/generic/__init__.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+__all__ = [
+    "AnyConstraint",
+    "BaseConstraint",
+    "Constraint",
+    "EmptyConstraint",
+    "MultiConstraint",
+    "UnionConstraint",
+    "parse_constraint",
```
mkniewallner, line 71: Shouldn't we keep `parse_single_constraint`, for backward compatibility?

radoering: I don't want to include it here (in the new package) because it shouldn't be used outside of this package (you don't need `parse_single_constraint` because you can always use `parse_constraint`). However, I see your point. We should make it available in the deprecated package. I'll do a little more refactoring.

mkniewallner (approved):

PR #484, opened 2022-09-25
Normalize sdist filenames

dimbleby:
Addresses https://github.com/python-poetry/poetry/issues/5782

Needs a companion fix in poetry proper: poetry "knows" the names of the sdists that poetry-core builds, so in case of mismatch source distributions won't be uploaded at all.

This MR leaves poetry insisting on normalizing the package name at upload (ie it did this normalization prior to this MR, and it still does).  That is: if you have a package `foo.bar` then that gets uploaded as `foo-bar`.  The advance that this MR makes is that both sdist and wheels are successfully uploaded now (and also poetry is more spec-compliant).

Currently it's not possible to (i) leave the package name unnormalized and also (ii) follow the packaging specs re wheel and sdist filenames - see https://github.com/pypi/warehouse/issues/10030.  So while in principle it's straightforward for poetry to do that - it should use the `pretty_name` on the metadata, not the canonicalized name - it would currently make things worse not better.


neersighted:
Haven't had time to take a close look yet -- where does this put us relative to [PEP 625](https://peps.python.org/pep-0625/)?

dimbleby:
this should be in agreement with that pep

radoering (approved):

radoering (approved):

PR #486, opened 2022-10-01
Fix sdist build for multiple readme files

wagnerluis1982:
Resolves: python-poetry/poetry#6633 

I fixed the issue by changing `Poetry.local_config["readme"]` type to a sequence. In the sdist builder, I am adding all the files found in that sequence.

- [x] Added downstream **tests**: python-poetry/poetry#6678
- No further **documentation** is required.


wagnerluis1982:
Downstream tests failed because of the change in `local_config`.

Before I implemented the current solution, I wondered if I should handle the multi-type (str | Sequence) nature of `readme` in sdist, which would solve the downstream test problem, but I wonder if this would not make recurring to inspect `readme` type in other builders, if appears.

neersighted (changes requested):
I don't think this fix is acceptable as-is -- in `factory.py` we explicitly handle the case of the readmes being a string or some other sequence (tuple/list), and I don't think it's unreasonable to do such downstream given it's in the schema.

The special casing and in-place mutation here is pretty gross and surprising, and hard to maintain. Let's properly handle both cases in the sdist builder, and then add some integration tests in Poetry.

wagnerluis1982:
> I don't think this fix is acceptable as-is -- in `factory.py` we explicitly handle the case of the readmes being a string or some other sequence (tuple/list), and I don't think it's unreasonable to do such downstream given it's in the schema.

I was expecting this :laughing:, thank you for the feedback. I will proceed that way.

src/poetry/core/masonry/builders/sdist.py
```diff
@@ -328,9 +329,13 @@ def find_files_to_add(self, exclude_build: bool = False) -> set[BuildIncludeFile
         # Include project files
         additional_files.add(Path("pyproject.toml"))
 
-        # add readme if it is specified
+        # add readme files if it is specified
         if "readme" in self._poetry.local_config:
-            additional_files.add(self._poetry.local_config["readme"])
+            r: str | Sequence = self._poetry.local_config["readme"]
```
neersighted: A `str` is a `Sequence`, so this type can be simplified.

neersighted (changes requested):

tests/masonry/builders/test_sdist.py
```diff
@@ -191,6 +191,24 @@ def test_find_files_to_add() -> None:
...
 
+def test_find_files_to_add_with_multiple_readme_files() -> None:
+    poetry = Factory().create_poetry(
+        Path(__file__).parent.parent.parent / "fixtures" / "with_readme_files"
+    )
+
+    builder = SdistBuilder(poetry)
+    result = [f.relative_to_source_root() for f in builder.find_files_to_add()]
+
+    assert sorted(result) == sorted(
```
neersighted, line 202: Use of sets here would better convey the intentions.

neersighted (changes requested):

src/poetry/core/masonry/builders/sdist.py
wagnerluis1982: I figured out this is wrong, it should be `Path | Iterable[Path]`...

BTW the pre-commit is not running mypy on my local... was it disabled?

tests/masonry/builders/test_sdist.py
wagnerluis1982: This was a no-brainer, I copied from the existing test above :sweat_smile:

neersighted: If you could change the above to use a set comprehension and set literal instead of `sorted(list())` that would be great 😄

src/poetry/core/masonry/builders/sdist.py
neersighted: I don't think `Path | ` is correct -- are we actually turning these into paths? I thought they were strings? If you're not sure, salt some `reveal_type()` in.

Also, `mypy` has to be run manually -- running it in pre-commit is not possible as we need our dependencies installed. `mypy` and `pytest` are run separately or provided by your IDE, and `pre-commit` runs everything else.

tests/masonry/builders/test_sdist.py
wagnerluis1982: Great idea, just did :+1:

tests/masonry/builders/test_sdist.py
```diff
@@ -187,7 +187,25 @@ def test_find_files_to_add() -> None:
...
+
+def test_find_files_to_add_with_multiple_readme_files() -> None:
+    poetry = Factory().create_poetry(
+        Path(__file__).parent.parent.parent / "fixtures" / "with_readme_files"
+    )
+
+    builder = SdistBuilder(poetry)
+    result = {f.relative_to_source_root() for f in builder.find_files_to_add()}
+
+    assert result == (
```
neersighted, line 202: These parenthesis are meaningless -- let's drop them.

src/poetry/core/masonry/builders/sdist.py
```diff
@@ -320,17 +321,21 @@ def find_files_to_add(self, exclude_build: bool = False) -> set[BuildIncludeFile
...
         additional_files.update(self.convert_script_files())
 
         # Include project files
         additional_files.add(Path("pyproject.toml"))
 
-        # add readme if it is specified
+        # add readme files if it is specified
         if "readme" in self._poetry.local_config:
-            additional_files.add(self._poetry.local_config["readme"])
+            r: str | Path | Iterable[Path] = self._poetry.local_config["readme"]
```
neersighted, line 334: This seems even more convoluted -- can you tell me where we are getting a conversion to `Path` from?

src/poetry/core/masonry/builders/sdist.py
```diff
@@ -320,17 +321,21 @@ def find_files_to_add(self, exclude_build: bool = False) -> set[BuildIncludeFile
         to_add = super().find_files_to_add(exclude_build)
 
         # add any additional files, starting with all LICENSE files
-        additional_files = set(self._path.glob("LICENSE*"))
+        additional_files: set[str | Path] = set(self._path.glob("LICENSE*"))
```
neersighted, line 324: Seems like these should all be `Path` -- when can `glob` return a `str`? `convert_script_files` is in this file and can only return `Path`.

neersighted (dismissed):

src/poetry/core/masonry/builders/sdist.py
wagnerluis1982: For the `Path |` inquiry, it's a bit hard to see.

What I discovered
1. Just after readme files are added to `additional_files`, there is the following loop to handle these files

    ```python
    for additional_file in additional_files:
        file = BuildIncludeFile(
            path=additional_file, project_root=self._path, source_root=self._path
        )
        ...
    ```

2. So, as you see `BuildIncludeFile` is called, this class has the following signature (note `path: Path | str`):

    ```python
    class BuildIncludeFile:
        def __init__(
            self,
            path: Path | str,
            project_root: Path | str,
            source_root: Path | str | None = None,
        ) -> None:
    ```

3. Another evidence, `additional_files` is initialized with `set(self._path.glob("LICENSE*"))`, that is typed as `set[Path]`.

Thus, I updated the code to consider the following types:
```python
additional_files: set[str | Path]
```
```python
r: str | Path | Iterable[str | Path]
```

src/poetry/core/masonry/builders/sdist.py
wagnerluis1982: `BuildIncludeFile` is handling the `str` to `Path` conversion

wagnerluis1982: In addition, I just confirmed `local_config["readme"]` is actually `str` only (or `Iterable[str]`.

I added `print(type(local_config["readme"][0]))` statement to `Poetry` constructor on my patched poetry-core lib and run `poetry build` on a project sample with 2 readme files.

The result was `<class 'tomlkit.items.String'>`, which is a subclass of `str`, declared as `class String(str, Item)`.

src/poetry/core/masonry/builders/sdist.py
wagnerluis1982: See the other discussion, I need to also add `str`, since is what `poetry build` add to `local_config["readme"]`.

Of course I can convert to `Path` when I add to the set, it is an idea :bulb:

src/poetry/core/masonry/builders/sdist.py
wagnerluis1982: I pushed a change typing `local_config["reade"]` simply as `str | Iterable[str]`

tests/masonry/builders/test_sdist.py
wagnerluis1982: A misstep, just adjusted :slightly_smiling_face:

wagnerluis1982:
@neersighted I added downstream tests to python-poetry/poetry#6678.

If you have something else in mind regarding this PR, let me know :-)

wagnerluis1982:
@neersighted this PR has exact 1 month. I've been told the next poetry release is nigh. How about to continue the discussions for this fix be included?

src/poetry/core/masonry/builders/sdist.py
wagnerluis1982: Making sure I'm understandable.

Variable `additional_files` evaluates **initially** after call to `glob` as a `set[Path]`. Period.

But then, I also need to include to the `set` the readme file(s) from `local_config["readme"]` which is a `str` or a `list[str]`.

wagnerluis1982: Evidences of `local_config["readme"]` be a `str` or `list[str]`

https://github.com/python-poetry/poetry-core/blob/0759db61f01f008edc3ffe2b790ac9ecc892da77/src/poetry/core/factory.py#L48

https://github.com/python-poetry/poetry-core/blob/0759db61f01f008edc3ffe2b790ac9ecc892da77/src/poetry/core/pyproject/toml.py#L72-L74

wagnerluis1982:
@neersighted would you be able to check this PR? 🎅

src/poetry/core/masonry/builders/sdist.py
```diff
@@ -322,17 +323,21 @@ def find_files_to_add(self, exclude_build: bool = False) -> set[BuildIncludeFile
...
 
         # add script files
         additional_files.update(self.convert_script_files())
 
         # Include project files
         additional_files.add(Path("pyproject.toml"))
 
-        # add readme if it is specified
+        # add readme files if it is specified
```
radoering, line 334:
```suggestion
        # add readme files if specified
```

src/poetry/core/masonry/builders/sdist.py
radoering: Although it's not necessary, I think it might be worth to just make sure that `additional_files` are always of type `Path` (for cleaner code and to avoid confusion). Proposal see below.

src/poetry/core/masonry/builders/sdist.py
```diff
@@ -322,17 +323,21 @@ def find_files_to_add(self, exclude_build: bool = False) -> set[BuildIncludeFile
...
         if "readme" in self._poetry.local_config:
-            additional_files.add(self._poetry.local_config["readme"])
+            r: str | Iterable[str] = self._poetry.local_config["readme"]
+            if isinstance(r, str):
+                additional_files.add(r)
+            else:
+                additional_files.update(r)
```
radoering, line 340:
```suggestion
            readme: str | Iterable[str] = self._poetry.local_config["readme"]
            if isinstance(readme, str):
                additional_files.add(Path(readme))
            else:
                additional_files.update(Path(r) for r in readme)
```

radoering (changes requested):
All in all, there seem to be only minor concerns here. I think with some small changes this can be merged.

src/poetry/core/masonry/builders/sdist.py
```diff
@@ -322,17 +323,21 @@ def find_files_to_add(self, exclude_build: bool = False) -> set[BuildIncludeFile
         to_add = super().find_files_to_add(exclude_build)
 
         # add any additional files, starting with all LICENSE files
-        additional_files = set(self._path.glob("LICENSE*"))
+        additional_files: set[str | Path] = set(self._path.glob("LICENSE*"))
```
wagnerluis1982, line 326:
```suggestion
        additional_files: set[Path] = set(self._path.glob("LICENSE*"))
```

src/poetry/core/masonry/builders/sdist.py
wagnerluis1982: I liked this way. I was trying to avoid extra work, but I completely agree that makes much cleaner.

wagnerluis1982:
@radoering 
> All in all, there seem to be only minor concerns here. I think with some small changes this can be merged.

Thank you for the suggestions, I just applied them :slightly_smiling_face:

radoering (approved):

PR #488, opened 2022-10-02
dependency: consider source_name in is_same_source_as

radoering:
As per https://github.com/python-poetry/poetry/pull/6679#discussion_r985253280


radoering:
Hmm, looking at the downstream tests perhaps not such a good idea after all. Have to take a closer look later.

radoering:
Discarding this for now because it breaks the heart of the dependency resolver. [_compatible_dependency()](https://github.com/python-poetry/poetry/blob/2c610de46805f6f0a31aacaf394c598a3bd9e122/src/poetry/mixology/term.py#L154) compares dependencies with `source_name` with dependencies created from packages via `to_dependency()` which of course never have a `source_name`. Changing that would probably require a major refactoring.

PR #492, opened 2022-10-05
Revert "prefer to avoid casts (#468)"

neersighted:
This reverts commit 743e09cd9014e1638a980be2d8814c9201f91347.

At runtime, during consumption via the prepare_metadata_for_build_wheel,
API, some of these values were in fact not what their typing was
asserted to be. This would cause a failure when the assertion failed.

### DIFF
--- a/src/poetry/core/constraints/generic/union_constraint.py
+++ b/src/poetry/core/constraints/generic/union_constraint.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+from typing import cast
+
 from poetry.core.constraints.generic.base_constraint import BaseConstraint
 from poetry.core.constraints.generic.constraint import Constraint
 from poetry.core.constraints.generic.empty_constraint import EmptyConstraint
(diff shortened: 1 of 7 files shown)


mkniewallner (approved):

Secrus (approved):

dimbleby:
`OutOfOrderTableProxy`?!?  aren't tomlkit types fun?

full reversion of this MR is an overshoot but I suppose it's simple enough to revert most of the revert

dimbleby:
```patch
diff --git a/src/poetry/core/pyproject/toml.py b/src/poetry/core/pyproject/toml.py
index 6c8e79c..6e55912 100644
--- a/src/poetry/core/pyproject/toml.py
+++ b/src/poetry/core/pyproject/toml.py
@@ -64,14 +64,14 @@ class PyProjectTOML:
         return self._build_system

     @property
-    def poetry_config(self) -> Table:
+    def poetry_config(self) -> dict[str, Any]:
         from tomlkit.exceptions import NonExistentKey

         try:
             tool = self.data["tool"]
             assert isinstance(tool, dict)
             config = tool["poetry"]
-            assert isinstance(config, Table)
+            assert isinstance(config, dict)
             return config
```
should be enough

neersighted:
I'm going with the full revert for 1.3.1 -- a PR with a V2 that fixes the assertion is very welcome.

PR #497, opened 2022-10-08
Package.satisfies() considers sources more carefully

dimbleby:
`Package.satisfies(dependency)` has been asking the wrong question about the sources: it doesn't need the sources to be identical, it needs the package source to satisfy the dependency source.

These are different mostly because of the confusing handling of pypi and legacy repositories
- on a dependency, setting source type and source name to None means "any source is fine"
- on a dependency, setting source type to None but specifying source name means "this specific repository"
- on a package, setting source type to None means "pypi"

I've also removed the check on features: it's not wanted by callers of this method in poetry, so it was just getting in the way.

A companion MR in poetry will follow, between them these should fix https://github.com/python-poetry/poetry/issues/6710

### DIFF
--- a/src/poetry/core/packages/package.py
+++ b/src/poetry/core/packages/package.py
@@ -556,16 +556,47 @@ class Package(PackageSpecification):
         """
         Helper method to check if this package satisfies a given dependency.
 
-        This is determined by assessing if this instance provides the package and
-        features specified by the given dependency. Further, version and source
-        types are checked.
+        This is determined by assessing if this instance provides the package specified
+        by the given dependency. Further, version and source types are checked.
         """
-        if not self.provides(dependency) or not dependency.constraint.allows(
-            self.version
-        ):
+        if self.name != dependency.name:
             return False
 
-        return ignore_source_type or self.is_same_source_as(dependency)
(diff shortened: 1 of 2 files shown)


neersighted:
In the medium term [PEP 658](https://peps.python.org/pep-0658/) should hopefully be supported by PyPI, and `PypiRepository` and `LegacyRepository` will collapse into `SimpleRepository` or similar, with a type of `simple` or `index`. That will clean this up quite a bit. In the medium term, how invasive would it be to make the `source_type` for `PypiRepository` `pypi` to reduce this ambiguity?

dimbleby:
> how invasive would it be to make the source_type for PypiRepository pypi

moderately.  eg this gets written to the lockfile so it's a breaking change there.

neersighted:
We have yet to release lock file 2.0 into the wild, so this is a very convenient time to do it.

neersighted:
691 is the JSON format and already supported -- I think you mean 658, which is the METADATA file.

You're not wrong and your comment in this PR is more than sufficient for future code spelunkers. It was simply an observation that if you wanted to solve this wart, this is an excellent time.

radoering (approved):

PR #500, opened 2022-10-11
Add integration test for project using repository `poetry-core` as a backend

mkniewallner:
Add an integration test to ensure that a project using the repository version of `poetry-core` is installable via `pip`.

Since it's not possible to use relative paths in `build-system.requires`, the test dynamically adds it by copying the project fixture into a temporary directory, then appending the definition based on the path location.

https://github.com/python-poetry/poetry-core/commit/7c474fdd16c88e6700733459f64809d0d4a93893 also slightly updates some fixtures used by unit tests to ensure we don't re-introduce https://github.com/python-poetry/poetry-core/pull/492.

### DIFF
--- a/tests/fixtures/sample_project/pyproject.toml
+++ b/tests/fixtures/sample_project/pyproject.toml
@@ -50,6 +50,10 @@ dataclasses = {version = "^0.7", python = ">=3.6.1,<3.7"}
 [tool.poetry.extras]
 db = [ "orator" ]
 
+# Allows testing https://github.com/python-poetry/poetry-core/pull/492
+[tool.black]
+preview = true
+
 [tool.poetry.group.dev.dependencies]
 pytest = "~3.4"
 
--- a/tests/fixtures/sample_project/pyproject.toml
+++ b/tests/fixtures/sample_project/pyproject.toml
@@ -50,7 +50,12 @@ dataclasses = {version = "^0.7", python = ">=3.6.1,<3.7"}
 [tool.poetry.extras]
 db = [ "orator" ]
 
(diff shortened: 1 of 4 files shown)


tests/fixtures/sample_project/pyproject.toml
```diff
@@ -50,6 +50,10 @@ dataclasses = {version = "^0.7", python = ">=3.6.1,<3.7"}
 [tool.poetry.extras]
 db = [ "orator" ]
 
+# Allows testing https://github.com/python-poetry/poetry-core/pull/492
+[tool.black]
```
neersighted, line 54: Can you elaborate on the purpose of this bit?

mkniewallner: Oh yeah, since the issue we had in 1.3.0 is rather fresh in my head, I thought that the link to the revert would give enough context, but not at all.

Basically, the issue in the original PR is that `tomlkit` can either return a `Table` for TOML tables or `OutOfOrderTableProxy` if the table is discontinuous and separated by unrelated tables.
As we explicitly asserted that `tool.poetry` was a `Table`, we ended up raising errors for people having discontinuous sections for `tool.poetry` (who would end up with `OutOfOrderTableProxy` instances).

So this change is mostly a non-regression test to ensure that we don't reintroduce the issue. It might be worth having a dedicated test for it rather than updating an existing fixture used in multiple tests, although this edge case is rather specific to `tomlkit`, which sooner or later will go away, so not sure it's worth doing so.

Regardless, the comment is clearly not explicit enough, so it would need to be updated.

neersighted (approved):

PR #504, opened 2022-10-11
ci: upgrade `pip` for 3.11 on all OSes

mkniewallner:
It looks like the random 3.11 failure does not only affect macOS, as can be seen [here](https://github.com/python-poetry/poetry-core/actions/runs/3230909619/jobs/5289871785), so let's upgrade pip for all OSes.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -64,7 +64,7 @@ jobs:
 
       # XXX: https://github.com/pypa/pip/issues/11352 causes random failures -- remove once fixed in a release.
       - name: Upgrade pip on 3.11 for macOS
-        if: ${{ matrix.python-version == '3.11-dev' && matrix.os == 'macOS' }}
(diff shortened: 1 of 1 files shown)


neersighted:
Can you port this to Poetry itself as well?

neersighted (approved):

.github/workflows/tests.yml
```diff
@@ -64,7 +64,7 @@ jobs:
 
       # XXX: https://github.com/pypa/pip/issues/11352 causes random failures -- remove once fixed in a release.
       - name: Upgrade pip on 3.11 for macOS
```
neersighted, line 66:
```suggestion
      - name: Upgrade pip on Python 3.11
```

neersighted (approved):

.github/workflows/tests.yml
mkniewallner: Woops, thanks for catching that.

PR #507, opened 2022-10-20
Allow non-existant path dependencies

adriangb:
Currently `poetry lock --no-update` fails when you remove any path dependency because it first loads all dependencies from the lock file (which still has a reference to the dependency you just deleted) and it fails at that point because when `DirectoryDependency` is instantiated with a non-existent path it immediately throws an exception.

This change allows the process to continue, which is the same behavior VCS dependencies have.

If the `pyproject.toml` has a path dependency that doesn't exist we will want to error for `poetry install` and `poetry lock` (`poetry build` with path dependencies is already kinda a no-go). `poetry install` already fails gracefully (if you lock the project then delete the path dependency and try to install; i.e. when no locking happens before we start installing), I opened https://github.com/python-poetry/poetry/pull/6844 to make `poetry lock` fail gracefully.

### DIFF
--- a/src/poetry/core/packages/directory_dependency.py
+++ b/src/poetry/core/packages/directory_dependency.py
@@ -13,6 +13,8 @@ from poetry.core.pyproject.toml import PyProjectTOML
 
 
 class DirectoryDependency(Dependency):
+    _full_path: Path | None
+
     def __init__(
         self,
         name: str,
@@ -25,10 +27,10 @@ class DirectoryDependency(Dependency):
     ) -> None:
         self._path = path
         self._base = base or Path.cwd()
-        self._full_path = path
+        self._full_path = None
         self._develop = develop
 
-        self._validate()
+        full_path = self._get_full_path(False, name)
 
         super().__init__(
             name,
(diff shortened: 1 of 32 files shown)


adriangb:
I looked into the version history for this file to see if I could find some rational for the way it currently works. This goes all the way back to the "Initial commit" https://github.com/python-poetry/poetry-core/blob/d095e4a4af8f190d55199dd6ee60f4437143800e/poetry_core/packages/directory_dependency.py#L35 so I guess at this point only @sdispater knows.

adriangb:
@radoering any thoughts on this?

radoering:
I think it makes sense, but I have to take a deeper look if python-poetry/poetry#6844 is sufficient (i.e. covers all commands). For example, does `poetry show` still fails gracefully?

Further, I think we should do a similar change for file dependencies because it's the same use case.

radoering:
related: python-poetry/poetry#668, #210

src/poetry/core/packages/directory_dependency.py
```diff
@@ -55,10 +40,44 @@ def __init__(
...
+        #   we'll delete it from the lockfile, but we need to load it from the
+        #   lockfile first to begin solving.
+        # - A user wants to dynamically link to / insert the dependency at runtime.
+        #   This may seem a bit strange but in theory is totally valid and we should
+        #   allow it.
+        if not self._path.is_absolute():
+            try:
+                self._full_path = self._base.joinpath(self._path).resolve()
+            except FileNotFoundError:
+                warn(f"Directory {self._path} does not exist", category=UserWarning)
```
radoering, line 66: I like mentioning the name of the dependency (how you did in the companion PR). Maybe, we can add it in the warning/error messages here, too.

src/poetry/core/packages/directory_dependency.py
```diff
@@ -55,10 +40,44 @@ def __init__(
             source_url=self._full_path.as_posix(),
             extras=extras,
         )
-
         # cache this function to avoid multiple IO reads and parsing
         self.supports_poetry = functools.lru_cache(maxsize=1)(self._supports_poetry)
 
+    def _validate(self) -> None:
```
radoering, line 46: I think we should try the approach mentioned in https://github.com/python-poetry/poetry-core/pull/210#issuecomment-1159010010. We need two modes for `_validate`: a warning mode and an exception mode. The warning mode can be used in the `__init__` and the exception mode in `full_path`. I think it should be save to assume that when accessing `full_path` the directory must exist (and be a valid project).

radoering (changes requested):

tests/test_factory.py
```diff
@@ -298,18 +298,6 @@ def test_create_poetry_omits_dev_dependencies_iff_with_dev_is_false() -> None:
...
-):
-    with pytest.raises(ValueError) as err:
-        Factory().create_poetry(fixtures_dir / "project_with_invalid_dev_deps")
-    assert "does not exist" in str(err.value)
-
-    Factory().create_poetry(
-        fixtures_dir / "project_with_invalid_dev_deps", with_groups=False
-    )
-
-
```
adriangb, line 312: Removing this test since we actually want to be able to create a Poetry object with invalid path deps, we just want it to fail for certain operations (install, show, etc.).

tests/packages/test_file_dependency.py
```diff
@@ -22,7 +22,7 @@
 
 def test_file_dependency_wrong_path() -> None:
     with pytest.raises(ValueError):
-        FileDependency("demo", DIST_PATH / "demo-0.2.0.tar.gz")
+        FileDependency("demo", DIST_PATH / "demo-0.2.0.tar.gz").full_path
```
adriangb, line 25: I think these tests should be testing each individual failure mode, but this was good enough for years so I suppose it will do

src/poetry/core/packages/file_dependency.py
```diff
@@ -43,10 +35,40 @@ def __init__(
             optional=optional,
             allows_prereleases=True,
             source_type="file",
-            source_url=self._full_path.as_posix(),
+            source_url=full_path.as_posix(),
             extras=extras,
         )
 
+    def _get_full_path(self, raise_if_invalid: bool, name: str) -> Path:
```
adriangb, line 42: Adding `name` was necessary because `.name` doesn't get set until `super().__init__` is called which already requires `full_path`.

src/poetry/core/packages/file_dependency.py
```diff
@@ -43,10 +35,40 @@ def __init__(
...
+        if self._full_path is not None:
+            return self._full_path
```
adriangb, line 44: This little bit of caching means that if we already checked in `__init__` we don't need to check again. We could also cache the negative case but it doesn't seem worth the complexity IMO.

src/poetry/core/packages/directory_dependency.py
```diff
@@ -55,10 +40,44 @@ def __init__(
...
+
+        if not self._full_path.exists():
+            warn(f"Directory {self._path} does not exist", category=UserWarning)
+            return
+
+        if self._full_path.is_file():
+            raise ValueError(f"{self._path} is a file, expected a directory")
+
+        if not is_python_project(self._full_path):
+            raise ValueError(
```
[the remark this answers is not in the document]
adriangb: I'd like to push back on this a bit. There's a real use case for creating a `DirectoryDependency` pointing to a path that does not exist: you deleted or renamed the dependency and are updating the lock file. I can't think of any use case where you'd want to point at an existing but invalid dependency, so in those cases I think failing fast is the best option.

adriangb:
So this works. Conceptually though I don't like the idea of having `.full_path` be the "validation". From the perspective of the `poetry` codebase it seems like it might be a bit confusing, even calling it something like `get_full_path_and_validate_dependency() -> Path` would be better to make it clear that this method gets called for validation. But that's more your call, you're more familiar with the coupling between the codebases.

radoering:
You have got a point there. On the one hand it's an easier transition from "validation in `__init__`" and you can't forget to validate, on the other hand it might be a bit surprising. We could also make the `validate` method "public" and require to call it explicitly. Some more eyes might be needed here. I'll discuss this with other maintainers.

adriangb:
Sounds good. Do you think we could move forward with what we've got in the meantime? I think what we've got here is better than what we had before and the behavior ("automatic" validation) is largely the same, so we could treat making `validate` public and somehow requiring it to be explicitly called a future refactoring. Mostly because I think there needs to be some thought put into what this means for other dependency types (VCS, direct, etc.) and where it should be called in `poetry`; I don't want that to stall this work.

radoering:
I took a deeper look into this topic and believe to have found a cleaner solution in #520. Please take a look if this approach fits your needs.

adriangb:
Yep should work

PR #508, opened 2022-10-22
strict flake8-type-checking

dimbleby:
recent flake8-type-checking encourages `cast()` using the name of the class rather than the actual class, presumably with the aim that you might find more things you don't need to import at runtime

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -6,7 +6,7 @@ ban-relative-imports = True
 format-greedy = 1
 inline-quotes = double
 enable-extensions = TC, TC1
-strict-type-checking = true
+type-checking-strict = true
 eradicate-whitelist-extend = ^-.*;
 extend-ignore =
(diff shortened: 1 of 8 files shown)


radoering:
However, it seems it doesn't notice that some imports can be moved into a `TYPE_CHECKING` block (e.g. `VCSDependency` in `test_factory.py`).

radoering:
Seems we have to take care of that ourselves: snok/flake8-type-checking#127

.flake8
```diff
@@ -6,7 +6,7 @@ ban-relative-imports = True
 format-greedy = 1
 inline-quotes = double
 enable-extensions = TC, TC1
-type-checking-exempt-modules = typing, typing-extensions
+strict-type-checking = true
```
radoering, line 9:
```suggestion
type-checking-strict = true
```
https://github.com/snok/flake8-type-checking#strict ?

radoering (approved):

PR #489, opened 2022-10-03
Refresh and automate updating licenses list

Secrus:
Resolves: python-poetry/poetry#6689

- [ ] Added **tests** for changed code. (no significant changes)
- [ ] Updated **documentation** for changed code. (internal changes)

PR #511, opened 2022-10-28
Add support for the unlicense

SurenNihalani:
Resolves: python-poetry#6689


mkniewallner:
#489 is in the process of automating the update, which will regularly refresh the list and include this specific license, so IMO it's better to wait for it to be merged rather than adding licenses manually.

SurenNihalani:
fair

PR #512, opened 2022-10-30
Use file:// urls for path and directory dependencies

dimbleby:
fixes https://github.com/python-poetry/poetry/issues/5273

### DIFF
--- a/src/poetry/core/packages/directory_dependency.py
+++ b/src/poetry/core/packages/directory_dependency.py
@@ -89,9 +89,7 @@ class DirectoryDependency(Dependency):
             extras = ",".join(sorted(self.extras))
             requirement += f"[{extras}]"
 
-        path = (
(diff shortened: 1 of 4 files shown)


radoering:
I wonder if this will break some "poetry-only" use cases where someone has a couple of projects relative to each other and everything works fine when the entire folder structure is moved.

Nevertheless, I think this fix is necessary because we are creating invalid PEP 508 names at the moment.

radoering (approved):

gregtesseract:
@radoering Could you please make a new release. 
Really keen on using this change.

neersighted:
We are working on a new release (and it's linked here); as this will have to be coordinated with a Poetry release (and you, as a user of the `poetry` frontend won't benefit until it's shipped by Poetry), you can follow progress there.

PR #513, opened 2022-10-30
constraints: make EmptyConstraint hashable and tidy up a bit

radoering:
follow up of #466 which handled generic constraints, this one handles version constraints


neersighted (approved):
LGTM overall; however I'm not a huge fan of BaseConstaint, that's neither here nor there for this PR. How painful would it be to refactor it into an ABC in the future?

radoering:
> How painful would it be to refactor it into an ABC in the future?

It's a bit of work (at least if we want to do it properly) because not all (concrete) sub classes implement all methods that raise `NotImplementedError` in `BaseConstraint`. I took the first step with `intersection` and `union`: #515

PR #514, opened 2022-10-31
clearer error handling

dimbleby:
- raise and catch more precise errors
- version-parsing errors on a package are caught and re-raised with text saying what the package is
  - fixes https://github.com/python-poetry/poetry/issues/4597

### DIFF
--- a/src/poetry/core/constraints/version/parser.py
+++ b/src/poetry/core/constraints/version/parser.py
@@ -4,6 +4,9 @@ import re
 
 from typing import TYPE_CHECKING
 
+from poetry.core.constraints.version.exceptions import ParseConstraintError
+from poetry.core.version.exceptions import InvalidVersion
+
 
 if TYPE_CHECKING:
     from poetry.core.constraints.version.version_constraint import VersionConstraint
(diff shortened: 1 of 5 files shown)


src/poetry/core/constraints/version/parser.py
```diff
@@ -4,6 +4,9 @@
 
 from typing import TYPE_CHECKING
 
+from poetry.core.constraints.version.exceptions import ParseConstraintError
```
radoering, line 7: The local import in line 145 can be removed.

src/poetry/core/packages/package.py
```diff
@@ -215,11 +217,13 @@ def _set_version(
...
         if not isinstance(version, Version):
-            self._version = Version.parse(version)
-            self._pretty_version = pretty_version or version
-        else:
-            self._version = version
-            self._pretty_version = pretty_version or self._version.text
+            try:
+                version = Version.parse(version)
+            except InvalidVersion:
+                raise ValueError(f"Invalid version '{version}' on package {self.name}")
```
radoering, line 223: Why a `ValueError` and not a more specific exception like `InvalidVersion` with an updated error message?

src/poetry/core/packages/package.py
```diff
@@ -261,8 +265,13 @@ def python_versions(self) -> str:
 
     @python_versions.setter
     def python_versions(self, value: str) -> None:
+        try:
+            constraint = parse_constraint(value)
+        except ParseConstraintError:
+            raise ValueError(f"Invalid python versions '{value}' on {self}")
```
radoering, line 271: Same as above.

src/poetry/core/packages/package.py
dimbleby: can't say I feel very strongly about it

radoering (approved):

PR #520, opened 2022-11-12
Allow non-existent path dependencies - reloaded

radoering:
Resolves: python-poetry/poetry#668
Resolves: python-poetry/poetry#6828

Supersedes: #210, #507

Requires: python-poetry/poetry#6844 (to preserve comprehensible output in case of error)

Note: Commits should not be squashed.

### DIFF
--- a/src/poetry/core/packages/path_dependency.py
+++ b/src/poetry/core/packages/path_dependency.py
@@ -45,6 +45,8 @@ class PathDependency(Dependency, ABC):
             source_url=self._full_path.as_posix(),
             extras=extras,
         )
+        # cache validation result to avoid unnecessary file system access
+        self._validation_error = self._validate()
         self.validate(raise_error=False)
 
     @property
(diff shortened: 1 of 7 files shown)


src/poetry/core/packages/path_dependency.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+class PathDependency(Dependency, ABC):
+    @abstractmethod
+    def __init__(
+        self,
+        name: str,
+        path: Path,
+        *,
+        source_type: str,
```
Secrus, line 24: how about `Literal['file'] | Literal['directory']` type or Enum?

Secrus (changes requested):

src/poetry/core/packages/path_dependency.py
radoering: I'd like to use `Literal` but it's new in 3.8. Using Enum would probably entail a larger refactoring since we're using strings for all types of dependencies. I think that's out of scope for this PR. And introducing an Enum only for path dependencies is not worth it, at least in my opinion.

neersighted (approved):

adriangb:
Awesome glad to see this in!

dndn10:
Hi, is this feature supposed to allow the usage of poetry install without having the actual path dependencies in the filesystem?

PR #523, opened 2022-11-15
feat: add possibility to set the pyproject TOML on a Poetry instance

DavidVujic:
Resolves part one of: [python-poetry#7039](https://github.com/python-poetry/poetry/issues/7039) 

### DIFF
--- a/src/poetry/core/poetry.py
+++ b/src/poetry/core/poetry.py
@@ -21,13 +21,17 @@ class Poetry:
     ) -> None:
         from poetry.core.pyproject.toml import PyProjectTOML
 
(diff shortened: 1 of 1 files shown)


neersighted:
I think this is an inferior duplicate of #6810.

DavidVujic:
> I think this is an inferior duplicate of [python-poetry/poetry#6810](https://github.com/python-poetry/poetry/pull/6810) (it's not necessary to add special support for overriding the project file as there are other places in the code that assume its location at the root of the project, and we can set CWD instead as in that PR).

Thank you for explaining and linking to the already ongoing work. I'll close this one!

PR #527, opened 2022-11-24
Allow optional output directory for building

AdamWRichardson:
This PR makes it possbile for the builder factory class to accept an optional output directory when building packages. This builder factory is used by the command line and if this is accepted I would like to add an extra option to the command line to allow placing build artifacts in a directory other than `dist`.

Of course I am happy to take comments and feedback.

Relates-to: python-poetry/poetry#3586


src/poetry/core/masonry/builder.py
```diff
@@ -21,7 +21,12 @@ def __init__(self, poetry: Poetry) -> None:
             "wheel": WheelBuilder,
         }
 
-    def build(self, fmt: str, executable: str | Path | None = None) -> None:
+    def build(
+        self,
+        fmt: str,
+        output_directory: Path | None = None,
+        executable: str | Path | None = None,
```
radoering, line 28:
```suggestion
        *,
        output_directory: Path | None = None,
        executable: str | Path | None = None,
```

Just some quick fly-by feedback: If we are sure that we can add the new parameter between `fmt` and `executable` without breaking anything (I've not checked if that is true.), we should take the chance and make these keyword-only arguments.

finswimmer: I would prefer to add the new parameter at the end. Thus we can makes sure we will not break anything. Furthermore I would prefer naming it `target_dir`, just like in the `build()` method in `Builder`.

AdamWRichardson: I'm also happy to put the output directory after executible if that would be better?

radoering: I'm fine with that. It's the safe option. Maybe, we can make at least the new option keyword-only.

AdamWRichardson: Recent change places `target_dir` at the end and makes it keyword-only

tests/masonry/builders/test_builder_factory.py
```diff
@@ -0,0 +1,67 @@
...
+        Builder(get_poetry("complete")).build("not_valid")
+
+
+@pytest.mark.parametrize("format", ["sdist", "wheel", "all"])
+def test_builder_creates_places_built_files_in_specified_directory(
+    tmp_path: Path, format: str
+) -> None:
+    poetry = get_poetry("complete")
+    Builder(poetry).build(format, target_dir=tmp_path)
+    assert all(archive.exists() for archive in tmp_path.glob(get_package_glob(poetry)))
```
radoering, line 52: This test passes even without the change (i.e., if the directory is empty).

tests/masonry/builders/test_builder_factory.py
```diff
@@ -0,0 +1,67 @@
...
+@pytest.mark.parametrize("format", ["sdist", "wheel", "all"])
+def test_builder_creates_packages_in_dist_directory_if_no_output_is_specified(
+    format: str,
+) -> None:
+    with get_project_context("complete") as project:
+        poetry = Factory().create_poetry(project)
+        Builder(poetry).build(format, target_dir=None)
+        package_directory = project / "dist"
+        assert package_directory.is_dir()
+        assert all(
```
radoering, line 64: Same issue. This check will pass even if the directory is empty.

radoering (changes requested):
I'd prefer to mimic the src folder structure and put the test into tests/masonry/test_builder.py.

tests/masonry/builders/test_builder_factory.py
AdamWRichardson: Good catch, thanks! I've changed the iterator to a tuple and added an assert on the length of the tuple

radoering (approved):

PR #529, opened 2022-11-27
markers: improve recognition of empty marker in a complex intersections

radoering:
Resolves: python-poetry/poetry-plugin-export#163
Closes: #528

Alternative approach to #528. Includes some minor cleanup to make `MultiMarker.of` and `MarkerUnion.of` more symmetrical (as far as that is possible). The relevant change is the handling of `MultiMarker` in `_of_marker_union_and_single_marker`.

### DIFF
--- a/src/poetry/core/version/markers.py
+++ b/src/poetry/core/version/markers.py
@@ -396,14 +396,15 @@ class MultiMarker(BaseMarker):
             old_markers = new_markers
             new_markers = []
             for marker in old_markers:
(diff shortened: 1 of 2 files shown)


dimbleby:
I added a testcase for an even-more-complex intersection in my MR #528, I leave it with you to consider whether you want to continue down this path to try and make that one work too

radoering:
Probably not worth it since DNF is a more general approach.

PR #530, opened 2022-11-27
Simplify marker simplification

dimbleby:
per the discussion from https://github.com/python-poetry/poetry-core/pull/528#issuecomment-1328237347, try to use reduction to DNF and CNF as much as possible to achieve marker simplifications, in the hope that:

- it's more general than the collection of heuristics we currently have
- it simplifies the code

I think this probably is an improvement - and the test coverage is pretty decent here, so there's reason to think that it doesn't break things too badly.  But more opinion is certainly wanted.

### DIFF
--- a/src/poetry/core/version/markers.py
+++ b/src/poetry/core/version/markers.py
@@ -702,21 +702,28 @@ def parse_marker(marker: str) -> BaseMarker:
     return markers
 
 
-def _compact_markers(tree_elements: Tree, tree_prefix: str = "") -> BaseMarker:
+def _compact_markers(
+    tree_elements: Tree, tree_prefix: str = "", top_level: bool = True
+) -> BaseMarker:
     from lark import Token
 
-    groups: list[BaseMarker] = [MultiMarker()]
+    # groups is a disjunction of conjunctions
+    # eg [[A, B], [C, D]] represents "(A and B) or (C and D)"
+    groups: list[list[BaseMarker]] = [[]]
+
     for token in tree_elements:
         if isinstance(token, Token):
             if token.type == f"{tree_prefix}BOOL_OP" and token.value == "or":
-                groups.append(MultiMarker())
+                groups.append([])
 
             continue
 
         if token.data == "marker":
-            groups[-1] = MultiMarker.of(
-                groups[-1], _compact_markers(token.children, tree_prefix=tree_prefix)
+            sub_marker = _compact_markers(
+                token.children, tree_prefix=tree_prefix, top_level=False
             )
+            groups[-1].append(sub_marker)
+
         elif token.data == f"{tree_prefix}item":
             name, op, value = token.children
             if value.type == f"{tree_prefix}MARKER_NAME":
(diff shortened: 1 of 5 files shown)


src/poetry/core/version/markers.py
```diff
@@ -443,34 +430,20 @@ def of(cls, *markers: BaseMarker) -> BaseMarker:
...
+        return dnf(multi)
 
+    def union(self, other: BaseMarker) -> BaseMarker:
         if isinstance(other, MarkerUnion):
-            return other.intersect(self)
-
-        new_markers = self._markers + [other]
-
-        multi = MultiMarker.of(*new_markers)
+            return other.union(self)
```
radoering, line 439: Can we drop this? It seems it just changes nothing but the order (and IMO without it, it's even better).

src/poetry/core/version/markers.py
```diff
@@ -682,39 +635,20 @@ def append(self, marker: BaseMarker) -> None:
...
-                    new_markers.append(intersection)
-        elif isinstance(other, MarkerUnion):
-            for our_marker in self._markers:
-                for their_marker in other.markers:
-                    intersection = our_marker.intersect(their_marker)
-
-                    if not intersection.is_empty():
-                        new_markers.append(intersection)
+        if isinstance(other, MultiMarker):
+            return other.intersect(self)
```
radoering, line 639: Same here. It seems it just changes nothing but the order (and IMO without it, it's even better).

src/poetry/core/version/markers.py
dimbleby: doing so results in union being not-quite-symmetrical, in a way that the unit tests notice.  Similar in the other one.

Perhaps bowing to the unit tests is allowing the tail to wag the dog though

tests/packages/utils/test_utils.py
```diff
@@ -41,7 +41,10 @@
                 ' "win32" and python_version < "3.6" and python_version >= "3.3" or'
                 ' sys_platform == "win32" and python_version < "3.3"'
             ),
-            {"python_version": [[("<", "3.6")]], "sys_platform": [[("==", "win32")]]},
+            {
+                "python_version": [[("<", "3.6")], [("<", "3.3")]],
+                "sys_platform": [[("==", "win32")]],
+            },
```
radoering, line 47: That looks like a regression, doesn't it?

dimbleby: I don't think it's a problem - per some of the other test cases here, `convert_markers` isn't too precise anyway.

Actually what this testcase reveals is a change in behaviour at `parse_marker()` which gets _some_ of the available simplifications by using the `.of()` methods - but misses out on the simplifications achieved by `cnf()` and `dnf()` in `union()` and `intersect()`.

I have another simplification in my pocket that reworks `_compact_markers()` in terms of `intersect()` and `union()`.   That results in parsed markers being as simplified as we can manage.

It's a little more disruptive to the unit test but I think I'm good with it, will push as a separate commit shortly - see what you think.

dimbleby:
Reckon I'm done fiddling with this now, I'm pretty happy with the much-reduced set of heuristic marker simplifications being largely replaced by more systematic `cnf()` / `dnf()`

remaining downstream failure is a benign rearrangement of a parsed marker per the final commit, if y'll are minded to merge this it's easy enough to fix up

radoering:
I added a general `intersection` and `union` function as proposed in https://github.com/python-poetry/poetry/pull/7173#discussion_r1045229028 and moved some duplicated code in there. What do you think about it?

dimbleby:
Fine by me.

I considered suggesting moving these to staticmethods on the `BaseMarker` so that callers would call eg `BaseMarker.union(*markers)`.  If the `BaseMarker` were ever so slightly less awkwardly named - eg just `Marker` would do - then I might favour that.  But as it stands, I judge that would make calling code more ugly, rather than less ugly.

(And I don't think it's worth renaming the `BaseMarker` just for this)

radoering (approved):

PR #535, opened 2022-11-30
Deprecate hash function from FileDependency

dunkmann00:
This will enable URLDependency to compute the hash for the file at the temporary path. The temporary path should be the path to the downloaded file from `URLDependency.url`.

Relates-to: python-poetry/poetry#7121

### DIFF
--- a/src/poetry/core/packages/file_dependency.py
+++ b/src/poetry/core/packages/file_dependency.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import hashlib
 import io
+import warnings
 
 from pathlib import Path
 from typing import Iterable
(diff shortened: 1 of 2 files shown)


dimbleby:
Given the way that https://github.com/python-poetry/poetry/pull/7121 is going, this want to become an MR that deprecates or removes the hash() method on the file dependency

dunkmann00:
I removed the `hash` function from `URLDependency` and the tests that went with it, so I think this should be good pending https://github.com/python-poetry/poetry/pull/7121.

radoering:
Please don't remove but deprecate it. See https://github.com/python-poetry/poetry-core/blob/51e625d8cc6d4793f3f88081e1da42cdffda9414/src/poetry/core/utils/helpers.py#L28-L34 for an example.

In the tests, please add a `with pytest.warns(DeprecationWarning):`

dunkmann00:
Okay, added it back with the deprecation warnings.

radoering (approved):

PR #537, opened 2022-12-04
respect metadata_directory at editable builds

dimbleby:
Missed editable builds at #487

### DIFF
--- a/tests/masonry/test_api.py
+++ b/tests/masonry/test_api.py
@@ -254,3 +254,30 @@ def test_build_wheel_with_metadata_directory() -> None:
                 path=str(os.path.join(wheel_tmp_dir, filename)),
                 files=["entry_points.txt"],
             )
+
+
+def test_build_editable_wheel_with_metadata_directory() -> None:
+    pkg_dir = Path(fixtures) / "complete"
+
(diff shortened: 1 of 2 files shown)


tests/masonry/test_api.py
```diff
@@ -254,3 +254,30 @@ def test_build_wheel_with_metadata_directory() -> None:
                 path=str(os.path.join(wheel_tmp_dir, filename)),
                 files=["entry_points.txt"],
             )
+
+
+def test_build_editable_wheel_with_metadata_directory() -> None:
```
radoering, line 259: It seems, the test succeeds even without the change? The same is true for the original test (for non-editable wheels) if I just do not pass a metadata directory. Can we improve the tests so that they fail if the metadata directory is not respected?

dimbleby: I dunno, the wheels are after all _supposed_ to be identical either way.

I don't much care about this.  The current MR is supposed to be an obvious fix to the omission last time, I don't know that I'm likely to find the enthusiasm to write new testcases.  I'm fine with you either merging it, or not.

dimbleby: I'm not saying it's difficult, I'm saying that I don't really care: and might never bother to do anything about it

PR #538, opened 2022-12-10
update vendored dependencies

dimbleby:
Especially packaging, per https://github.com/python-poetry/poetry/issues/7161

reviewers should note that the patch that is applied to jsonschema (so as to avoid having to pull in importlib-resources) became ever so slightly more complicated, as jsonschema shuffled some code around


dimbleby:
turns out that this doesn't have anything much to do with that issue over in poetry after all.

Might just as well merge it anyway, keeping up to date is good, but this pull request certainly doesn't need to be released in any sort of hurry.

radoering (approved):

PR #542, opened 2022-12-22
validate extras against dependencies and in schema

Panaetius:
Resolves: python-poetry/poetry/issues/7226

With this change, `extras` are validated to only contain valid characters and extras that reference dependencies that can't be found in the main dependency group raise a warning in `poetry check` (not in `poetry lock`/`poetry install`, though)


src/poetry/core/factory.py
```diff
@@ -409,6 +409,20 @@ def validate(
                             'Use "allow-prereleases" instead.'
                         )
 
+                if "extras" in config:
+                    for extra_name, requirements in config["extras"].items():
+                        extra_name = canonicalize_name(extra_name)
+
+                        for req in requirements:
+                            for dependency in config["dependencies"].keys():
+                                if req == dependency:
```
radoering, line 418: I think it's necessary to `canonicalize_name()` requirements as well as dependencies.

Panaetius:
I've addressed the PR comments.

@radoering do you have any thoughts on whether this being a warning in `poetry check` as per this PR is fine or should it be added to `check_result["errors"]` instead?

radoering:
Good point. Comparing it to the other warnings and errors, I think it should be an error.

It's just an edge case but I just noticed extras should be checked if there are no dependencies. Can you unindent the new block by one level and replace `config["dependencies"]` by `config.get("dependencies", {})`?

radoering (approved):

lovetheguitar:
@radoering FYI, this was a breaking change for some. 

E.g. isort 5.10.1 in our .pre-commit-config.yaml failed to install everywhere because of the 1.5 release.

https://github.com/PyCQA/isort/blob/9e4b5ee05264e6e566b18bea88b3073abd4b1014/pyproject.toml#L51

I guess version pinning of extras is not allowed/correct there (anymore)?

radoering:
> I guess version pinning of extras is not allowed/correct there (anymore)?

It's still possible to define constraints for extras the same way as before: By defining an optional dependency as explained in the [documentation](https://python-poetry.org/docs/pyproject#extras) and now done by isort in PyCQA/isort#2078. You can verify that the latest isort release has the same constraint on pip-shims as before:

```
$ curl -s https://pypi.org/pypi/isort/5.12.0/json | jq .info.requires_dist | grep pip-shims
  "pip-shims (>=0.5.2) ; extra == \"pipfile-deprecated-finder\"",
```

To the best of my knowledge, the way isort specified the extra before is not documented and was never officially supported by poetry. We were not aware that this was possible and even resulted in valid metadata.

PR #543, opened 2022-12-22
PEP 440 compliance: do not implicitly allow pre-releases

radoering:
Extracted from #402. Besides being a precondition of #402 it's a fix on its own. And it requires some coordination with downstream.

Requires: python-poetry/poetry#7225 and python-poetry/poetry#7236

PEP 440 says:

> Pre-releases of any kind, including developmental releases, are implicitly excluded from all version specifiers, unless they are already present on the system, explicitly requested by the user, or if the only available version that satisfies the version specifier is a pre-release.

Nothing special about "`>` _pre-release_" or "`>=` _pre-release_"

### DIFF
--- a/.github/workflows/downstream.yml
+++ b/.github/workflows/downstream.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        ref: ["master", "1.3"]
+        ref: ["master"]
       fail-fast: false
     defaults:
       run:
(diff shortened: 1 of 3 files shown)


neersighted (changes requested):
LGTM with a negative test (discussed in maintainer's call).

radoering:
Adapted existing test instead of removing it.

neersighted (approved):

PR #545, opened 2022-12-23
builders/wheel: Ensure dist-info is written deterministically

rpurdie:
glob() returns values in "on disk" order. To make the RECORD file deterministic and consistent between builds we need to sort the data before adding to the records list.

Signed-off-by: Richard Purdie <richard.purdie@linuxfoundation.org>

Resolves: python-poetry#

### DIFF
--- a/src/poetry/core/masonry/builders/wheel.py
+++ b/src/poetry/core/masonry/builders/wheel.py
@@ -294,7 +294,7 @@ class WheelBuilder(Builder):
 
     def _copy_dist_info(self, wheel: zipfile.ZipFile, source: Path) -> None:
         dist_info = Path(self.dist_info)
-        for file in source.glob("**/*"):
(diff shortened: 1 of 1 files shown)


radoering:
Interesting. I suppose, we should sort here as well since `_add_file()` is called in the loop?

https://github.com/python-poetry/poetry-core/blob/87186813480c9e120bf65cebe7bf8dd69aabc82c/src/poetry/core/masonry/builders/wheel.py#L193

rpurdie:
> Interesting. I suppose, we should sort here as well since `_add_file()` is called in the loop?
> 
> https://github.com/python-poetry/poetry-core/blob/87186813480c9e120bf65cebe7bf8dd69aabc82c/src/poetry/core/masonry/builders/wheel.py#L193

Yes, that looks like it would have the same issue.

radoering (approved):

PR #549, opened 2023-01-05
version.text replaces pretty_version

dimbleby:
Companion to https://github.com/python-poetry/poetry/pull/7305

Downstream tests should fail until https://github.com/python-poetry/poetry/pull/7305 propagates

I suppose the `Package` could instead continue to accept the pretty_version in its constructor but just ignore it: while that would be backwards-compatible it would also be rather misleading.  

I don't think it's worth it myself - but if you really wanted to go through the deprecation cycle I suppose we could check for not-None pretty_version and warn

### DIFF
--- a/src/poetry/core/packages/package.py
+++ b/src/poetry/core/packages/package.py
@@ -54,7 +54,6 @@ class Package(PackageSpecification):
         self,
         name: str,
         version: str | Version,
-        pretty_version: str | None = None,
         source_type: str | None = None,
         source_url: str | None = None,
         source_reference: str | None = None,
@@ -79,7 +78,7 @@ class Package(PackageSpecification):
             features=features,
         )
 
-        self._set_version(version, pretty_version)
+        self._set_version(version)
 
         self.description = ""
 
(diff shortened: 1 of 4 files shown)


radoering:
> but if you really wanted to go through the deprecation cycle I suppose we could check for not-None pretty_version and warn

I'd prefer this route and appreciate if you can do as proposed.

src/poetry/core/packages/package.py
```diff
@@ -69,6 +70,13 @@ def __init__(
         """
         from poetry.core.version.markers import AnyMarker
 
+        if pretty_version is not None:
+            warnings.warn(
+                "The `pretty_version` parameter is deprecated and will be removed"
+                " in a future release.",
+                DeprecationWarning,
+            )
```
radoering, line 78:
```suggestion
                DeprecationWarning,
                stacklevel=2,
            )
```

It seems like stacklevel is missing in the other deprecations in this file, too.

radoering:
Looks like the deprecation revealed further usage. 😄

radoering (approved):

PR #550, opened 2023-01-08
remove special-case any/empty code on MarkerUnion

dimbleby:
`MarkerUnion` had code handling the possibility that one of its markers is `AnyMarker` or `EmptyMarker`, neither of which should ever happen.

Conceivably it makes sense to have this code anyway for defensive purposes, but in that case there ought to be similar on the `MultiMarker`.  I've instead preferred to remove it.  

The testcase that I've removed was testing that the string representation of such a union ignored the any / empty markers.  But this was nonsense: if there ever actually was an "any" marker in there, then the whole union ought to have been "any".

While passing I've also removed `is_empty()` from the `AnyMarker` and `is_any()` from the `EmptyMarker`: they too can inherit the default implementation (`return False`) from the base class.

### DIFF
--- a/src/poetry/core/version/markers.py
+++ b/src/poetry/core/version/markers.py
@@ -128,9 +128,6 @@ class AnyMarker(BaseMarker):
     def is_any(self) -> bool:
         return True
 
-    def is_empty(self) -> bool:
-        return False
-
     def validate(self, environment: dict[str, Any] | None) -> bool:
         return True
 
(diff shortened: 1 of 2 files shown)


dimbleby:
not sure I agree with sonarcloud's "improvement" here, and I certainly have no idea why this merge request provokes it into making the suggestion...

radoering (approved):

PR #554, opened 2023-02-08
Don't commit poetry.lock

dimbleby:
As https://github.com/python-poetry/poetry/issues/7488 notes, poetry's own documentation recommends not committing lockfiles for libraries.

This project should practice what it preaches.

(It could be reasonable to reject this MR on the grounds that it's in some way more complicated than that - but then the docs should acknowledge that complication)


radoering:
I'm a slight -1 for this and +1 for changing the docs because with a lock file you always know which dependencies work for sure. Without a lock file, you will get more failures unrelated to your changes. Then, you'll have to realize the failure has nothing to do with your change, search for a successful CI run, check if some dependency changed, ...

ralbertazzi:
Now that https://github.com/python-poetry/poetry/pull/7506 has been merged I think the documentation is aligned with us keeping the lock file. I would close this PR, and - why not - have a round of dependencies update as the documentation suggests :)

radoering:
I'll close this for now since no maintainer has expressed agreement so far and the documentation no longer recommends not to commit the lockfile.

PR #555, opened 2023-02-12
Prefer pathlib

dimbleby:
I've been sitting for a while on some tweaks to move poetry-core in the direction of preferring `pathlib` over `os` for file and path manipulations, but they've not seemed interesting enough to be worth an MR.

However I notice that they in passing should fix https://github.com/python-poetry/poetry/issues/7505, which makes it all seem a bit more worthwhile.

This first push deliberately doesn't include the fixes - I'm using the CI pipeline to do my windows testing and am hoping to see that the testcase that I've added fails, per that issue.  Then I'll follow up with the changes that are supposed to fix it.

### DIFF
--- a/src/poetry/core/json/__init__.py
+++ b/src/poetry/core/json/__init__.py
@@ -1,12 +1,12 @@
 from __future__ import annotations
 
 import json
-import os
 
+from pathlib import Path
 from typing import Any
 
 
-SCHEMA_DIR = os.path.join(os.path.dirname(__file__), "schemas")
+SCHEMA_DIR = Path(__file__).parent / "schemas"
 
 
 class ValidationError(ValueError):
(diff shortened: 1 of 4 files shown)


gpongelli:
Hi,
I've tested this branch locally and I confirm this fixes https://github.com/python-poetry/poetry/issues/7505 .

Steps I've done:
1. clone locally [dimbleby's prefer-pathlib branch](https://github.com/dimbleby/poetry-core/tree/prefer-pathlib)
2. clone locally [poetry ](https://github.com/python-poetry/poetry) and checkout tag 1.4.0
3. change poetry's pyproject.toml to reference local folder , run `poetry lock` to obtain

```
[[package]]
name = "poetry-core"
version = "1.5.0"
...
[package.source]
type = "directory"
url = "../poetry-core"
```
4. run `pipx.exe install --verbose --suffix=@loc .\poetry\`
5. into the native extension project, I've run `poetry@loc.exe build`

Result:
- no `UserWarning: Duplicate name: warning`
- RECORD file does **not** have duplicated hashes


I'm on Windows 11 OS, powershell environment.

Please consider merging this branch ASAP for next poetry release, thank you.

radoering:
I think we can include this in a patch release (and bump the required core version in the next poetry patch release). However, I don't want to remove any deprecation in a patch release (even if it is unlikely anybody still uses it). Do you mind if I drop the first commit in this PR? (You're welcome to create a deprecation removal PR for the next minor release if you like. There might be some more things that are deprecated for a long time.)

dimbleby:
feel free to rearrange as you wish

(per opening remarks, I'd intended this series as more of a tidying-up sequence of commits rather actually-fixes-something; it's mostly luck that it turns out to fix something that someone cares about)

radoering (approved):

PR #556, opened 2023-02-19
performance: cache parsed markers, constraints and versions

radoering:
Some more fallout from #530. The parsing of marker strings became a bit slower with measurable effects in real-world examples. It seems in real-world examples, the same marker is used many times for different dependencies. Since markers are immutable, we can just cache the parsed markers and mitigate the performance regression.

Performance for shootout example with warm cache:

|command|1.3.2|master|PR (only markers)|PR (constraints, versions)|
|---|---|---|---|---|
|lock|46 s|57 s|47 s|35 s|
|lock --no-update|14 s|18 s|14 s|10 s|

(There seems to be no significant increase in peak memory usage due to the caching.)

Update: added times for caching parsed constraints and versions in addition to markers.

### DIFF
--- a/src/poetry/core/version/markers.py
+++ b/src/poetry/core/version/markers.py
@@ -1,8 +1,11 @@
 from __future__ import annotations
 
+import functools
 import itertools
 import re
 
+from abc import ABC
+from abc import abstractmethod
 from typing import TYPE_CHECKING
 from typing import Any
 from typing import Callable
@@ -53,10 +56,12 @@ PYTHON_VERSION_MARKERS = {"python_version", "python_full_version"}
 _parser = Parser(GRAMMAR_PEP_508_MARKERS, "lalr")
 
 
-class BaseMarker:
+class BaseMarker(ABC):
+    @abstractmethod
     def intersect(self, other: BaseMarker) -> BaseMarker:
         raise NotImplementedError()
 
+    @abstractmethod
     def union(self, other: BaseMarker) -> BaseMarker:
         raise NotImplementedError()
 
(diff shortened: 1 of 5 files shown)


src/poetry/core/version/markers.py
```diff
@@ -725,6 +743,7 @@ def is_empty(self) -> bool:
         return all(m.is_empty() for m in self._markers)
 
 
+@functools.lru_cache(maxsize=None)
```
wagnerluis1982, line 746: How many markers do we have for an average installation?

Only thinking if wouldn't make sense to set a boundary to the cache here.

dimbleby (approved):

dimbleby:
looks fine to me though fwiw I don't reproduce any performance delta on the "big" install that I have lying around.

dimbleby:
if this is real: I wonder whether constraint parsing and version parsing would similarly benefit

src/poetry/core/version/markers.py
radoering: > How many markers do we have for an average installation?

Interesting question, but difficult to say. I tested some projects, but it's subjective if they are typical.

```
CacheInfo(hits=16070, misses=25, maxsize=None, currsize=25)
CacheInfo(hits=6369, misses=36, maxsize=None, currsize=36)
CacheInfo(hits=55, misses=14, maxsize=None, currsize=14)
CacheInfo(hits=289, misses=25, maxsize=None, currsize=25)
CacheInfo(hits=17518, misses=48, maxsize=None, currsize=48)
CacheInfo(hits=1124, misses=45, maxsize=None, currsize=45)
CacheInfo(hits=67168, misses=60, maxsize=None, currsize=60)
CacheInfo(hits=66985, misses=78, maxsize=None, currsize=78)
```

`currsize` is the number of unique markers. `hits + misses` is the number of all parsed markers.

> Only thinking if wouldn't make sense to set a boundary to the cache here.

I don't think it's necessary. On further testing I could even observe a reduction of peak memory usage. I suppose this is because now we only create one object per unique marker, before we created several ones (and probably held a lot of them until the end).

radoering:
> if this is real: I wonder whether constraint parsing and version parsing would similarly benefit

Great idea. Although there is more variance in constraints and versions than in markers, it makes a significant difference, too.

I updated the times in the PR description. Further, I tested the change on another largish (non-public) project:

|command|without PR|only markers|markers,constraints,versions|
|---|---|---|---|
|poetry lock --no-update|150 s|135 s|115 s|

Cache info for private project:
```
CacheInfo(hits=66985, misses=78, maxsize=None, currsize=78) poetry.core.version.markers.parse_marker
CacheInfo(hits=452083, misses=739, maxsize=None, currsize=739) poetry.core.constraints.version.parser.parse_constraint
CacheInfo(hits=73960, misses=196, maxsize=None, currsize=196) poetry.core.constraints.generic.parser.parse_constraint
CacheInfo(hits=19933, misses=581, maxsize=None, currsize=580) poetry.core.version.pep440.parser.PEP440Parser.parse
```

The peak memory usage even decreased when caching constraints and versions.

neersighted (approved):

PR #559, opened 2023-03-05
Fix the PR template to match the main branch name

PeterJCLaw:


radoering (changes requested):
You can remove the part about the develop branch, too. Pull requests should always be based on the main branch.

PeterJCLaw:
That feels a little out of scope for this change?

(Also, as a first-time and external contributor, it's not clear to me that I could have known that that portion of the template was inaccurate)

wagnerluis1982:
I think that even `develop` branch information is also inaccurate, as there is no such a branch any longer.

radoering:
@PeterJCLaw I didn't mean that you could have known that part. I'm sorry if my comment was misleading. I just wanted to say that we should fix this part of the note as well if we edit the note anyway.

As for the scope of this PR, I don't think that we need two PRs to fix one note in the docs.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -6,8 +6,7 @@ Resolves: python-poetry#<!-- add issue number/link here -->
 - [ ] Updated **documentation** for changed code.
 
 <!--
-**Note**: If your Pull Request introduces a new feature or changes the current behavior, it should be based
-on the `develop` branch. If it's a bug fix or only a documentation update, it should be based on the `master` branch.
+**Note**: All Pull Request should be based on the `main` branch.
```
wagnerluis1982, line 9: Better to use the plural form.

I also think that using **must** put more emphasis

```suggestion
**Note**: All Pull Requests must be based on the `main` branch.
```

radoering: I tend to "should" but no strong feelings about that.

+1 for plural

radoering (approved):

PR #565, opened 2023-03-14
More specific error message when project name contains disallowed characters

MolarFox:
## Context
Recently I was working on a project using poetry, and ended up in a state where I was unable to use `poetry build`, due to getting a "`No file/folder found for package <project_name>`" error repeatedly, despite the folder structure etc being correct to what poetry expects to see

After a bunch of google searching and diving into the source code myself, I was able to see that the cause was that the project name I was using was "molpro-dirman", containing an illegal `-` character. Once I sussed this out and [changed all the project naming](https://github.com/MolarFox-Prototyping/molpro-dirman/commit/d5247f16fa54174a6c084c797f1e6e26f94391e8) to use an underscore instead, everything worked nominally.

In this change I've added an error trap and more specific error message when this case arises for the next person :slightly_smiling_face: 

## Changes
- Catch specific subcase where dir not found due to substitution of disallowed chars
    - Log more specific error message in this case

## Checklist
- [ ] Added **tests** for changed code. (n/a / minor error trapping code change only)
- [ ] Updated **documentation** for changed code. (n/a / minor change / self-documenting changes)

### DIFF
--- a/src/poetry/core/masonry/utils/module.py
+++ b/src/poetry/core/masonry/utils/module.py
@@ -66,9 +66,9 @@ class Module:
                         }
                     ]
                 else:
-                    raise ModuleOrPackageNotFound(
-                        f"No file/folder found for package {name}"
-                    )
+                    if self._name != name:
+                        raise ModuleOrPackageNotFound(f"Project file/folder contains disallowed characters - expected naming: {name}")
+                    raise ModuleOrPackageNotFound(f"No file/folder found for package {name}")
 
         for package in packages:
             formats = package.get("format")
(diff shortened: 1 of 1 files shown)


dimbleby:
You've misunderstood something.  Not only are hyphens allowed in project names, this very project has a hyphen in its project name: https://github.com/python-poetry/poetry-core/blob/e33e9d1613338f045b55cae1a7953dec7d86f311/pyproject.toml#L2

Also your pull request has nothing much to do with what you've reported: a hyphen is not an uppercase underscore!

dimbleby:
also the code that you've added should be unhittable: `self._name` has already been normalized by a call to `module_name()`

Perhaps you could usefully include `module._name` in the existing exception

PR #569, opened 2023-03-28
raise InvalidMarker for invalid marker

dimbleby:
fixies https://github.com/python-poetry/poetry/issues/7729

the bonus testcases are just some things I encountered while making a mess of reproducing the original

### DIFF
--- a/src/poetry/core/version/markers.py
+++ b/src/poetry/core/version/markers.py
@@ -319,7 +319,7 @@ class SingleMarker(SingleMarkerLike[Union[BaseConstraint, VersionConstraint]]):
         # Extract operator and value
         m = self._CONSTRAINT_RE.match(constraint_string)
         if m is None:
-            raise InvalidMarker(f"Invalid marker '{constraint_string}'")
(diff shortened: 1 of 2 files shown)


dimbleby:
haha, pre-commit insists on making the testcase uglier.  Whatever.

src/poetry/core/version/markers.py
```diff
@@ -357,7 +356,10 @@ def __init__(
                 values = re.split("[ ,]+", self._value)
                 constraint_string = glue.join(f"{op} {value}" for value in values)
 
+        try:
             parsed_constraint = parser(constraint_string)
+        except ParseConstraintError as e:
+            raise InvalidMarker(f"Invalid marker: {original_constraint_string}") from e
```
radoering, line 362: Since it's only the constraint maybe as follows?

```suggestion
            raise InvalidMarker(f"Invalid marker constraint for "{name}": {original_constraint_string}") from e
```

radoering (approved):

PR #571, opened 2023-04-16
python 3.12 in CI

dimbleby:
sister to https://github.com/python-poetry/poetry/pull/7803, let's see what's broken


dimbleby:
didn't really expect that to pass, but I had forgotten that the "downstream" tests are python 3.10 only so I guess that makes sense

neersighted (approved):

PR #573, opened 2023-04-21
Remove tox

Secrus:
Resolves: python-poetry/poetry#7826
Resolves: python-poetry/poetry#7824

There is no real reason to keep tox, so let's remove it.


dimbleby:
tox was previously installing some dependencies that are now not being installed by any means, I expect you'll need to mimic `tests.yml` and bootstrap poetry, poetry install, etc

alternatively I'm not sure why it's useful for these tests to be separate from all the others, maybe just tag on a `pytest --integration tests/integration` at the end of `tests.yml` and delete `integration.yml` altogether.

radoering:
> alternatively I'm not sure why it's useful for these tests to be separate from all the others, maybe just tag on a `pytest --integration tests/integration` at the end of `tests.yml` and delete `integration.yml` altogether.

I suppose  in `tests.yml`, we are testing an editable install of poetry-core that is installed via poetry whereas in `integration.yml` we let tox install poetry-core, which IIRC builds an sdist and installs it (via pip?). I wonder if it's worth to keep this part (test installed sdist)?

Or maybe it makes more sense to extend https://github.com/python-poetry/poetry-core/blob/e81166e744fb3944ad4089cd526a73182e1d3483/tests/integration/test_pep517_backend.py#L22

which already tests that pip can build poetry-core so that we also verify that pip can install poetry-core built by poetry?

finswimmer:
>  I wonder if it's worth to keep this part (test installed sdist)?

I generally prefer testing against installed sdist over testing against an editable install. It better uncovers path problems.

dimbleby:
while the discussion about testing against an sdist is kind of interesting, it's surely at best by some sort of good luck that the integration tests happened to be executing this, by being the only part of either poetry or poetry-core that still uses tox.

If y'all think that such tests are valuable then https://github.com/python-poetry/poetry-core/pull/573#issuecomment-1518521302 sounds right: add explicit tests covering the things that you think are worth covering.

imo there's no need to block the removal of tox here on doing that.

radoering:
> alternatively I'm not sure why it's useful for these tests to be separate from all the others, maybe just tag on a `pytest --integration tests/integration` at the end of `tests.yml` and delete `integration.yml` altogether.

What about this suggestion? In poetry, we already do it [that way](https://github.com/python-poetry/poetry/blob/b17b0820dd423a9c6c7df7cfb7d5958715168c94/.github/workflows/main.yml#L105). Since the integration tests themself only take about half a minute I don't think it's worth to keep them separated.

Secrus:
> What about this suggestion? In poetry, we already do it [that way](https://github.com/python-poetry/poetry/blob/b17b0820dd423a9c6c7df7cfb7d5958715168c94/.github/workflows/main.yml#L105). Since the integration tests themself only take about half a minute I don't think it's worth to keep them separated.

Sure, done.

.github/workflows/tests.yml
```diff
@@ -71,5 +71,8 @@ jobs:
       - name: Run pytest
         run: poetry run python -m pytest -p no:sugar -q tests/
 
+      - name: Run pytest
```
radoering, line 74:
```suggestion
      - name: Run pytest integration
```

... or something else so we don't have the same name twice

Secrus: Fixed

pyproject.toml
```diff
@@ -42,7 +41,7 @@ tomli-w = "^1.0.0"
 mypy = ">=1.0"
 types-jsonschema = ">=4.4.4"
 types-setuptools = ">=57.4.14"
-
+virtualenv = ">=20.21"
```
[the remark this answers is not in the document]
Secrus: Right, moved.

radoering (approved):

PR #579, opened 2023-05-05
fix allows_any() with local versions

dimbleby:
Make `allows_any()` more like `intersect()`.  

It doesn't make sense that `1.2.3+cpu` and `1.2.3` have an intersection (at `1.2.3+cpu`), while `"1.2.3+cpu".allows_any("1.2.3")` is `False`.

(Possible alternative implementation is just to take the intersection and ask whether it is empty.  Conceptually that's clean, perhaps it's pointless to worry about the unnecessary construction of the `EmptyConstraint()` when the answer is "no").

see also https://github.com/python-poetry/poetry/issues/7876

### DIFF
--- a/src/poetry/core/constraints/version/version_range.py
+++ b/src/poetry/core/constraints/version/version_range.py
@@ -124,7 +124,14 @@ class VersionRange(VersionRangeConstraint):
             return False
 
         if isinstance(other, Version):
-            return self.allows(other)
+            if self.allows(other):
+                return True
+
+            # Although `>=1.2.3+local` does not allow the exact version `1.2.3`, both of
+            # those versions do allow `1.2.3+local`.
+            return (
+                self.min is not None and self.min.is_local() and other.allows(self.min)
+            )
 
         if isinstance(other, VersionUnion):
             return any(self.allows_any(constraint) for constraint in other.ranges)
@@ -143,10 +150,20 @@ class VersionRange(VersionRangeConstraint):
         if isinstance(other, VersionUnion):
             return other.intersect(self)
 
-        # A range and a Version just yields the version if it's in the range.
         if isinstance(other, Version):
+            # A range and a Version just yields the version if it's in the range.
             if self.allows(other):
                 return other
+
+            # `>=1.2.3+local` intersects `1.2.3` to return `>=1.2.3+local,<1.2.4`.
+            if self.min is not None and self.min.is_local() and other.allows(self.min):
+                upper = other.stable.next_patch()
+                return VersionRange(
+                    min=self.min,
+                    max=upper,
+                    include_min=self.include_min,
+                    include_max=False,
+                )
 
             return EmptyConstraint()
 
--- a/src/poetry/core/constraints/version/version.py
+++ b/src/poetry/core/constraints/version/version.py
@@ -93,19 +93,20 @@ class Version(PEP440Version, VersionRangeConstraint):
         )
 
     def allows_any(self, other: VersionConstraint) -> bool:
+        intersection = self.intersect(other)
+        return not intersection.is_empty()
+
+    def intersect(self, other: VersionConstraint) -> VersionConstraint:
         if isinstance(other, Version):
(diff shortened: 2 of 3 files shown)


dimbleby:
ugh, as always, things get worse when you look at them!

```python
#!/usr/bin/env python3

from poetry.core.constraints.version import Version, VersionRange

v123 = Version.parse("1.2.3")
v123_local = Version.parse("1.2.3+local")
v123_local_etc = VersionRange(min=v123_local, include_min=True)

print(f"{v123_local.intersect(v123).is_empty()=}")
print(f"{v123_local_etc.intersect(v123).is_empty()=}")
```

the range `v123_local_etc` is surely larger than the individual version `v123_local`, so it doesn't make a lot of sense that the second intersection is empty and the first is not.

dimbleby:
another paradox in the current implementation:

```python
v123 = Version.parse("1.2.3")
v123_local = Version.parse("1.2.3+local")
v123_local_etc = VersionRange(min=v123_local, include_min=True)

print(f"{v123.allows(v123_local)=}")
print(f"{v123.allows_any(v123_local_etc)=}")
```

`1.2.3` allows `1.2.3+local`, but does not allow any of `>=1.2.3+local`

dimbleby:
Not planning on doing anything more with this, AFAIK it should be good to merge.  Confirmed that this allows the sequence:
```
$ poetry source add --default pytorch-index https://download.pytorch.org/whl/cu118
$ poetry add torch torchvision torchaudio
```

to work

I think it's maybe still wrong when calculating the intersection of eg "1.2.3" and "<=1.2.3+local" which ought to be a range something like ">=1.2.3, <=1.2.3+local" but today probably comes out as just "1.2.3".  However I don't have the enthusiasm to do anything about that now, and this MR should be a clear improvement even if there is more that could be done.  Maybe another time.

src/poetry/core/constraints/version/version_range.py
```diff
@@ -143,11 +150,21 @@ def intersect(self, other: VersionConstraint) -> VersionConstraint:
...
 
-        # A range and a Version just yields the version if it's in the range.
         if isinstance(other, Version):
+            # A range and a Version just yields the version if it's in the range.
             if self.allows(other):
                 return other
 
+            # `>=1.2.3+local` intersects `1.2.3` to return `>=1.2.3+local,<1.2.4`.
+            if self.min is not None and self.min.is_local() and other.allows(self.min):
+                upper = other.stable.next_patch()
```
radoering, line 160:
```suggestion
                upper = other.stable.next()
```

`next_patch()` will only work for versions with 3 parts, won't it? We should probably add a test with a version with less parts and another test with more parts.

radoering (changes requested):

src/poetry/core/constraints/version/version_range.py
dimbleby: `next_patc()` is correct for versions iwth fewer than three parts, no?

Happy to see someone submit a follow-up to deal with 1.2.3.4+local, meanwhile this seems like an edge case that I don't care to worry about...

radoering: I see it now. Strictly speaking, you can't specify the upper bound. Independent from the precision of the version it's always like `major.minor.patch. ... .0.0.0. ... .1`. In the example, we mistakenly allow `1.2.3.1`. That's probably something we have to accept at a certain precision...

dimbleby: Yeah, if y'all want to insist that this MR makes everything perfect then sorry: not gonna happen.  Feel free to make it better or to follow up with MRs that improve it.

My position is that this is a definite improvement, solving an actual problem, and that's what I have the time / energy to do right now.  Take it or leave it!

radoering: I was just trying to say that a perfect solution might not be possible with our concept of version ranges. I totally agree that this MR is an improvement.

radoering (approved):

PR #581, opened 2023-05-09
feat: add [tool.poetry.tasks] section

finswimmer:
Resolves: python-poetry#

### DIFF
--- a/src/poetry/core/json/schemas/poetry-schema.json
+++ b/src/poetry/core/json/schemas/poetry-schema.json
@@ -206,6 +206,15 @@
         }
       }
     },
+    "tasks": {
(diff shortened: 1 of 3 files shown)


BEllis:
You can already do this with poethepoet

https://pypi.org/project/poethepoet/

radoering:
> You can already do this with poethepoet

Yeah, there is already a discussion in python-poetry/poetry#7896 that was forgotten to link here.

I'm closing this issue because even if we decide to proceed with python-poetry/poetry#7896 (which does not seem likely at the moment), there is no need to make a change in poetry-core. The schema can be extended in poetry alone.

PR #585, opened 2023-05-12
avoid duplicate extras in lockfiles

dimbleby:
there are duplicate extras in poetry.lock files.

I'm not entirely sure what are the circumstances that trigger this but it shows up in
- one of the fixtures in the companion pull request in `poetry` proper
- poetry's own lockfile (see the `tests-no-zope` extras on `attrs`)

Because I don't know exactly how to trigger this, I don't know how to add an explicit testcase for it.  But as I say there's a `poetry` testcase that currently demonstrates the faulty behaviour and which is fixed by this; and post-fix it's kinda self-testing - of course a set can't contain duplicates as a list could

I expect that the entanglement between the `poetry` and `poetry-core` pipelines will make it challenging to figure out how to merge this

### DIFF
--- a/src/poetry/core/factory.py
+++ b/src/poetry/core/factory.py
@@ -177,7 +177,7 @@ class Factory:
         extras = config.get("extras", {})
         for extra_name, requirements in extras.items():
             extra_name = canonicalize_name(extra_name)
-            package.extras[extra_name] = []
+            package.extras[extra_name] = set()
 
             # Checking for dependency
             for req in requirements:
@@ -186,7 +186,7 @@ class Factory:
                 for dep in package.requires:
                     if dep.name == req.name:
                         dep.in_extras.append(extra_name)
-                        package.extras[extra_name].append(dep)
+                        package.extras[extra_name].add(dep)
 
         if "build" in config:
             build = config["build"]
--- a/src/poetry/core/packages/package.py
+++ b/src/poetry/core/packages/package.py
@@ -107,7 +107,7 @@ class Package(PackageSpecification):
         self._license: License | None = None
         self.readmes: tuple[Path, ...] = ()
 
-        self.extras: dict[NormalizedName, list[Dependency]] = {}
+        self.extras: dict[NormalizedName, set[Dependency]] = {}
 
         self._dependency_groups: dict[str, DependencyGroup] = {}
 
--- a/tests/masonry/builders/test_sdist.py
+++ b/tests/masonry/builders/test_sdist.py
@@ -75,7 +75,7 @@ def test_convert_dependencies() -> None:
     assert result == (main, extras)
 
     package = ProjectPackage("foo", "1.2.3")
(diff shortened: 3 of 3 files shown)


radoering:
> I'm not entirely sure what are the circumstances that trigger this but it shows up in

Just duplicate extras in metadata:

```
$ curl -ssL https://pypi.org/pypi/attrs/17.4.0/json | jq '.info.requires_dist' | grep zope.interface | grep dev
  "zope.interface; extra == 'dev'",
  "zope.interface; extra == 'dev'",
```

Or duplicate extras due to normalization:

```
$ curl -ssL https://pypi.org/pypi/attrs/22.2.0/json | jq '.info.requires_dist | sort_by(.)'
[
  ...
  "cloudpickle ; (platform_python_implementation == \"CPython\") and extra == 'tests-no-zope'",
  "cloudpickle ; (platform_python_implementation == \"CPython\") and extra == 'tests_no_zope'",
  ...
  "hypothesis ; extra == 'tests-no-zope'",
  "hypothesis ; extra == 'tests_no_zope'",
  "mypy (<0.990,>=0.971) ; (platform_python_implementation == \"CPython\") and extra == 'tests-no-zope'",
  "mypy (<0.990,>=0.971) ; (platform_python_implementation == \"CPython\") and extra == 'tests_no_zope'",
  ...
  "pympler ; extra == 'tests-no-zope'",
  "pympler ; extra == 'tests_no_zope'",
  "pytest (>=4.3.0) ; extra == 'tests-no-zope'",
  "pytest (>=4.3.0) ; extra == 'tests_no_zope'",
  "pytest-mypy-plugins ; (platform_python_implementation == \"CPython\" and python_version < \"3.11\") and extra == 'tests-no-zope'",
  "pytest-mypy-plugins ; (platform_python_implementation == \"CPython\" and python_version < \"3.11\") and extra == 'tests_no_zope'",
  "pytest-xdist[psutil] ; extra == 'tests-no-zope'",
  "pytest-xdist[psutil] ; extra == 'tests_no_zope'",
  ...
]
```

> I expect that the entanglement between the poetry and poetry-core pipelines will make it challenging to figure out how to merge this

It's just two tests that are failing. We could simply replace

```
assert package.extras == {
    "extras-a": [get_dependency("pendulum", ">=1.4.4")],
    "extras-b": [get_dependency("cachy", ">=0.2.0")],
}
```
with

```
assert set(package.extras) == {"extras-a", "extras-b"}
assert set(package["extras-a"]) == {get_dependency("pendulum", ">=1.4.4")}
assert set(package["extras-b"]) == {get_dependency("cachy", ">=0.2.0")}
```
to make them forward compatible.

radoering:
I wonder if this will introduce a new bug for an edge case since we do not consider markers when comparing dependencies for equality. I suppose dependencies are not yet merged by constraint when they are added to the set. If we have two dependencies with the same constraint but different markers, we will probably lose information.

PR #586, opened 2023-05-14
include missing file from vendored packaging

dimbleby:
recent round of dependency updates omitted a new file added to the vendored `packaging`.

possibly this in some sense doesn't matter, it may be that poetry-core happens never to use this file.  Still, I suppose it was an accident


Secrus (approved):

radoering:
> Still, I suppose it was an accident

Absolutely, thanks for catching this. It seems there are even two more missing files (but not from packaging): #587

PR #587, opened 2023-05-14
add further missing files from vendoring

radoering:
Due to #586 I checked again and there were two more files. I need to pay better attention to new files when updating vendored packages in the future.

### DIFF
--- a/src/poetry/core/_vendor/pkgutil_resolve_name.LICENSE
+++ b/src/poetry/core/_vendor/pkgutil_resolve_name.LICENSE
@@ -1,0 +1,75 @@
+The MIT License (MIT)
+
+Copyright (c) 2020 Thomas Grainger.
+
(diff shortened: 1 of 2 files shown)


Secrus (approved):

dimbleby:
hmm, those files don't show up when I run the vendoring, how curious.

why are they ending up at the top-level?  feels like something is off here

dimbleby:
so I'm probably not getting them because I'm running with too recent a python, and `pkgutil-resolve-name` is only wanted on `python < 3.9`

Missing parts of `packaging` seemed definitely hairy, but it might be sensible to omit this - cf the treatment of importlib and zipp.  It's used only by the jsonschema cli, which is of no interest

radoering:
I hate vendoring.

> but it might be sensible to omit this - cf the treatment of importlib and zipp. It's used only by the jsonschema cli, which is of no interest

Agreed.

PR #598, opened 2023-05-29
Update wheel.py to support generic tag when no extensions found

bsdz:
If no extensions are built (due to conditions) then do not tag as platform specific wheel.

Resolves: python-poetry#8039

If this PR looks reasonable I'm happy to add some documentation. I'd need pointers on where to add tests.

### DIFF
--- a/src/poetry/core/masonry/builders/wheel.py
+++ b/src/poetry/core/masonry/builders/wheel.py
@@ -67,6 +67,7 @@ class WheelBuilder(Builder):
             self._original_path = original.parent
         self._editable = editable
         self._metadata_directory = metadata_directory
+        self._has_libs = None        
 
     @classmethod
     def make_in(
@@ -189,6 +190,7 @@ class WheelBuilder(Builder):
                         # The result of building the extensions
                         # does not exist, this may due to conditional
                         # builds, so we assume that it's okay
+                        self._has_libs = False                        
                         return
 
                     lib = libs[0]
@@ -370,8 +372,12 @@ for t in packaging_tags.sys_tags():
         return decode(output).strip().splitlines()
 
     @property
+    def is_pure_lib(self) -> bool:
+        return self._package.build_script is None or self._has_libs == False
+    
+    @property
     def tag(self) -> str:
-        if self._package.build_script:
+        if not self.is_pure_lib:
             if self.executable != Path(sys.executable):
                 # poetry-core is not run in the build environment
                 # -> this is probably not a PEP 517 build but a poetry build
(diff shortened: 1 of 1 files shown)


dimbleby:
Surely there are lots of other ways for platform-specific files to make their way into a wheel eg as scripts, or by being built out-of-band and copied around, or whatever.  This change will break the wheel name for anyone who is relying on such things.

(Also as a matter of coding style: `bool | None` is a weird type for a field that can take two values, it wants to be a boolean - perhaps with a better name!  And I don't know that `str(boolean).lower()` is really an improvement on `"true" if boolean else "false"`).

bsdz:
> Surely there are lots of other ways for platform-specific files to make their way into a wheel eg as scripts, or by being built out-of-band and copied around, or whatever.  This change will break the wheel name for anyone who is relying on such things.

I'm open to suggestions on fixing this use case. See the issue I raised alongside with a mypyc example (but also applies to cython).

> (Also as a matter of coding style: `bool | None` is a weird type for a field that can take two values, it wants to be a boolean - perhaps with a better name!  And I don't know that `str(boolean).lower()` is really an improvement on `"true" if boolean else "false"`).

I disagree. It's quite common to have a tenary variable being true, false and "don't know".

> 
> Also: no testcase.

I added a comment to this effect and, now, given your comment suggests this PR is DoA perhaps writing a test case would have been futile in any case.

dimbleby:
Unfortunately it is easier to pick holes than to propose solutions.  Building with extensions etc is an under-loved part of the code and may need more invasive changes to address this sort of thing.

> ... ternary variable ...

but this isn't a ternary variable:  it never takes the value `True`.

bsdz:
> > ... ternary variable ...
> 
> but this isn't a ternary variable: it never takes the value `True`.

It doesn't mean it's not a ternary variable. At the outset the variable is indeterminate as no test has been done to test if it "has libs" or not. Once it has been determined there are no libs then it is set to False. I could have added

```python
                    if not libs:
                        # The result of building the extensions
                        # does not exist, this may due to conditional
                        # builds, so we assume that it's okay
                        self._has_libs = False
                        return
                   else:
                        self._has_libs = True
```

but there would be no point as the variable will never be tested for being True. 

It's still a ternary variable.

However, I feel we are arguing over semantics here.

dimbleby:
it's certainly easier to debate whether a variable that takes two values can properly be described as ternary, than it is to figure out how to achieve the goal of this MR!

bsdz:
> Also as a matter of coding style: `bool | None` is a weird type for a field that can take two values..

Oh wow!, look [its been used in this repo several times already](https://github.com/search?q=org%3Apython-poetry%20%22bool%20%7C%20None%22&type=code) as well as [thousands more times](https://github.com/search?q=%22bool+%7C+None%22&type=code) in other prominent python projects.

dimbleby:
I think we are talking past each other.  Of course `bool | None` is not in itself a bad type - for cases where there are three distinct states that the code cares about.  

At least in the first of those examples that you linked - I didn't go through them all! - `None` and `True` and `False` have different meanings.

In this MR there are only two states that the code cares about: "I have built and there are no libs" or the opposite of that.  So a two-state type is appropriate.

PR #600, opened 2023-06-03
Address TODO in pyproject/tables

Secrus:
There was a TODO comment left, this addresses it.

### DIFF
--- a/src/poetry/core/pyproject/tables.py
+++ b/src/poetry/core/pyproject/tables.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
-from contextlib import suppress
+from dataclasses import dataclass
+from dataclasses import field
 from pathlib import Path
 from typing import TYPE_CHECKING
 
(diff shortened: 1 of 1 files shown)


src/poetry/core/pyproject/tables.py
```diff
@@ -9,18 +10,15 @@
...
-        )
-        self.requires = requires if requires is not None else ["setuptools", "wheel"]
-        self._dependencies: list[Dependency] | None = None
+    build_backend: str | None = field(default=None)
+    requires: list[str] = field(default_factory=list)
+    _dependencies: list[Dependency] | None = field(default=None, init=False)
+
+    def __post_init__(self) -> None:
+        self.build_backend = self.build_backend or "setuptools.build_meta:__legacy__"
+        self.requires = self.requires or ["setuptools", "wheel"]
```
radoering, line 21: According to the old implementation `self.requires` should only be set to this default if it's `None` (not if it's an empty list). However, I assume this distinction is not necessary because each build-backend requires at least itself so there is no case where an empty list makes sense.

By the way, we should probably require `setuptools >= 40.8.0`. I suppose in older versions there is no `"setuptools.build_meta:__legacy__"`, cf [build](https://github.com/pypa/build/blob/e0c59cc4e17e63cfe9dbd4f366fae6d51b073282/src/build/__init__.py#L51-L52)

radoering (changes requested):

src/poetry/core/pyproject/tables.py
```diff
@@ -9,18 +10,15 @@
...
-        self, build_backend: str | None = None, requires: list[str] | None = None
-    ) -> None:
-        self.build_backend = (
-            build_backend
-            if build_backend is not None
-            else "setuptools.build_meta:__legacy__"
-        )
-        self.requires = requires if requires is not None else ["setuptools", "wheel"]
-        self._dependencies: list[Dependency] | None = None
+    build_backend: str | None = field(default=None)
```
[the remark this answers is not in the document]
Secrus: Went with this suggestion and restored the previous code, deleted the TODO and did a small caching refactor in the surrounding code.

src/poetry/core/pyproject/tables.py
Secrus: Restored previous code

PR #602, opened 2023-06-04
improve use of subprocess

dimbleby:
a companion to https://github.com/python-poetry/poetry/pull/8060, using `text=True` for subprocess.

### DIFF
--- a/src/poetry/core/masonry/builders/wheel.py
+++ b/src/poetry/core/masonry/builders/wheel.py
@@ -27,7 +27,6 @@ from poetry.core.masonry.builders.sdist import SdistBuilder
 from poetry.core.masonry.utils.helpers import distribution_name
 from poetry.core.masonry.utils.helpers import normalize_file_permissions
 from poetry.core.masonry.utils.package_include import PackageInclude
-from poetry.core.utils.helpers import decode
 from poetry.core.utils.helpers import temporary_directory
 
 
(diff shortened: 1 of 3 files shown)


src/poetry/core/masonry/builders/wheel.py
```diff
@@ -361,13 +361,14 @@ def _get_sys_tags(self) -> list[str]:
 """,
                 ],
                 stderr=subprocess.STDOUT,
+                text=True,
             )
         except subprocess.CalledProcessError as e:
             raise RuntimeError(
                 "Failed to get sys_tags for python interpreter"
                 f" '{self.executable.as_posix()}':\n{decode(e.output)}"
```
radoering, line 369: If we pass `text=True` `e.output` will be decoded, too. Then we can remove `poetry.core.utils.helpers.decode` altogether since it was only introduced for this part in the last patch release.

radoering (changes requested):

src/poetry/core/masonry/builders/wheel.py
dimbleby: it wasn't clear to me whether that was true or not - some of the subprocess errors say "This is always bytes when any output was captured regardless of the text=True setting" and some of them just... don't say that.

radoering: OK, that was only an observation on my part. If we can't count on that, you are right to do the `decode()`.

dimbleby: however experiment seems to confirm that this one is indeed a string

PR #609, opened 2023-06-24
perf: cache results of `cnf`, `dnf` and `_merge_single_markers`

radoering:
Picking up the suggestion in https://github.com/python-poetry/poetry/pull/7257#issuecomment-1604514982

In order to cache `cnf` and `dnf`, we can no longer consider composite markers and constraints with the same sub elements but a different order to be equal because order is relevant when building an intersection or union. Neglecting order leads to different results if the result of an "equal" marker with a different order has already been cached. This can even result in a RecursionError, as the test suite demonstrates.

Performance measurements see https://github.com/python-poetry/poetry/pull/7257#issuecomment-1605501632

### DIFF
--- a/src/poetry/core/constraints/generic/constraint.py
+++ b/src/poetry/core/constraints/generic/constraint.py
@@ -123,9 +123,9 @@ class Constraint(BaseConstraint):
         return other.intersect(self)
 
     def union(self, other: BaseConstraint) -> BaseConstraint:
-        if isinstance(other, Constraint):
-            from poetry.core.constraints.generic.union_constraint import UnionConstraint
+        from poetry.core.constraints.generic.union_constraint import UnionConstraint
 
+        if isinstance(other, Constraint):
             if other == self:
                 return self
 
@@ -139,6 +139,10 @@ class Constraint(BaseConstraint):
                 return UnionConstraint(self, other)
 
             return AnyConstraint()
+
+        # to preserve order (functionally not necessary)
+        if isinstance(other, UnionConstraint):
+            return UnionConstraint(self).union(other)
 
         return other.union(self)
 
--- a/src/poetry/core/constraints/generic/multi_constraint.py
+++ b/src/poetry/core/constraints/generic/multi_constraint.py
@@ -128,11 +128,7 @@ class MultiConstraint(BaseConstraint):
         return self._constraints == other._constraints
 
     def __hash__(self) -> int:
-        h = hash("multi")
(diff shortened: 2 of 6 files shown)


dimbleby:
> In order to cache cnf and dnf, we can no longer consider composite markers and constraints with the same sub elements but a different order to be equal because order is relevant when building an intersection or union. Neglecting order leads to different results if the result of an "equal" marker with a different order has already been cached. This can even result in a RecursionError, as the test suite demonstrates.

this all makes me a bit uncomfortable: 

- in my mental model markers with the same set of sub-markers really should be equal
- it's worrying that the code is so close to falling into recursion errors etc, perhaps it's all more fragile than we thought

on the other hand there are quite a lot of tests for this stuff, and if they're all passing then I suppose it's probably fine...!

should `__hash__` implementations be updated to match updated `__eq__`?  instead of xor-ing in each individual marker (which was order-independent) it might make more sense to xor in `hash(tuple(sub-markers))`?  Or just construct a tuple that goes eg `("union", *self._markers)` and hash that.

(If they're ordered, perhaps submarkers should be a tuple rather than a list all along, so as to make them immutable??)

re the `SingleMarker`: converting to a string for hash / equality purposes feels a bit weird?  A typical trick is to define a method that returns a tuple of all the identifying features and use that for both hash and equality.  I suppose that's kinda what you've done but with a string instead of a tuple...

radoering:
> * in my mental model markers with the same set of sub-markers really should be equal

I don't think so (anymore). It's just a matter of concept what's defined as equal.

If markers with the same subset of markers should be equal, shouldn't (more generally) markers that can be transformed into each other be equal, too? However, I don't think that works well with our algorithm.

IMO, a better concept is that only markers that have the same string representation are equal. It's easier to achieve and fits better with our algorithm.

> should `__hash__` implementations be updated to match updated `__eq__`?

It's not necessary but it might make sense if the hash calculation does not become more expensive so we get less different markers with equal hashes and thus have to do less comparisons for equality.

> (If they're ordered, perhaps submarkers should be a tuple rather than a list all along, so as to make them immutable??)

IIRC, we have always considered submarkers immutable so making them a tuple instead of a list might make sense.

> re the `SingleMarker`: converting to a string for hash / equality purposes feels a bit weird? A typical trick is to define a method that returns a tuple of all the identifying features and use that for both hash and equality. I suppose that's kinda what you've done but with a string instead of a tuple...

Agreed.

Secrus (approved):

PR #618, opened 2023-07-30
replace clone to clone_shallow and copy only as much as you need attributes.

higumachan:
Resolves: [python-poetry#8250](https://github.com/python-poetry/poetry/issues/8250)

- [X] Updated **documentation** for changed code.

## Description

This change was made under the assumption that if the tests pass, there should be no problem. However, just replacing the clone with a shallow copy caused several tests to fail, so I made "one level deeper" copies sufficient to fix those tests.

I understand that there is a trade-off with code maintainability regarding this change, so I would appreciate some feedback.

If we are to implement this while ensuring future changeability, I think we could solve this by creating code similar to [Cow (Copy on Write)](https://en.wikipedia.org/wiki/Copy-on-write).

Thanks.


mtkennerly:
Hi, just cross-posting from https://github.com/python-poetry/poetry/pull/6348#event-10036998403:

I have [a plugin](https://github.com/mtkennerly/poetry-dynamic-versioning) that does VCS-based dynamic versioning, and this would break the handling for dependencies that also use the plugin. If you make this the default, please also add an option for Git dependencies to specify a custom depth, including the full history. For example:

```toml
[tool.poetry.dependencies]
requests = { git = "https://github.com/requests/requests.git", depth = 0 }
```

dimbleby:
@mtkennerly you have misunderstood this MR, it has nothing at all to do with git clone

eamanu:
Did you run a profiler to know if this fix the issue?

src/poetry/core/packages/package.py
```diff
@@ -481,7 +481,8 @@ def without_optional_dependency_groups(self: T) -> T:
         """
         Returns a clone of the package without optional dependency groups.
         """
-        package = self.clone()
+        package = self.clone_shallow()
+        package._dependency_groups = package._dependency_groups.copy()
```
eamanu, line 485: a little comment explaining why we are copying the package._depndency_groups wouldbe great

radoering:
Superseded by #663.

PR #633, opened 2023-08-30
Fix editable installs with extension modules

robbotorigami:
Resolves: python-poetry/poetry#8377

Credit to @matthewwardrop for the fix (#335). I've rebased onto the current main branch and added unit tests for the modified functionality. I've also tested against the projects I was having issues with and confirmed that this resolves the issues I was having.

### DIFF
--- a/src/poetry/core/masonry/builders/wheel.py
+++ b/src/poetry/core/masonry/builders/wheel.py
@@ -182,6 +182,11 @@ class WheelBuilder(Builder):
                     finally:
                         os.chdir(current_path)
 
+                    if self._editable:
+                        # For an editable install, the extension modules will be built
+                        # in-place - so there's no need to copy them into the zip
+                        return
+
                     build_dir = self._path / "build"
                     libs: list[Path] = list(build_dir.glob("lib.*"))
                     if not libs:
--- a/src/poetry/core/masonry/builders/wheel.py
+++ b/src/poetry/core/masonry/builders/wheel.py
@@ -221,6 +221,15 @@ class WheelBuilder(Builder):
             )
 
     def _run_build_command(self, setup: Path) -> None:
+        if self._editable:
+            subprocess.check_call(
+                [
+                    self.executable.as_posix(),
+                    str(setup),
+                    "build_ext",
+                    "--inplace",
+                ]
+            )
         subprocess.check_call(
             [
                 self.executable.as_posix(),
--- a/tests/masonry/builders/test_wheel.py
+++ b/tests/masonry/builders/test_wheel.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import importlib.machinery
 import os
 import re
 import shutil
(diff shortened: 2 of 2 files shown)


src/poetry/core/masonry/builders/wheel.py
```diff
@@ -216,6 +216,15 @@ def _copy_file_scripts(self, wheel: zipfile.ZipFile) -> None:
...
+        if self._editable:
+            subprocess.check_call(
+                [
+                    self.executable.as_posix(),
+                    str(setup),
+                    "build_ext",
+                    "--inplace",
+                ]
+            )
         subprocess.check_call(
```
dimbleby, line 228: is it necessary / useful also to call `setup.py build` in the editable case, or was `setup.by build_ext --inplace` enough?

dimbleby: not sure about any of that.  since this code is only executed for packages with a build script, poetry surely can't be relying on it do deal with whatever `build_scripts` does?  (Probably it's equivalent to poetry's `WheelBuilder._copy_file_scripts()`?)

maybe what we are learning here is that it should have just been `build_ext` all along; possibly `--in-place` for editable intalls.

src/poetry/core/masonry/builders/wheel.py
```diff
@@ -184,7 +184,7 @@ def _build(self, wheel: zipfile.ZipFile) -> None:
 
                     build_dir = self._path / "build"
                     libs: list[Path] = list(build_dir.glob("lib.*"))
-                    if not libs:
+                    if self._editable or not libs:
```
dimbleby, line 187: nit: pointless to have constructed the build_dir and done the glob if we're just going to duck out anyway

src/poetry/core/masonry/builders/wheel.py
robbotorigami: I'm inclined to agree with you about `build_scripts`; I took a closer look and I don't think there's anything that poetry needs out of it. I've never encountered `build_clib`, so I'm not sure what the usage pattern would be. It doesn't support a build inplace flag, so I assume it would be enough to just call it as a separate command. If preferred, I'd be happy to modify this to e.g.:

```python
        build_ext_cmd = [
            self.executable.as_posix(),
            str(setup),
            "build_ext",
        ]
        if self._editable:
            build_ext_cmd.append("--inplace")
        subprocess.check_call(build_ext_cmd)

        # Optionally add:
        subprocess.check_call(
            [
                self.executable.as_posix(),
                str(setup),
                "build_clib",
            ]
        )
```

dimbleby: I don't think `build_clib` is relevant either, that looks as though it deals with `libraries` in setup.py, and poetry's generated `setup.py` does not include that.

So yeah, the first half of that is what I think this probably ought to look like; so long as it doesn't seem to break anything...

robbotorigami: I think if it were to be used, `libraries` would get injected into setup_kwargs inside build.py similar to `ext_modules`. 

I'm happy to drop it if it's not something the maintainers want to support, just don't want to inadvertently break anyone's use case haha

src/poetry/core/masonry/builders/wheel.py
robbotorigami: Agreed, especially with the explanatory comment below it, it probably makes sense to pull it out into a separate check.

Also...
Looking at it, using the first folder named `lib.*` is actually pretty sketchy, it won't work correctly if there are multiple build targets being created, e.g. if the build is ran for multiple python versions without cleaning the build dir in between - which explains an issue I've ran into. I'll make a separate issue for it (and if I have time later find a fix)

src/poetry/core/masonry/builders/wheel.py
dimbleby: > libraries would get injected into setup_kwargs inside build.py similar to ext_modules.

ah, perhaps that's right.  

don't think I mind much either way then.  `build_ext` followed by `build` still feels logically wasteful; but `build_ext` followed by `build_clib` isn't such an obvious win

robbotorigami: Sounds good, I think I'll stick with `build_ext` followed by `build` in that case, it does feel a little more wasteful, but also might be a little less fragile in the (perhaps unlikely) event that another useful subcommand is added to `build`.

radoering (approved):
LGTM 👍

PR #636, opened 2023-09-03
markers: add special handling for `extra`

radoering:

`extra` is not like any other marker, it is, well, "extra". Whereas normal markers can only have one value at a time, `extra` can have multiple values at the same time.

For example, a marker like `sys_platform == "win32" and sys_platform == "linux"` can never be fulfilled because `sys_platform` can only be either `"win32"` or `"linux"` (not both). By contrast, a marker like `extra == "a" and extra == "b"` can be fulfilled if both extras are requested.

Therefore, this PR disables some simplifications for `extra` markers. Further, it adds the option to validate environments with multiple extras and makes the validation robust against unnormalized names.

### DIFF
--- a/src/poetry/core/version/markers.py
+++ b/src/poetry/core/version/markers.py
@@ -13,6 +13,8 @@ from typing import Generic
 from typing import TypeVar
 from typing import Union
 
+from packaging.utils import canonicalize_name
+
 from poetry.core.constraints.generic import BaseConstraint
 from poetry.core.constraints.generic import Constraint
 from poetry.core.constraints.generic import MultiConstraint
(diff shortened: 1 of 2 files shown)


src/poetry/core/version/markers.py
```diff
@@ -244,8 +247,25 @@ def validate(self, environment: dict[str, Any] | None) -> bool:
             return True
 
         # The type of constraint returned by the parser matches our constraint: either
-        # both are BaseConstraint or both are VersionConstraint.  But it's hard for mypy
-        # to know that.
+        # both are BaseConstraint or both are VersionConstraint. But it's hard for mypy
```
dimbleby, line 250: you have increased the distance between this comment and the code that it is talking about; and increased the verbiage to compensate

surely better just to leave the comment and the code next to each other?!

src/poetry/core/version/markers.py
```diff
@@ -244,8 +247,25 @@ def validate(self, environment: dict[str, Any] | None) -> bool:
...
+        # "extra == 'a'" will be true if "a" is one of the active extras.
+        # "extra != 'a'" will be true if "a" is not one of the active extras.
+        # Further, extra names are normalized for comparison.
+        if self._name == "extra":
+            extras = environment["extra"]
+            if isinstance(extras, str):
+                extras = (extras,)
+            assert isinstance(extras, Collection)
+            assert isinstance(self._constraint, Constraint)
+            normalized_value = canonicalize_name(self._constraint.value)
```
dimbleby, line 263: are the environment extras guaranteed normalized at this point?

src/poetry/core/version/markers.py
radoering: You are right. In an earlier version, I used `allows()` and needed two more ignores before I realized I can just use `in` / `not in` since I have to make a distinction anyway.

src/poetry/core/version/markers.py
radoering: They are normalized in the only usage so far (in the companion PR) but it might make sense to be more robust and normalize (again) since environment is just defined as `dict[str, Any]`. What do you think?

dimbleby: I don't very much mind either way, so long as you're confident that there is no bug.

As a fan of having the typechecking catch as many bugs as possible I'm tempted to encourage defining a `TypedDict` or some other actually structured object to describe the marker environment: and letting the effects of that work their way through the code...  

But I am also sympathetic to the line of argument that says that if you wanted to write code that way then you didn't want to be writing python at all: and you have carefully reviewed the one place that puts `extras` here and it's fine.

Secrus (approved):

PR #641, opened 2023-09-16
Use pathlib instead of os.path module

Secrus:
Migrated from `os.path` to `pathlib` usage.

### DIFF
--- a/src/poetry/core/packages/utils/utils.py
+++ b/src/poetry/core/packages/utils/utils.py
@@ -1,7 +1,6 @@
 from __future__ import annotations
 
 import functools
-import posixpath
 import re
 import sys
 
@@ -106,7 +105,7 @@ def is_url(name: str) -> bool:
     ]
 
 
-def strip_extras(path: str) -> tuple[str, str | None]:
+def strip_extras(path: str) -> tuple[Path, str | None]:
     m = re.match(r"^(.+)(\[[^\]]+\])$", path)
     extras = None
     if m:
@@ -115,7 +114,7 @@ def strip_extras(path: str) -> tuple[str, str | None]:
     else:
         path_no_extras = path
 
-    return path_no_extras, extras
+    return Path(path_no_extras), extras
 
 
 @functools.lru_cache(maxsize=None)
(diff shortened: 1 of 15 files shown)


tests/utils/test_helpers.py
```diff
@@ -147,11 +146,11 @@ def test_temporary_directory_python_3_9_or_older(mocker: MockerFixture) -> None:
     mocked_temp_dir = mocker.patch("tempfile.TemporaryDirectory")
     mocked_mkdtemp = mocker.patch("tempfile.mkdtemp")
 
-    mocked_mkdtemp.return_value = "hello from test"
+    mocked_mkdtemp.return_value = Path("hello from test")
```
radoering, line 149: That does not make sense, since we are mocking `tempfile.mkdtemp`.

tests/utils/test_helpers.py
```diff
@@ -135,7 +134,7 @@ def test_temporary_directory_python_3_10_or_newer(mocker: MockerFixture) -> None
 
     mocker.patch.object(sys, "version_info", (3, 10))
     with temporary_directory() as tmp:
-        assert tmp
+        assert str(tmp)
```
radoering, line 137: This `assert` has always been quite weak. Now it does not make any sense since the string representation of a `Path` will always be truthy as far as I know. We should probably do it similar to the test some lines below and set the return value of `mocked_temp_dir`. Then, we can check for the correct path.

radoering (changes requested):

tests/utils/test_helpers.py
Secrus: Do you have any better idea? I am tempted to just `noqa` this and leave it as it was...

radoering: Does it result in a warning? I wonder how ruff would decide that this string should be `Path`?

src/poetry/core/packages/utils/utils.py
```diff
@@ -133,19 +132,19 @@ def is_python_project(path: Path) -> bool:
...
     if ext in ARCHIVE_EXTENSIONS:
         return True
     return False
 
 
-def splitext(path: str) -> tuple[str, str]:
+def splitext(path: Path) -> tuple[str, str]:
     """Like os.path.splitext, but take off .tar too"""
-    base, ext = posixpath.splitext(path)
+    base, ext = path.stem, path.suffix
```
radoering, line 145: That's not the same:

```
>>> posixpath.splitext('foo/bar.baz')
('foo/bar', '.baz')
>>> path = Path('foo/bar.baz')
>>> path.stem, path.suffix
('bar', '.baz')
```

However, it does not matter for our uses so it might still be ok (if we fix the docstring).

On the other side, we could change the signature to allow `str | Path` and keep the old implementation. We don't even need a `noqa` as it seems. I slightly tend to this variant. Then, we don't have to convert the `str` to `Path` in `link.py`.

src/poetry/core/packages/utils/utils.py
```diff
@@ -133,19 +132,19 @@ def is_python_project(path: Path) -> bool:
     return pyproject or setuptools_project
 
 
-def is_archive_file(name: str) -> bool:
+def is_archive_file(name: Path) -> bool:
```
radoering, line 135:
```suggestion
def is_archive_file(name: str | Path) -> bool:
```

From an interface's point of view, I think it makes sense to allow plain file names.

tests/utils/test_helpers.py
Secrus: I had a brainfart last time I guess, fixed.

tests/utils/test_helpers.py
Secrus: Fixed.

radoering (approved):

PR #642, opened 2023-09-17
jsonschema -> fastjsonschema

dimbleby:
switching from jsonschema to fastjsonschema, with the following intended advantages:

- jsonschema recently became not-pure-python, meaning that current versions are unsuitable for vendoring
- fastjsonschema has fewer dependencies, so that's less to vendor
- perhaps it's faster?!?

If I understand correctly it would even be possible to use fastjsonschema to [generate the schema-validation code](https://horejsek.github.io/python-fastjsonschema/#fastjsonschema.compile_to_code) ahead of time and ship that generated code instead of vendoring fastjsonschema itself.  That's what [setuptools seems to do](https://github.com/pypa/setuptools/blob/2255e6366c70b9813d115ae0a0bba329affbd0ac/setuptools/config/_validate_pyproject/fastjsonschema_validations.py).  Maybe some other time.

Probably this will cause failures in the downstream `poetry` tests, let's get opinion on whether this is a good direction before starting the tedious compatibility dance.

image is quite pleasing!


radoering:
> Probably this will cause failures in the downstream `poetry` tests, let's get opinion on whether this is a good direction before starting the tedious compatibility dance.
> 
> image is quite pleasing!

Looks amazing. I'm all in. (I would just give others some time to express any concerns.)

Secrus:
LGTM!

radoering (approved):

PR #645, opened 2023-09-30
Fix for handling of version ranges

### LINKED ISSUE #8475: poetry never resolves package combination: google-cloud-aiplatform kfp
- **Poetry version**: `1.6.1`
- **Python version**: `3.11.5`
- **OS version and name**: `docker` image `python:3.11.5`
- **pyproject.toml**: *Providing reproducible docker command instead*

- [X] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) stable Poetry version, installed using a recommended method.
- [X] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [X] I have consulted the [FAQ](https://python-poetry.org/docs/faq/) and [blog](https://python-poetry.org/blog/) for any relevant entries or release notes.
- [X] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option) and have included the output below.

## Issue
The following combination of packages never resolves: `google-cloud-aiplatform kfp`

Poetry is stuck in `Resolving dependencies...` after >10h.

Reproduced in both Native **Macos 13.5.1 (22G90)** and in **Docker**.

In Macos envs, cache and poetry.lock was removed like this, but despite this the error persists:

``` shell
poetry env remove --all \
&& poetry cache clear --all -q . \
&& rm -rf $(poetry config cache-dir)/artifacts \
&& rm -f poetry.lock
```

## Tested package combinations

❌ **Doesn't resolve** *(within reasonable time)*
- `google-cloud-aiplatform zenml mlflow gcsfs kfp` <-- **This combo was left running >10h**
- `google-cloud-aiplatform kfp`

✅ **Does resolve**
- `google-cloud-aiplatform zenml`
- `google-cloud-aiplatform zenml mlflow`
- `google-cloud-aiplatform zenml mlflow gcsfs`
- `google-cloud-aiplatform gcsfs`

## How to reproduce

``` shell
docker run --rm python:3.11.5 sh -c " \
curl -sSL https://install.python-poetry.org | python3 - --version=1.6.1 \
&& poetry --version \
&& mkdir app && cd app \
&& /root/.local/bin/poetry init --name=app --description=app --author=test@localhost --python='>=3.10,<3.12' -n \
&& /root/.local/bin/poetry add google-cloud-aiplatform kfp"
```

**Output**

```
Retrieving Poetry metadata

# Welcome to Poetry!

This will download and install the latest version of Poetry,
a dependency and package manager for Python.

It will add the `poetry` command to Poetry's bin directory, located at:

/root/.local/bin

You can uninstall at any time by executing this script with the --uninstall option,
and these changes will be reverted.

Installing Poetry (1.6.1)
Installing Poetry (1.6.1): Creating environment
Installing Poetry (1.6.1): Installing Poetry
Installing Poetry (1.6.1): Creating script
Installing Poetry (1.6.1): Done

Poetry (1.6.1) is installed now. Great!

To get started you need Poetry's bin directory (/root/.local/bin) in your `PATH`
environment variable.

Add `export PATH="/root/.local/bin:$PATH"` to your shell configuration file.

Alternatively, you can call Poetry explicitly with `/root/.local/bin/poetry`.

You can test that everything is set up by executing:

`poetry --version`

Creating virtualenv app-9TtSrW0h-py3.11 in /root/.cache/pypoetry/virtualenvs
Using version ^1.33.1 for google-cloud-aiplatform
Using version ^2.3.0 for kfp

Updating dependencies
Resolving dependencies...
```

dimbleby:
don't include first prerelease in the allowed max for an exclusive range.

Fixes https://github.com/python-poetry/poetry/issues/8475, https://github.com/python-poetry/poetry/issues/8405, https://github.com/python-poetry/poetry/issues/8202 

(actually 8202 seems to have been resolved by changes in the available dependencies on pypi and so I can't reproduce it now: but I'm pretty sure these will all be the same)

Absurdly I can't find a simple testcase that takes two versions or version ranges and compares or intersects them or something to demonstrate the point.  So I confess that I don't fully understand what was going wrong before this fix.

However I do understand that those issues are 
- something to do with wildcard constraints (eg `=2.*`) which poetry interprets as having a lower bound that is the first dev-release (eg `2.dev0`)
- and something to do with how they interact with regular but adjacent constraints like `<2`

so I've made this fix somewhat on gut: `<2` certainly shouldn't allow `2.dev0` and so it's weird that `allowed_max` would return that value.

What can I say?  I'm confused, but this looks like a good fix and it does resolve those issues...

### DIFF
--- a/src/poetry/core/constraints/version/version_range_constraint.py
+++ b/src/poetry/core/constraints/version/version_range_constraint.py
@@ -60,7 +60,7 @@ class VersionRangeConstraint(VersionConstraint):
         # The exclusive ordered comparison <V MUST NOT allow a pre-release
         # of the specified version unless the specified version is itself a pre-release.
         # https://peps.python.org/pep-0440/#exclusive-ordered-comparison
-        return self.max.first_devrelease()
+        return self.max.before_first_devrelease()
 
     def allows_lower(self, other: VersionRangeConstraint) -> bool:
         _this, _other = self.allowed_min, other.allowed_min
--- a/src/poetry/core/version/pep440/version.py
+++ b/src/poetry/core/version/pep440/version.py
@@ -297,6 +297,15 @@ class PEP440Version:
             pre=ReleaseTag(RELEASE_PHASE_ID_ALPHA),
         )
 
+    def before_first_devrelease(self: T) -> T:
+        return self.__class__(
+            epoch=self.epoch,
+            release=self.release,
+            pre=self.pre,
+            post=self.post,
+            dev=_NEG_INF_TAG,
+        )
+
     def first_devrelease(self: T) -> T:
         return self.__class__(
             epoch=self.epoch,


dimbleby:
I now think that the real problem - or anyway among the real problems - is ambiguity in poetry-core's treatment of the max in a strict inequality range.

- when dealing with a parsed value such as `<2` the max is `2` but the `allowed_max` should be `2.dev0`, because pre-releases are excluded by `<`
- however when dealing with a value calculated by operations between other versions or ranges, the `max` should be considered to be correct as it is, not to be revised by `allowed_max`
  - eg consider `(>1,<3).difference(2)`.  This should come out as `>1,<2 || >2,<3`: except that in the first part of that prereleases _should_ be allowed and so subsequent calculations that treat this as though it were `>1,<2.dev0` are wrong

A possible approach is to handle this at parsing and abandon `allowed_max` altogether.  ie `parse_single_constraint("<2")` should construct the object with `max=2.dev0` all along.  Then never use `allowed_max` at all.

This also solves those issues reported in poetry - but causes carnage in the test scripts: the main (but probably not only) problem being that parsing and then printing `<2` then gives `<2.dev0` - which should indeed be equivalent, but is not so pretty...

radoering:
> A possible approach is to handle this at parsing and abandon allowed_max altogether. ie parse_single_constraint("<2") should construct the object with max=2.dev0 all along. Then never use allowed_max at all.
>
> This also solves those issues reported in poetry - but causes carnage in the test scripts: the main (but probably not only) problem being that parsing and then printing `<2` gives `<2.dev0` - which should indeed be equivalent, but is not so pretty...

If we went down this path, maybe we should also change `__str__` so that `<2` and `<2.dev0` would always be printed as `<2`? I suppose explicit constraints like `<2.dev0` are rare so that complaints that `<2.dev0` gives `<2` are neglectable.

dimbleby:
that seems plausible and I have a branch in which I tried it.  Among the chaos in the unit tests I see at least these wrinkles

- lots of failures where the unit tests compare a parsed value to an explicitly constructed `Version` and now get a different answer: to be expected, but will be tedious to work through
- there's a distinction between version ranges as used in version requirements, and versions as used in markers / python versions.  So as not to break the world it's probably unwise to make changes for the latter two
- the good news is that printing `<2.dev0` as just `<2` is I think fine in the PEP440 context; because the rules about prereleases give these strings the same meaning
- this approach both fixes and introduces bugs when dealing with `!=` specifiers
  - eg `!=2` is represented internally as `<2 || >2`.  Today, the `allowed_max` business re-interprets that as `<2.dev0 || >2` ie it wrongly disallows the prereleases
  - eg after the change, `(>1).union(!=2)` for some reason becomes `>1,<2 || >2`.  Without `allowed_max` that's correct as an internal representation, but becomes wrong once re-parsed - it really does want to be written eg in lockfiles as `>1,!=2`

probably there's more that I haven't yet spotted

This feels like the right direction conceptually but I've no idea when or whether I'm likely to find the time or energy to do anything about it

eickr:
>I now think that the real problem - or anyway among the real problems - is ambiguity in poetry-core's treatment of the max in a strict inequality range.
> * when dealing with a parsed value such as `<2` the max is `2` but the `allowed_max` should be `2.dev0`, because pre-releases are excluded by `<`

Pre-releases must be excluded by `<` unless the specified version is itself a pre-release, yes. But development releases are pre-releases for any release, including pre-/post-releases.  
The versions `2aN.devN`, `2aN`, `2bN.devN`, `2bN`, `2rcN.devN`, `2rcN` are all less than `2.dev0`.  
If you wanted a pre-release allowed exclusive upper bound for the latest version starting with 1 or less then the exclusive ordered comparison equivalent syntax should be `<2a0.dev0`.

dimbleby:
You are mistaken, but thanks anyway

https://peps.python.org/pep-0440/#summary-of-permitted-suffixes-and-relative-ordering

radoering:
Superseded by #939

PR #663, opened 2023-11-14
treat collections in packages as immutable

dimbleby:
intended as a more-or-less maintainable version of #618.  See also https://github.com/python-poetry/poetry/pull/8671

by annotating collection types in packages as immutable, we "prove" that the shallow copy is a safe way to clone them

of course this is still not too hard to break: just by adding in something mutable later.

I see performance improvement during locking by about 20-25% on a couple of sample projects.  It's not a complete no-brainer to me that this is change is safe enough to accept, but I think it _probably_ is: and that's a good amount of performance to leave on the table.  Still, I'm ready to be overruled.


radoering:
I tend to say it's worth it.

> of course this is still not too hard to break: just by adding in something mutable later.

Maybe, we should add a comment in the `__init__` methods of `PackageSpecification`, `Package` and `Dependency` just before the attributes are defined. Something like "Attributes must be immutable so that clone() is safe! (For performance reasons, clone just creates a copy instead of a deepcopy.)" It's still not much but at least it's as close as possible to the place where the disaster will happen.

Is there a reason why you left `ProjectPackage` alone? It inherits from `Package` and has some additional mutable attributes. I did not check but I suppose it does not matter because we only have one `ProjectPackage` and probably don't care about cloning. If we want to be safe without spending too much effort we could also override `clone` in `ProjectPackage` to make a deepcopy. I don't have a preference yet...

dimbleby:
both of your suggestions sound reasonable: if you're active and interested then feel free to make updates directly else I'll try to get to it - but likely not for at least a few days

radoering (changes requested):
For transparency: I'm planning to merge this shortly before the core release because I think merging it now without a release will make development on other Poetry PRs that require core changes more difficult.

PR #670, opened 2023-12-02
Use `importlib.resources` to load schema and licenses

edgarrmondragon:
Again, this is one step towards resolving https://github.com/python-poetry/poetry/issues/2965.

Another attempt after https://github.com/python-poetry/poetry-core/pull/668 fell through.


src/poetry/core/json/__init__.py
```diff
@@ -9,7 +9,11 @@
 
 from fastjsonschema.exceptions import JsonSchemaException
 
+from poetry.core.json import schemas
+from poetry.core.utils._compat import importlib_resources
 
+
+# TODO: Remove this?
```
edgarrmondragon, line 16: We can put this behind a `__getattr__()` and raise a deprecation warning too, though I doubt anyone's using it?

vendors/pyproject.toml
```diff
@@ -22,6 +22,7 @@ classifiers = [
 python = "^3.8"
 
 fastjsonschema = "^2.18.0"
+importlib-resources = { version = ">=1.3", python = "<3.9" }
```
dimbleby, line 25: `<3.9` is not helpful here: importlib-resources has to be either vendored with poetry-core, or not.  

This is a decision that is made at `make vendor/update` time: it is not desirable that maintainers and contributors using different python versions get different answers

dimbleby (commented):
my overall feeling is that this is a lot of code to be vendoring for a rather niche use case - poetry in a zipapp - which, presumably, has been unsupported for all this time anyway.

Maybe wait until python 3.8 goes eol next year and then come back to this?

radoering:
I'd like to avoid additional vendoring, too. An alternative to waiting for the EOL of Python 3.8 might be to keep the old way for Python 3.8 and only use `importlib.resources` for Python 3.9 and above.

src/poetry/core/json/__init__.py
dimbleby: it is used in poetry itself

edgarrmondragon:
@radoering @dimbleby this is ready for another review 🙂

radoering (approved):

PR #674, opened 2023-12-12
Fix issue with creating new projects Windows

brycepg:
The previous code contained a bug where git couldn't reside in relative subdirectory.

It appears to have done this in attempt to get the absolute path of the git executable. Instead, I just use resolve to resolve potential symlinks, etc

The work around was to cd into a random folder so I don't really think it was adding much security, and the package manager
scoop installed git.exe at C:/Users/Bryce/scoop/shims/scoop.cmd

Resolves: python-poetry#7320


brycepg:
Do we want a regression test? There are plenty of tests for git overall

radoering:
> Do we want a regression test?

We always want a test if it is feasible. 😄

PR #675, opened 2023-12-16
Allow path dependency that is missing build-system

oysols:
Missing [build-system] table should be allowed.
Poetry handles this fine, except for this check.

> Tools should not require the existence of the [build-system] table.
https://packaging.python.org/en/latest/specifications/pyproject-toml/

---

### DIFF
--- a/src/poetry/core/packages/dependency.py
+++ b/src/poetry/core/packages/dependency.py
@@ -406,8 +406,7 @@ class Dependency(PackageSpecification):
             if os.path.isdir(p) and (os.path.sep in name or name.startswith(".")):
                 if not is_python_project(Path(name)):
                     raise ValueError(
-                        f"Directory {name!r} is not installable. File 'setup.[py|cfg]' "
-                        "not found."
+                        f"Directory {name!r} is not installable. Not a Python project."
                     )
                 link = Link(path_to_url(p))
             elif is_archive_file(p):
--- a/src/poetry/core/packages/utils/utils.py
+++ b/src/poetry/core/packages/utils/utils.py
@@ -18,7 +18,6 @@ from urllib.request import url2pathname
 from poetry.core.constraints.version import Version
 from poetry.core.constraints.version import VersionRange
 from poetry.core.constraints.version import parse_marker_version_constraint
-from poetry.core.pyproject.toml import PyProjectTOML
 from poetry.core.version.markers import SingleMarker
 from poetry.core.version.markers import SingleMarkerLike
 from poetry.core.version.markers import dnf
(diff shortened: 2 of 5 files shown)


dimbleby:
per the PEP, the correct check for `supports_pep517` would be

> If the pyproject.toml file is absent, or the build-backend key is missing, the source tree is not using this specification

the section that you quote relates to PEP518's `build-system`, not PEP517's `build-backend`

oysols:
I agree. This implements the simplified solution to this issue. The alternative would be deep validation of the toml file.

radoering:
I still don't understand what kind of problem you're trying to solve. Is it just that the name of the method does not match the current implementation or is there a deeper issue? (Afaik, a path dependency without a build system (and no setup.py/setup.cfg) is not installable from Poetry's point of view and thus is not supported so maybe we should just rename the method to `is_installable_python_project()`?)

Further, the following message is now even more outdated than before:

https://github.com/python-poetry/poetry-core/blob/56c03764e1359615adfcacab12ababd7c57a0e90/src/poetry/core/packages/dependency.py#L407-L411

oysols:
A path dependency without a build system IS installable by Poetry, but this check blocked the installation. I tried to describe the issue with the new test case. As I read the spec (as linked earlier) that pyproject.toml should be valid and installable.

dimbleby:
indeed according to `pip`'s reading, even an empty `pyproject.toml` is enough to define an installable package:
```console
$ pip install .
Processing /home/dch/foo
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Preparing metadata (pyproject.toml) ... done
Building wheels for collected packages: UNKNOWN
  Building wheel for UNKNOWN (pyproject.toml) ... done
  Created wheel for UNKNOWN: filename=UNKNOWN-0.0.0-py3-none-any.whl size=923 sha256=1c1ca2513e561152af25077228c0b0adcc3e7bf389b9423e5db0fa0b185a6ada
  Stored in directory: /tmp/pip-ephem-wheel-cache-jgkqf7p1/wheels/74/36/59/5f272a8602c0655288a982d06e0ec48e94c6c3ccfe9d1f1d69
Successfully built UNKNOWN
Installing collected packages: UNKNOWN
Successfully installed UNKNOWN-0.0.0
```

radoering:
> A path dependency without a build system IS installable by Poetry

Ah, of course it is since `setuptools` supports pyproject.toml. 🤦

Then, this makes sense of course and we should only change the error message I have already mentioned in https://github.com/python-poetry/poetry-core/pull/675#issuecomment-1859192332.

> indeed according to pip's reading, even an empty pyproject.toml is enough to define an installable package:

Weird...

radoering (approved):

PR #679, opened 2024-01-04
Make build formats dict openly available

Secrus:
- ~~Added **tests** for changed code.~~ No significant changes.

This allows for poetry plugins to easily register custom builders. Small change, I don't see any tests that would be needed for that.

### DIFF
--- a/src/poetry/core/masonry/builder.py
+++ b/src/poetry/core/masonry/builder.py
@@ -2,6 +2,9 @@ from __future__ import annotations
 
 from typing import TYPE_CHECKING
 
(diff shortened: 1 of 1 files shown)


radoering:
Do you have an example for such a plugin?

Secrus:
> Do you have an example for such a plugin?

A plugin for building distro-specific packages (`.deb`, `.rpm`) for example.

radoering:
Probably obsolete if we decide for https://github.com/python-poetry/poetry-core/pull/680#issuecomment-1879031775. I suppose it's better to have an interface for poetry plugins in somewhere in poetry itself instead of poetry-core anyway.

PR #680, opened 2024-01-04
Rename `Builder` to `BaseBuilder`

Secrus:
A small change of class name. While working on #679 It was confusing to have `poetry.core.masonry.builders.builder` and `poetry.core.masonry.builder` have the same name for 2 different classes.

### DIFF
--- a/src/poetry/core/masonry/builders/builder.py
+++ b/src/poetry/core/masonry/builders/builder.py
@@ -30,7 +30,7 @@ Summary: {summary}
 logger = logging.getLogger(__name__)
 
 
-class Builder:
+class BaseBuilder:
     format: str | None = None
 
     def __init__(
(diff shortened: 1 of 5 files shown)


bswck:
Thumbs up on that one, I got confused too.

Secrus:
Damn, the downstream...

radoering:
Probably requires a deprecation like this https://github.com/python-poetry/poetry/blob/1.3/src/poetry/repositories/pool.py

dimbleby:
another option - because `BaseBuilder` is such an ugly name! - would be simply to deprecate and remove the `poetry.core.masonry.builder` altogether

if I am right it is used only by the `poetry build` command. and it is so trivial that the code could reasonably just live in `poetry` proper all along: there's little value in `poetry-core` exposing a class for this.

PR #685, opened 2024-01-21
Fix encoding warnings with PEP 597 enabled

danyeaw:
Resolves: https://github.com/python-poetry/poetry/issues/8892

- [X] Updated **documentation** for changed code.

This PR goes along with https://github.com/python-poetry/poetry/pull/8893 to fix Encoding Warning with PEP 597 warnings enabled and enables the warnings in CI.


src/poetry/core/vcs/__init__.py
```diff
@@ -28,6 +28,7 @@ def get_vcs(directory: Path) -> Git | None:
                 [executable(), "rev-parse", "--show-toplevel"],
                 stderr=subprocess.STDOUT,
                 text=True,
+                encoding="utf-8",
```
dimbleby, line 31: why can we assume that git produces utf8 output?

danyeaw: @dimbleby Thanks for the feedback, I updated the encoding to locale.

radoering: Now, we get a warning in the tests. Shall we [suppress it](https://docs.python.org/3/library/warnings.html#temporarily-suppressing-warnings) in this case?

radoering: That seems to be a tricky one. On my Windows machine, this only works with "utf-8". Minimal example:

```
>>> subprocess.check_output('git rev-parse --show-toplevel')
b'C:/temp/r\xc3\xa9p\xc3\xb6\n'
>>> subprocess.check_output('git rev-parse --show-toplevel', text=True)
'C:/temp/rÃ©pÃ¶\n'
>>> subprocess.check_output('git rev-parse --show-toplevel', text=True, encoding='locale')
'C:/temp/rÃ©pÃ¶\n'
>>> subprocess.check_output('git rev-parse --show-toplevel', text=True, encoding='utf-8')
'C:/temp/répö\n'
```

But obviously in our CI it has to be "locale" (I added a test that succeeds on my machine but fails in our CI). I'm wondering if there is a universal solution?

radoering: OK, the issue with the test was not about encoding but about running the test alone and running all tests. There was another test that left the git access in a broken state and thereby made the new test fail.

To put it in a nutshell: no difference between my machine and CI. It looks like we can just assume utf-8, here.

danyeaw:
Thanks @radoering for helping out! Another option is that the encoding warnings and encoding="locale" are both only available in Python 3.10+. I could add a sys.version check and only explicitly pass in the encoding for this git command then.

radoering:
Actually, "locale" does not seem to be an option in this case. The new test fails on Windows (also on Python 3.10+) if I set the encoding to locale. Also see https://github.com/python-poetry/poetry-core/pull/685#discussion_r1468428840. Thus, in the end this PR revealed a real bug in the existing code.

radoering (approved):

danyeaw:
Thanks @radoering!

PR #690, opened 2024-01-27
deprecate scripts depending on extras

radoering:
Related to python-poetry/poetry#8902 and python-poetry/poetry#8903


src/poetry/core/factory.py
```diff
@@ -464,11 +464,20 @@ def validate(
                         continue
 
                     extras = script.get("extras", [])
+                    if extras:
+                        result["warnings"].append(
+                            f'The script "{name}" depends on an extra. Scripts'
+                            " depending on extras are deprecated and support for them"
+                            " might be removed in a future version of"
```
Secrus, line 471:
```suggestion
                            " will be removed in a future version of"
```
Let's avoid "maybe"s in deprecations. Especially since this will be changed with PEP 621 support.

src/poetry/core/masonry/builders/builder.py
```diff
@@ -311,7 +311,7 @@ def convert_entry_points(self) -> dict[str, list[str]]:
                     f"Use of callable in script specification ({name}) is"
```
Secrus, line 311: I was thinking we could remove this deprecated feature and make this method a bit cleaner. It has been here for a long time now.

src/poetry/core/masonry/builders/builder.py
```diff
@@ -322,6 +322,16 @@ def convert_entry_points(self) -> dict[str, list[str]]:
                 continue
 
             extras = specification.get("extras", [])
+            if extras:
+                warnings.warn(
+                    f'The script "{name}" depends on an extra. Scripts depending on'
+                    " extras are deprecated and support for them might be removed in a"
```
Secrus, line 328:
```suggestion
                    " extras are deprecated and support for them will be removed in a"
```
Same as above.

tests/masonry/builders/fixtures/licenses_and_copying/pyproject.toml
```diff
@@ -43,7 +43,6 @@ time = ["pendulum"]
 [tool.poetry.scripts]
 my-script = "my_package:main"
 my-2nd-script = "my_package:main2"
-extra-script = {reference = "my_package.extra:main", extras = ["time"], type = "console"}
```
Secrus, line 46: How about we deprecate the dict/table syntax all together? With extras deprecation, it makes no sense to keep it around, since we don't support any other type than `console`.

Secrus (changes requested):

src/poetry/core/masonry/builders/builder.py
radoering: I'd probably play it safe and keep it for now. (I feel like we have to be extra cautious in poetry-core (compared to poetry) because it's the backend. Remember breaking isort pre-commit hooks? 😅) Further, I'm planning to focus on PEP 621 support after the next release. Even if we try to make the transition as smooth as possible, I fear we might introduce some unintended breaking changes. Long story short, I think the version after the next release might be a major version update (if I can manage PEP 621 support) and that will be a good opportunity to remove most of the deprecated stuff.

Secrus: Ok, fine

tests/masonry/builders/fixtures/licenses_and_copying/pyproject.toml
Secrus: Sounds good to me.

Secrus (approved):

PR #692, opened 2024-01-28
handle deprecation of poetry.core.masonry.builder.Builder

dimbleby:
fixing deprecation warnings in the unit tests

### DIFF
--- a/tests/masonry/builders/test_complete.py
+++ b/tests/masonry/builders/test_complete.py
@@ -18,7 +18,8 @@ import pytest
 
 from poetry.core import __version__
 from poetry.core.factory import Factory
-from poetry.core.masonry.builder import Builder
+from poetry.core.masonry.builders.sdist import SdistBuilder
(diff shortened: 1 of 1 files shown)


radoering (approved):

moto-timo:
I think this breaks the Yocto Project usage, where we do not want to build the full poetry but DO want to support the poetry-core PEP-517 backend.
https://git.yoctoproject.org/poky/tree/meta/classes-recipe/python_poetry_core.bbclass
https://git.yoctoproject.org/poky/tree/scripts/lib/recipetool/create_buildsys_python.py#n738

moto-timo:
Nevermind. A colleague pointed out that the poetry.core.masonry.builders.wheel is still there and I noticed nothing seems to have changed the "poetry.core.masonry.api" build-backend callout. I tested a build of tomlkit (and ran it's tests) and it was JustFine(TM).

PR #695, opened 2024-02-05
Error out when poetry -C argument is not a pyproject directory

Cypher1:
Poetry's current handling of the `--directory` (or `-C`) argument does not protect against incorrectly specified paths (e.g. misspellings and simple mistakes).
In mono repo projects, and other cases where a parent directory has a `pyproject.toml` this can lead to commands being run in an unexpected package environment, leading to error prone and difficult to catch behaviour in multi package repositories.

This change retains the default behavior of `poetry` when no `-C` argument is specified (i.e. walking up the tree from the current directory) but ensures that the `-C` argument is valid when it is manually specified.

This prevents incorrectly typed directories from breaking builds & CI scripts, while retaining the current default behaviour, minimising change.

Some tests are also updated where they previously pass paths to `pyproject.toml` files to the `cwd` arg, (instead of project directories).


Cypher1:
@radoering would you be free to review this?

radoering:
I am currently busy preparing Poetry 1.8. I'll take a look after that's done. (The corresponding poetry-core has already been released at the beginning of February so you caught an unlucky timing.)

Cypher1:
No problem at all. Thx!

radoering:
We have to be careful not to accidentally introduce a breaking change and therefore evaluate all uses of `create_poetry()` with `cwd != None` in poetry-core and poetry:

usage|evaluation|remark
---|---|---
[poetry.core.masonry.api](https://github.com/python-poetry/poetry-core/blob/219c52e6cb8b086c78f25bbfa1168d0a3096fb31/src/poetry/core/masonry/api.py#L42) (4x)|⚠️❓| That could be a dangerous one, since it's the API. I'm not sure if we can assume that poetry-core is never called from a subdirectory.
[poetry.console.application](https://github.com/python-poetry/poetry/blob/120e7373035c2f1580c51b7b80eefda62010a1f4/src/poetry/console/application.py#L124-L134)|⚠️|If `directory` is not set, we explicitly pass the cwd. I think this PR will introduce an unwanted change in this case. However, we can make a downstream change to just pass `None` instead.
[poetry.console.commands.self.self_command](https://github.com/python-poetry/poetry/blob/120e7373035c2f1580c51b7b80eefda62010a1f4/src/poetry/console/commands/self/self_command.py#L84-L88)|✅|This one should be safe since we pass the directory of the pyproject.toml.
[poetry.inspection.info](https://github.com/python-poetry/poetry/blob/120e7373035c2f1580c51b7b80eefda62010a1f4/src/poetry/inspection/info.py#L476-L478)|✅|This one should be safe since we pass the directory of the pyproject.toml.
[poetry.installation.executor](https://github.com/python-poetry/poetry/blob/120e7373035c2f1580c51b7b80eefda62010a1f4/src/poetry/installation/executor.py#L705-L707)|✅|This one should be safe since we pass the directory of the pyproject.toml.

Cypher1:
@radoering How can AI help with these concerns? I think the current state of the `-C` argument is not ideal and leads to significant issues when maintaining a large (i.e. multi-package) mono repo, still I am totally on board with backwards compatibility's important, even when the current implementation does not 100% match the documented (and arguably expected) behaviour.

Looking forward to hear more about the project's needs on this front.
Cheers,
Jay

abn:
Don't think this is the correct spot for handling an input validation from the poetry cli. 

This should be handled at https://github.com/python-poetry/poetry/blob/d08c75afcc097b3b6a73c704f64f7cebbc2765c1/src/poetry/console/application.py#L123-L131

Cypher1:
Sorry @abn , missed your Mar 5th response, that sounds great! I'll have a go when I'm able

PR #699, opened 2024-02-16
convert_script_files does not require script to exist

dimbleby:
Scripts can be generated during build.  But poetry-core insists that scripts exist ahead of time.

fixes <https://github.com/python-poetry/poetry/issues/8856>

obviously this is giving up some validation: it is now possible to build a wheel that points at a non-existent script.  But
- I do not see an easy way of getting that back
- probably it is pretty obvious when that happens
- it already is possible to point at non-existent code in an entrypoint-style script, so in some sense this just puts things on an equal footing...

### DIFF
--- a/src/poetry/core/masonry/builders/builder.py
+++ b/src/poetry/core/masonry/builders/builder.py
@@ -364,16 +364,6 @@ class Builder:
 
                 abs_path = Path.joinpath(self._path, source)
 
-                if not abs_path.exists():
-                    raise RuntimeError(
-                        f"{abs_path} in script specification ({name}) is not found."
-                    )
-
-                if not abs_path.is_file():
-                    raise RuntimeError(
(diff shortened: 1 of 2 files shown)


src/poetry/core/masonry/builders/builder.py
```diff
@@ -364,16 +364,6 @@ def convert_script_files(self) -> list[Path]:
 
                 abs_path = Path.joinpath(self._path, source)
 
-                if not abs_path.exists():
```
radoering, line 367: We could put this under `if not self._package.build_script:` to keep the validation. Without a build script I assume the files must exist at this point.

weisslj: In #710 I am doing this, with added test case,  @dimbleby encouraged me to take over. Please review :pray:

radoering:
Superseded by #710

PR #701, opened 2024-02-24
normalize source vcs urls

abn:
This change ensures when referencing source url for package specifications of source type "git" or for VCS dependencies, a normalized version returned at all times.

Relates-to: python-poetry/poetry#8999

This change should ensure that we write out a valid url for cases where `git@github.com:org/project.git` us used as source url.

### DIFF
--- a/src/poetry/core/packages/specification.py
+++ b/src/poetry/core/packages/specification.py
@@ -35,7 +35,7 @@ class PackageSpecification:
         self._pretty_name = name
         self._name = canonicalize_name(name)
         self._source_type = source_type
-        self._source_url = source_url
+        self._source_url = self._normalize_source_url(source_type, source_url)
         self._source_reference = source_reference
         self._source_resolved_reference = source_resolved_reference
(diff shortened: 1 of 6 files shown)


radoering:
FYI: According to https://secure.phabricator.com/T11004, `git@example.org:demo/demo.git` and `ssh://git@example.org/demo/demo.git` might not be equivalent in general (for other hosts than github), but maybe we can assume it anyway. 🤷  I mentioned this at https://github.com/pypa/packaging.python.org/pull/1506#issuecomment-1962270380, too.

abn:
I considered this. While it is a bit unclear how reliably this is true, we already make this assumption in the code elsewhere. 

https://github.com/python-poetry/poetry-core/blob/219c52e6cb8b086c78f25bbfa1168d0a3096fb31/src/poetry/core/packages/vcs_dependency.py#L111-L115

So, I suspect, for right now this is an alright assumption.

Secrus (approved):

radoering (approved):

PR #702, opened 2024-02-25
follow through on deprecations

dimbleby:
I believe all these deprecations were made in 2022 or earlier

one concession that I have made is leaving `callable` in [the schema](https://github.com/python-poetry/poetry-core/blob/71d26d7f90ed7295a50ab0d735f3cd5eb20773c5/src/poetry/core/json/schemas/poetry-schema.json#L645-L647).  The code will no longer pay any attention to this - even to issue a deprecation warning.  But I guess that there are still `pyproject.toml` out there that include it and that do not bound their build backend


abn (approved):

eamanu:
Hi, @dimbleby  This test shouldn't be updated as well? https://github.com/python-poetry/poetry-core/blob/main/tests/masonry/builders/test_builder.py#L305

script_callable_legacy_table/ does not exists

dimbleby:
you are eleven months too late to leave useful comments

if you have an improvement, submit a pull request

radoering:
> This test shouldn't be updated as well?

Looks like you are right. PR welcome.

PR #704, opened 2024-02-26
get_vcs: support Cygwin git under Windows

lazka:
(I haven't created an issue since the issue template suggested to skip it, but I can create one if wanted)

In MSYS2 we currently use native Python in combination with cygwin git, the former working with Windows paths, while the latter returns Unix paths, which means get_vcs() will find to find a git repo, but with a bogus git directory.

It's quite easy to fix by using git in a way so it returns relative paths, without adding any special cases.

1) Improve the test coverage of get_vcs() to make sure we don't break things
2) Switch git to use relative paths instead of absolute paths
3) Some further cleanup to not change the working dir, while at it.

See the separate commits for details.


abn (approved):

lazka:
thanks!

PR #710, opened 2024-03-15
convert_script_files does not require script to exist

weisslj:
Scripts can be generated during build. But poetry-core insists that scripts exist ahead of time.

Resolves: python-poetry/poetry#8856

This is a new merge request based on #699, with suggestion from @radoering applied, @dimbleby encouraged me to take over.

- [x] Updated **documentation** for changed code. **Not necessary I think**

cc @bavalpey


tests/masonry/builders/test_wheel.py
```diff
@@ -451,3 +451,16 @@ def test_build_py_only_included() -> None:
 
     with zipfile.ZipFile(str(whl)) as z:
         assert "build_with_build_py_only/generated/file.py" in z.namelist()
+
+
+def test_generated_script_file() -> None:
+    """Tests that a file that is generated by build.py can be used as script."""
+    root = fixtures_dir / "generated_script_file"
```
radoering, line 458: After executing the test once, the generated file is in the source tree and the test succeeds even without the change. Can you please use the `tmp_path` fixture, `shutil.copytree()` the project and work on the temporary path?

radoering (changes requested):
LGTM except for a small detail regarding the test.

tests/masonry/builders/test_wheel.py
weisslj: @radoering thanks for the suggestion! Done in a380348.

radoering (approved):

PR #714, opened 2024-03-26
typing: Mapping is sufficient for DependencyConstraint

radoering:
Noticed while working on PEP 621 support.

### DIFF
--- a/src/poetry/core/factory.py
+++ b/src/poetry/core/factory.py
@@ -2,10 +2,10 @@ from __future__ import annotations
 
 import logging
 
+from collections.abc import Mapping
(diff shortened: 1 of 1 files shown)


src/poetry/core/factory.py
```diff
@@ -2,10 +2,10 @@
 
 import logging
 
+from collections.abc import Mapping
```
abn, line 5: Shouldn't this be in type checking?

radoering: No, because it's used in an instance check.

abn (approved):

PR #715, opened 2024-03-27
Add build config settings to core

earonesty:
The build() function takes optional config-settings which can be useful for building gpu-specific packages, editable-mode settings for setuptools, or other build-tool specific configurations.   If used, should resolve a number of issues.   For example: https://github.com/python-poetry/poetry/issues/8909

(Sorry for the million commits, im editing this in a browser.)

See: https://build.pypa.io/en/stable/api.html.   Typing matches that.

If there's no objections, will finish adding tests, etc.

### DIFF
--- a/src/poetry/core/factory.py
+++ b/src/poetry/core/factory.py
@@ -280,7 +280,7 @@ class Factory:
                     groups=groups,
                     optional=optional,
                     develop=constraint.get("develop", False),
-                    config_settings=constraint.get("config_settings", {}),
+                    config_settings=constraint.get("config-settings", {}),
                     extras=constraint.get("extras", []),
                 )
             elif "file" in constraint:
@@ -323,7 +323,7 @@ class Factory:
                         optional=optional,
                         base=root_dir,
                         develop=constraint.get("develop", False),
-                        config_settings=constraint.get("config_settings", {}),
+                        config_settings=constraint.get("config-settings", {}),
                         extras=constraint.get("extras", []),
                     )
             elif "url" in constraint:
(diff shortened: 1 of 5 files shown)


radoering:
> If there's no objections, will finish adding tests, etc.

I think, we are open for this setting. Please also create a downstream (draft) PR that shows the usage of this setting. (You can replace the poetry-core dependency with a git dependency pointing to your branch in a temporary commit that will be dropped before merging.)

radoering (changes requested):
I suppose, this setting is relevant for all types of dependencies because it's not just relevant for directories but also for sdists?

earonesty:
Yes.

igor-sadchenko:
Hello @earonesty 
Any updates here?
Maybe do you need help with this PR?

earonesty:
I never finished the other side of it.  This PR is complete from what I can
tell

abn:
@earonesty apologies for getting to this sooner. Really appreciate your effort in getting this proposed. I have proposed python-poetry/poetry#10129 which will close this out unfortunately.

While I can see the value add in having this configurable in the dependency spec, I do not think this is the appropriate solution for this as things stand. By the nature of these config settings, they are not universal in most cases. A configuration that works for a user on linux machine might not work for someone on an osx machine (or vice-versa) etc. 

This is why python-poetry/poetry#10129 proposes the introduction of a configuration setting instead of an addition to the `pyproject.toml` file. 

Yes, there are some cases where it might make sense to have a "default" config for every developer (think corporate environment with uniform developer environments). For these cases, I propose that users simply commit a `poetry.toml` with the config to their project until we stablise the config and maybe propose a "standard" way to provide project specific configuration without compromising the ability for users to override their local and/or project local config.

Please also let me know if there are areas that the proposal falls short on for your original use case.

Once again, thank you for your contributions.

PR #716, opened 2024-03-28
Migrate anonymous Mapping[str, Any] to TypedDicts where their content is fully defined.

dylan-robins:
When developing a plugin for Poetry, I found that despite the "include", "exclude" and "package" mapping content being fully defined in the documentation, no type hints were available to help me. Therefore I propose this minor change. It has no impact on current functionality, however with this IDEs like PyCharm and VSCode are aware of the structure of these dicts.

- [ ] Added **tests** for changed code. **NOT APPLICABLE**
- [ ] Updated **documentation** for changed code. **NOT APPLICABLE** (I think)

**Other similar changes may be useful elsewhere in the code**. I'll continue to explore the codebase and see what I can do.

### DIFF
--- a/src/poetry/core/packages/project_package.py
+++ b/src/poetry/core/packages/project_package.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 import warnings
 
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Literal, TypedDict
 from typing import Any
 from typing import Mapping
 from typing import Sequence
@@ -17,6 +17,18 @@ if TYPE_CHECKING:
 
 from poetry.core.packages.package import Package
 from poetry.core.packages.utils.utils import create_nested_marker
+
+SupportedPackageFormats = Literal["sdist", "wheel"]
+
+class PackageSpec(TypedDict):
+    include: str
+    to: str
+    format: list[SupportedPackageFormats]
+
+
+class IncludeSpec(TypedDict):
+    path: str
+    format: list[SupportedPackageFormats]
 
 
 class ProjectPackage(Package):
@@ -40,9 +52,9 @@ class ProjectPackage(Package):
         # (For performance reasons, clone only creates a copy instead of a deep copy).
 
         self.build_config: Mapping[str, Any] = {}
-        self.packages: Sequence[Mapping[str, Any]] = []
-        self.include: Sequence[Mapping[str, Any]] = []
-        self.exclude: Sequence[Mapping[str, Any]] = []
+        self.packages: Sequence[PackageSpec] = []
+        self.include: Sequence[IncludeSpec] = []
+        self.exclude: Sequence[IncludeSpec] = []
         self.custom_urls: Mapping[str, str] = {}
 
         if self._python_versions == "*":
--- a/src/poetry/core/packages/project_package.py
+++ b/src/poetry/core/packages/project_package.py
@@ -18,7 +18,12 @@ if TYPE_CHECKING:
 from poetry.core.packages.package import Package
 from poetry.core.packages.utils.utils import create_nested_marker
 
+
 SupportedPackageFormats = Literal["sdist", "wheel"]
+
+
+BuildConfigSpec = TypedDict("BuildConfigSpec", {"script": str, "generate-setup-file": bool})
+
 
 class PackageSpec(TypedDict):
     include: str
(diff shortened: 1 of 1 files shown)


dylan-robins:
I do not know why the integration tests are failing, I've added typing_extensions as a dependency to the main pyproject.toml however the integration tests still indicate the following:

```
  File "/home/robinsdy/poetry-core/src/poetry/core/masonry/api.py", line 12, in <module>
    from poetry.core.factory import Factory
  File "/home/robinsdy/poetry-core/src/poetry/core/factory.py", line 15, in <module>
    from poetry.core.packages.project_package import BuildConfigSpec
  File "/home/robinsdy/poetry-core/src/poetry/core/packages/project_package.py", line 11, in <module>
    from typing_extensions import NotRequired
ModuleNotFoundError: No module named 'typing_extensions'
```

I am not entirely sure what this is, I'll look further into how your tests are constructed tomorrow but if anyone else wants to sort this out that'd be great :)

radoering:
In general, our stance is that `poetry-core` (as a build backend) should not have any dependencies. Since `typing-extensions` is a small package we might think about vendoring it. However, I'm not sure if it's worth it and would like to hear more opinions.

@dimbleby Since you are familiar with vendoring and contributing many typing improvements, I'd appreciate your opinion.

dimbleby:
In https://github.com/python-poetry/poetry/pull/9251 we have established that so long as all imports of type-extensions can be moved behind `if TYPE_CHECKING:`, it is not actually necessary to have type-extensions installed in the environment - and therefore not necessary to depend on it.

Over in poetry I have objected to this approach: it feels fragile and non-obvious and who cares about another dependency anyway?

Over here it still feels fragile and non-obvious - but the case for wanting to avoid the dependency is stronger.  

So that is an option.

dylan-robins:
Ah I see, thanks for the info @dimbleby. I also don't like this approach, but I do see the reasoning behind it.

Another side-effect of this that I don't like is that I've had to add `if TYPE_CHECKING`-guarded casts inside `poetry.core.factory.Factory.configure_package`:

https://github.com/python-poetry/poetry-core/blob/c47aad19d42793ccc7bb466790936230f115c635/src/poetry/core/factory.py#L202C1-L210C41

I don't like this at all, like you said it seems non-obvious and *very* fragile. Any ideas for how to make mypy happy without having this ugliness in the code?

dimbleby:
Iirc you can cast to the string `"Foo"` rather than the type `Foo`

src/poetry/core/factory.py
```diff
@@ -197,9 +200,11 @@ def configure_package(
 
         if "build" in config:
             build = config["build"]
+            if not build:
+                build = {}
```
radoering, line 204:
```suggestion
            build = config.get("build", {})
```
?

src/poetry/core/masonry/builders/builder.py
```diff
@@ -66,7 +66,7 @@ def _module(self) -> Module:
                 formats = [formats]
 
             if (
-                formats
+                len(formats) > 0
```
radoering, line 69: Does that make a difference? Isn't the former more Pythonic?

src/poetry/core/packages/project_package.py
```diff
@@ -3,20 +3,38 @@
...
+    )
+
+    class PackageSpec(TypedDict):
+        include: str
+        to: str
+        format: list[SupportedPackageFormats]
+
+    class IncludeSpec(TypedDict):
+        path: str
+        format: list[SupportedPackageFormats]
```
radoering, line 37: Since these were `Mappings` before and thus immutable for a reason (see comment in `ProjectPackage.__init__`), we probably should declare all attributes as [ReadOnly](https://typing.readthedocs.io/en/latest/spec/typeddict.html#typing-readonly-type-qualifier).

src/poetry/core/packages/project_package.py
```diff
@@ -3,20 +3,38 @@
...
+
+    BuildConfigSpec = TypedDict(
+        "BuildConfigSpec",
+        {"script": NotRequired[str], "generate-setup-file": NotRequired[bool]},
+    )
+
+    class PackageSpec(TypedDict):
+        include: str
+        to: str
+        format: list[SupportedPackageFormats]
```
radoering, line 33: Should probably be a `Sequence`.

src/poetry/core/packages/project_package.py
radoering, line 37: Should probably be a `Sequence`.

radoering (changes requested):

PR #724, opened 2024-04-26
Project tidying

dimbleby:
bits and pieces of tidying in pyproject.toml and dependencies

- make ruff configuration consistent with the main project
- make mypy configuration consistent with the main project
- ~remove unused fixtures, and unused dependency `virtualenv` (though it shows up via `pre-commit` anyway)~


dimbleby:
apparently I forgot `pytest --integration`, so `virtualenv` and the fixtures had better go back

dimbleby:
guessing the pipeline failure is because github have changed something about the environment that they are providing for macos, and the CI cache does not spot the difference

Secrus (approved):

Secrus (approved):

PR #734, opened 2024-05-22
Add error message specifying that we can't parse pyproject.toml

mscheifer:
Resolves: python-poetry#9394 (Mostly)

## Problem

Poetry gives confusing error messages when there is a parse error in the pyproject.toml file.

For example:
```toml
[tool.poetry]
name = "test-project"
version = "0.0.1"
description = ""
authors = []

[tool.poetry.dependencies]
python = "~3.11"
numpy = "^1.26.3"
numpy = "^1.26.3" # <-- Duplicate specification

[build-system]
requires = ["poetry-core>=1.0.0"]
build-backend = "poetry.core.masonry.api"
```
```
$ poetry run ruff check

Cannot overwrite a value (at line 10, column 18)
```
The error is confusing because it's unclear that this error relates to parsing the pyproject.toml file. For `poetry run` specifically it's also unclear if the error is coming from poetry or the command you are running.

## Change

With this change instead we will see.
```
$ poetry run ruff check

poetry cannot parse /home/matthew/code/poetry/pyproject.toml
```

And you can see the actual error with -v
```
$ poetry run -v ruff check
  TOMLDecodeError

  Cannot overwrite a value (at line 10, column 18)

  at ~/.cache/pypoetry/virtualenvs/poetry-z8mHJNjZ-py3.10/lib/python3.10/site-packages/tomli/_parser.py:349 in key_value_rule
      345│         nest = out.data.get_or_create_nest(abs_key_parent)
      346│     except KeyError:
      347│         raise suffixed_err(src, pos, "Cannot overwrite a value") from None
      348│     if key_stem in nest:
    → 349│         raise suffixed_err(src, pos, "Cannot overwrite a value")
      350│     # Mark inline table and array namespaces recursively immutable
      351│     if isinstance(value, (dict, list)):
      352│         out.flags.set(header + key, Flags.FROZEN, recursive=True)
      353│     nest[key_stem] = value

The following error occurred when trying to handle this error:


  PyProjectException

  poetry cannot parse /home/matthew/code/poetry/pyproject.toml

  at ~/.cache/pypoetry/virtualenvs/poetry-z8mHJNjZ-py3.10/lib/python3.10/site-packages/poetry/core/pyproject/toml.py:37 in data
       33│                         self._data = tomllib.load(f)
       34│                 except tomllib.TOMLDecodeError as e:
       35│                     from poetry.core.pyproject.exceptions import PyProjectException
       36│ 
    →  37│                     raise PyProjectException(
       38│                         f"poetry cannot parse {self._path.as_posix()}"
       39│                     ) from e
       40│ 
       41│         return self._data
```

## Open questions
Should we try to include the toml parse error at 0 verbosity somehow? Or alternatively, we could explicitly say "re-run with -v" in the message.

Please let me know if anything should be changed to fit the project code standards.


src/poetry/core/pyproject/toml.py
```diff
@@ -28,8 +28,15 @@ def data(self) -> dict[str, Any]:
...
-                with self.path.open("rb") as f:
-                    self._data = tomllib.load(f)
+                try:
+                    with self.path.open("rb") as f:
+                        self._data = tomllib.load(f)
+                except tomllib.TOMLDecodeError as e:
+                    from poetry.core.pyproject.exceptions import PyProjectException
+
+                    raise PyProjectException(
+                        f"poetry cannot parse {self._path.as_posix()}"
```
radoering, line 38: I do not like the wording. Mentioning "poetry" does not give any useful information (and strictly speaking it is not completely correct because we are in `poetry-core`, which can be used as a standalone build backend).

I'd propose either just `"Could not parse ..."`, which is used in other places or maybe better `"... is not a valid TOML file"` or similar so that it is clear that not just Poetry could not parse the toml file but it is just invalid in general.

> Should we try to include the toml parse error at 0 verbosity somehow?

I think so, for example `"... is not a valid TOML file: {e}"`

radoering (changes requested):

src/poetry/core/pyproject/toml.py
luckydonald: Yes, it does. Doing `poetry run python big_project.py`  it's very helpful to know it's not that complex 1000 files web app, but instead the package manager failing.

luckydonald:
I think the `(at line 10, column 18)`  part is equally helpful.

I'd even go as far as to suggest the most likely cause, and how to fix it, maybe similar to:

``` poetry-core cannot parse toml file at '/home/luckydonald/code/poetry/pyproject.toml'.
  Error: Cannot overwrite a value (at line 10, column 18)
This is often caused by a duplicate entry.
```

radoering:
> This give everyone a clear cause to look out for.
> Yes, that might "break" if the text changes, but if it does, it's will be just as cryptic as before, nothing lost.

That's ok with me.

src/poetry/core/pyproject/toml.py
mscheifer: >I think so, for example "... is not a valid TOML file: {e}"

Ahh cool, that's pretty elegant.

I do think there's value in mentioning that `poetry` (or `poetry-core`) is what has encountered the error, because of the case @luckydonald mentioned. You won't necessarily know which tool to pass `-v` too. (`poetry run -v python big_project.py` vs `poetry run python big_project.py -v`) But if if all the information to see what the problem is is there are 0 verbosity then this is less of a big deal.

Here's an example of another error message that mentions poetry: https://github.com/python-poetry/poetry-core/blob/94f933f50ade83cb7f13a2c521477aa4b8a0b1ee/src/poetry/core/factory.py#L494

mscheifer:
> Couldn't find a good way to check for that duplication from the exception alone. I'd just do
> 
> ```python
> if e.msg = 'Cannot overwrite a value':
>   txt += "\nThis is often caused by a duplicate entry."
> ```

Done. Getting the message out of a `ValueError` (which `TOMLDecodeError` is a thin subclass of) is a bit less clean though. You have to do `args[0]` as far as I can tell.

> This give everyone a clear cause to look out for. Yes, that might "break" if the text changes, but if it does, it's will be just as cryptic as before, nothing lost.

The unit test would also fail, so it would be caught and updated eventually.

src/poetry/core/pyproject/toml.py
```diff
@@ -28,8 +28,18 @@ def data(self) -> dict[str, Any]:
...
             else:
-                with self.path.open("rb") as f:
-                    self._data = tomllib.load(f)
+                try:
+                    with self.path.open("rb") as f:
+                        self._data = tomllib.load(f)
+                except tomllib.TOMLDecodeError as e:
+                    from poetry.core.pyproject.exceptions import PyProjectException
+
+                    msg = f"{self._path.as_posix()} is not a valid TOML file. {e}"
```
radoering, line 37:
```suggestion
                    msg = (
                        f"{self._path.as_posix()} is not a valid TOML file.\n"
                        f"{e.__class__.__name__}: {e}"
                    )
```

I think, I'd even print the name of the exception.

src/poetry/core/pyproject/toml.py
```diff
@@ -28,8 +28,18 @@ def data(self) -> dict[str, Any]:
...
+
+                    if e.args[0].startswith("Cannot overwrite a value"):
+                        msg += "\nThis is often caused by a duplicate entry"
```
radoering, line 40:
```suggestion
                    if str(e).startswith("Cannot overwrite a value"):
                        msg += "\nThis is often caused by a duplicate entry."
```

`str` should be more robust. It is unlikely but `args` could be empty.

radoering (changes requested):

tests/pyproject/test_pyproject_toml.py
```diff
@@ -72,3 +72,20 @@ def test_pyproject_toml_non_existent(pyproject_toml: Path) -> None:
...
+        / "fixtures"
+        / "project_duplicate_dependency"
+        / "pyproject.toml"
+    )
+
+    with pytest.raises(PyProjectException) as excval:
+        _ = PyProjectTOML(pyproject_toml).build_system
+
+    assert (
+        f"{pyproject_toml.as_posix()} is not a valid TOML file. Cannot overwrite a value (at line 7, column 16)\nThis is often caused by a duplicate entry"
```
mscheifer, line 89:
```suggestion
        f"{pyproject_toml.as_posix()} is not a valid TOML file\nTOMLDecodeError: Cannot overwrite a value (at line 7, column 16)\nThis is often caused by a duplicate entry"
```

tests/pyproject/test_pyproject_toml.py
```diff
@@ -72,3 +72,20 @@ def test_pyproject_toml_non_existent(pyproject_toml: Path) -> None:
...
+        / "pyproject.toml"
+    )
+
+    with pytest.raises(PyProjectException) as excval:
+        _ = PyProjectTOML(pyproject_toml).build_system
+
+    assert (
+        f"{pyproject_toml.as_posix()} is not a valid TOML file.\nTOMLDecodeError: Cannot overwrite a value (at line 7, column 16)\nThis is often caused by a duplicate entry"
+        in str(excval.value)
+    )
```
radoering, line 91:
```suggestion
    assert (
        f"{pyproject_toml.as_posix()} is not a valid TOML file.\n"
        "TOMLDecodeError: Cannot overwrite a value (at line 7, column 16)\n"
        "This is often caused by a duplicate entry"
    ) in str(excval.value)
```

radoering (approved):

maitreyakv:
Thanks for this @mscheifer!

PR #739, opened 2024-06-24
Add monthly downloads badge to README

Aman-Clement:
Added a monthly downloads badge by pip Trends to the README.
View more at - https://piptrends.com/widgets/poetry-core

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,7 @@
 [![Python Versions](https://img.shields.io/pypi/pyversions/poetry-core)](https://pypi.org/project/poetry-core/)
 [![License: MIT](https://img.shields.io/badge/License-MIT-yellow.svg)](https://opensource.org/licenses/MIT)
 [![](https://github.com/python-poetry/poetry-core/workflows/Tests/badge.svg)](https://github.com/python-poetry/poetry-core/actions?query=workflow%3ATests)
+[![poetry-core Downloads Last Month](https://assets.piptrends.com/get-last-month-downloads-badge/poetry-core.svg 'poetry-core Downloads Last Month by pip Trends')](https://piptrends.com/package/poetry-core)
 
 A [PEP 517](https://www.python.org/dev/peps/pep-0517/) build backend implementation developed for
 [Poetry](https://github.com/python-poetry/poetry). This project is intended to be a lightweight, fully compliant,


ixmatus:
This person tried to open this exact PR against a project I'm the maintainer for as well (https://github.com/ixmatus/inflector/pull/16). This is a promotional campaign and they (or colleagues) appear to be doing it on reddit too, I think this is suspicious activity. They are creating an external dependency on a web property that (as far as I can tell) is not officially blessed by pypi. I recommend rejecting this PR.

Users can get package statistics from `libraries.io` which is linked to directly by pypi itself.

edgarrmondragon:
Yeah I don't see how this is better than shields.io:

PyPI - Downloads

ayushjain01:
Thank you for your feedback and concerns. I'd like to clarify that this is not a promotional campaign. As stated in the PR description, our intention is simply to share a tool we have developed, which we believe could be beneficial to the community.

Our badge system offers highly accurate statistics, which is why we wanted to make it available. If the link to the pip trends page is a concern, we are more than willing to remove it.

We respect your decision if you prefer not to add more badges to your README. Our primary goal is to address any misunderstandings and assure you that this is a genuine effort to contribute to the open-source community by sharing our work on GitHub.

Thank you for considering our contribution.

ayushjain01:
> Yeah I don't see how this is better than shields.io:
> 
> PyPI - Downloads


Here's how our badge is better than the existing badges - 
Sheilds.io - PyPI - Downloads
pepy.tech - [Downloads](https://pepy.tech/project/poetry-core)
pip trends -[poetry-core Downloads Last Month](https://piptrends.com/package/poetry-core)


You can see the difference in the download counts on all these badges. The pip trends badge count is the most accurate count and can be verified using - 
```sql
SELECT COUNT(*) AS num_downloads
FROM `bigquery-public-data.pypi.file_downloads`
WHERE file.project = 'poetry-core'
  -- Only query the last 30 days of history
  AND DATE(timestamp)
    BETWEEN DATE_SUB(CURRENT_DATE(), INTERVAL 30 DAY)
    AND CURRENT_DATE()
```

You can read more and verify using this link - https://packaging.python.org/en/latest/guides/analyzing-pypi-package-downloads/#counting-package-downloads

Secrus:
Yeah, I don't think this is really needed, pass

PR #743, opened 2024-07-15
Deprecated string defined scripts in favor of reference. Enabled gui scripts.

IgnaciodelaTorreArias:
Resolves: 
Deprecating string scripts
python-poetry-core [TODO](https://github.com/python-poetry/poetry-core/blob/06d72bc0f1f05086a1cc0e760075c519af33db5b/src/poetry/core/masonry/builders/builder.py#L301)
Continuation to my [PR](https://github.com/python-poetry/poetry/pull/9549#issue-2407767175) on python-poetry/poetry

- [x] Added **tests** for changed code. Updated previous Test

As for the deprecation work changes tests toml's to reference scripts, except the ones that shouldn't of course.
Documentation updated on [PR](https://github.com/python-poetry/poetry/pull/9549#issue-2407767175)

**Warning**

On tests the deprecation raises a warning on [test_builder_convert_entry_points](https://github.com/python-poetry/poetry-core/blob/06d72bc0f1f05086a1cc0e760075c519af33db5b/tests/masonry/builders/test_builder.py#L243C1-L243C39) that i don't know how to solve

### DIFF
--- a/src/poetry/core/factory.py
+++ b/src/poetry/core/factory.py
@@ -448,7 +448,9 @@ class Factory:
 
                 for name, script in scripts.items():
                     if isinstance(script, str):
-                        result["warnings"].append('String defined scripts are deprecated in favor of reference definition')
+                        result["warnings"].append(
+                            "String defined scripts are deprecated in favor of reference definition"
+                        )
 
                     if not isinstance(script, dict):
                         continue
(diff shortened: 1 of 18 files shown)


radoering:
I do not think that we have to deprecate anything to support GUI scripts, do we?

radoering:
Ah, I see. In the light of #708, which deprecates most of `tool.poetry.scripts` anyway, I do not think it makes sense to address this TODO now. (If I was completely sure that #708 will be included in the next version, I would even say that we do not have to extend `tool.poetry.scripts` at all.)

PR #744, opened 2024-07-22
Update everything

dimbleby:
mostly motivated by ruff - it is annoying when the ruff that I actually use disagrees with the precommit ruff.  

Updated dependencies and vendored dependencies at the same time: because why not?


src/poetry/core/packages/utils/link.py
```diff
@@ -266,17 +266,6 @@ def is_egg(self) -> bool:
     def is_sdist(self) -> bool:
         return self.ext in {".tar.bz2", ".tar.gz", ".zip"}
 
-    @cached_property
-    def is_artifact(self) -> bool:
```
Secrus, line 270: why was this removed?

dimbleby: SIM103 didn't like the way that it was written: but since the code is unused it is cleaner to remove it altogether than to rearrange it

Secrus (approved):

PR #753, opened 2024-08-30
Switch to in-tree vendoring instead of patching the sys.path

Secrus:
Resolves: python-poetry/poetry#8157


dimbleby:
Did you try running `make vendor/update`?

I expect that will reveal things you might have missed, and also give a hint as to whether the maintenance burden this brings is likely to be painful or not.

abn:
@Secrus I am not sure this is trivial to maintain long term. While the issue mentioned for plugins is not ideal, I get the distinct feeling that there are sufficient workarounds such that plugins could maintain the workarounds more easily than us having to carry the burden of maintaining this.

Just for reference the original discussion around this is at https://github.com/python-poetry/poetry-core/pull/25#discussion_r414052082.

Note that this is before plugins were a thing.

abn:
Also, if I recall correctly, this change also was enabling downstream OS distro package maintainers to remove (or patch) vendor directories to re-package according to their requirements. This change will revert us back to a situation where they will have to start carrying patches.

Secrus:
> Also, if I recall correctly, this change also was enabling downstream OS distro package maintainers to remove (or patch) vendor directories to re-package according to their requirements. This change will revert us back to a situation where they will have to start carrying patches.

Right now it puts us at a disadvantage, cause dependencies from the core override the poetry dependencies, which is a failure waiting to happen. The same happens for plugins (as noted in the issue this addresses). I hate to be selfish here, but I prefer selfish over errorous.

PR #762, opened 2024-09-28
fix: nested directory support for wheel

AsifArmanRahman:
Resolves: python-poetry#9721


radoering (approved):
I improved the test a bit and added an analoguous test for sdists.

Hint: The `# type: ignore` was only necessary because the type hint had been wrong.

PR #770, opened 2024-10-05
remove support for 'x' in version constraints

dimbleby:
removes the most bizarre of the many ways in which poetry tries to handle nonsensical version constraints

fixes https://github.com/python-poetry/poetry/issues/5990

### DIFF
--- a/src/poetry/core/constraints/version/patterns.py
+++ b/src/poetry/core/constraints/version/patterns.py
@@ -17,7 +17,7 @@ TILDE_PEP440_CONSTRAINT = re.compile(
     rf"^~=\s*(?P<version>{VERSION_PATTERN})$", re.VERBOSE | re.IGNORECASE
 )
 X_CONSTRAINT = re.compile(
-    r"^(?P<op>!=|==)?\s*v?(?P<version>(\d+)(?:\.(\d+))?(?:\.(\d+))?)(?:\.[xX*])+$"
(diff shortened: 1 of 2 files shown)


src/poetry/core/constraints/version/patterns.py
```diff
@@ -17,7 +17,7 @@
     rf"^~=\s*(?P<version>{VERSION_PATTERN})$", re.VERBOSE | re.IGNORECASE
 )
 X_CONSTRAINT = re.compile(
```
Secrus, line 19: Isn't this whole pattern only for `x` in version constraints? Didn't dive too deep into the code, but on quick glance, just changing it still leaves some code related to "x constraint"

dimbleby: The code uses the term "x constraint" for some things with `*` in them, but so far as I know this is the only place that is allowing a literal `x`

radoering:
I think I am fine with the change since it is an undocumented feature.

I thought about deprecating it first because it will hurt some people for sure, especially since it is in the build backend. However, a deprecation is not that simple in this case.

@Secrus Feel free to merge it if you agree to remove it.

src/poetry/core/constraints/version/patterns.py
Secrus: Ah, okay, didn't notice the `*` in the old regex. My bad

Secrus (approved):

PR #747, opened 2024-08-01
Add Python 3.13 to `PYTHON_VERSION` and `AVAILABLE_PYTHONS`

jakob-keller:
Resolves: [python-poetry#9598](https://github.com/python-poetry/poetry/issues/9598)

- [x] Added **tests** for changed code.
- [ ] Updated **documentation** for changed code.

PR #776, opened 2024-10-12
release: bump version to 1.9.1

radoering:
### Added

- Add `3.13` to the list of available Python versions ([#747](https://github.com/python-poetry/poetry-core/pull/747)).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "poetry-core"
-version = "1.9.0"
+version = "1.9.1"
 description = "Poetry PEP 517 Build Backend"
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -11,6 +11,9 @@ keywords = ["packaging", "dependency", "poetry"]
 classifiers = [
     "Topic :: Software Development :: Build Tools",
     "Topic :: Software Development :: Libraries :: Python Modules",
+    "Topic :: Software Development :: Libraries :: Python Modules",
```
Secrus, line 14: Duplicated classifier

radoering: Good catch, I wondered why it was marked as changed but I missed the reason.

Secrus (approved):

PR #778, opened 2024-10-13
post 1.9.1 chore

radoering:
In addition to the obligatory changelog merge, some CI changes that were necessary in the 1.9 branch and will be mandatory for the next release from main are included.


.github/workflows/release.yaml
```diff
@@ -0,0 +1,58 @@
+name: Release
+
+on:
+  release:
+    types: [published]
```
Secrus, line 5: So basically, we release to PyPI once the release on GH is created?

radoering: exactly

Secrus (approved):

PR #780, opened 2024-10-16
sdist: make `clean_tar_info` a `classmethod` again because it is used as such for generating PyPI fixtures downstream

radoering:
I found the reason for `clean_tar_info` being a `classmethod` (cf https://github.com/python-poetry/poetry-core/pull/766#discussion_r1789074635):

https://github.com/python-poetry/poetry/blob/dcd5ae1f67abb9ed49a0e7b89235b6f3c4e30661/tests/repositories/fixtures/pypi.org/generate.py#L360

### DIFF
--- a/src/poetry/core/masonry/builders/sdist.py
+++ b/src/poetry/core/masonry/builders/sdist.py
@@ -104,12 +104,13 @@ class SdistBuilder(Builder):
         logger.info(f"Built <comment>{target.name}</comment>")
         return target
 
+    @classmethod
     def add_file_to_tar(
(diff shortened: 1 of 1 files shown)


Secrus:
I'd rather see that being fixed in the fixture generator since it's only importing the builder for that method (which is weird if you ask me). I made a fix here, but it turned out to be meaningless PR.

radoering:
> I'd rather see that being fixed in the fixture generator since it's only importing the builder for that method

So copy & paste?

Secrus:
> So copy & paste?

Either that or pull that method into an util function.

PR #784, opened 2024-10-27
Update license name for EPL-2.0

netomi:
Resolves: python-poetry/poetry#9803

Test case was removed based on suggestion of @Secrus .

Not sure what documentation I should change in this regard, so I skip it for now, but let me know if I need to update something.


dimbleby:
the file you have change is autogenerated, you do not want to be changing that.

Maybe [here](https://github.com/python-poetry/poetry-core/blob/589ecb03f28b30d9a604dc13b51fba55f3981ef5/src/poetry/core/spdx/license.py#L109)

netomi:
oh ty, will update the PR and try to understand how the json file got generated

dimbleby:
I dont think you need to care about the autogenerated file, just revert the change to it

netomi:
yeah, updated my PR and force-pushed, ty guys.

tests/spdx/test_license.py
```diff
@@ -12,6 +12,12 @@ def test_classifier_name() -> None:
     )
 
 
+def test_classifier_name_epl_2_0() -> None:
```
Secrus, line 15: I don't think this test case makes much sense. I get that it checks your exact issue case, but I don't see why we should check this exact license among others that we support. I am ok with no test case for that and treat it as a simple data update.

netomi: I am fine with leaving out a testcase for such a trivial test. I just wanted to check myself that my change resulted in the correct license name that is later used for generating the classifier.

netomi:
Added a test case to ensure that all known licenses return a valid trove classifier.
While adding the test I figured that the Aladdin license was also failing, so fixed it right away.

Secrus (approved):

PR #791, opened 2024-11-21
vcs: fix parsing of basic auth http(s) credentials

abn:
Poetry today does not correctly handle `poetry add 'git+https://username:password@github.com/user/repo.git'`. While we do not recommend sensitive values to be persisted in `pyptoject.toml`, `poetry.lock` or in any other locations, this is the case today when the values are manually put in the `pyproject.toml`.

Eventually, we need to figure out a way to parse the credentials and store it into the password manager and not persist it in any files that are distributed. That is largely a change in Poetry and not in core. However, there are aspects of this that should be handled here. I will follow up with another PR to handle env var credentials as specified in PEP610.

I have also removed some unused code.

Closes: #536

### DIFF
--- a/src/poetry/core/vcs/git.py
+++ b/src/poetry/core/vcs/git.py
@@ -18,6 +18,16 @@ PATH = r"[%\w~.\-\+/\\\$]+"
 NAME = r"[%\w~.\-]+"
 REV = r"[^@#]+?"
 SUBDIR = r"[\w\-/\\]+"
+PATTERN_SUFFIX = (
+    r"(?:"
+    rf"#(?:egg=.+?&subdirectory=|subdirectory=)(?P<subdirectory>{SUBDIR})"
+    r"|"
+    r"#egg=?.+"
+    r"|"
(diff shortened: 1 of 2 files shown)


src/poetry/core/vcs/git.py
```diff
@@ -11,31 +11,46 @@
 
 
 PROTOCOL = r"\w+"
-USER = r"[a-zA-Z0-9_.-]+"
+URL_RESTRICTED = r"[^/\?#:@]"
```
radoering, line 14: LGTM. I just wonder if there is a reference for this / if this is complete or if it is just a best guess (which might also be ok)?

Secrus:
How does this work with #536? Is it a replacement for that PR?

src/poetry/core/vcs/git.py
abn: We could use https://url.spec.whatwg.org/#forbidden-host-code-point.  This one is also tricky as the value may not be b64 encoded at the point of parsing.

radoering: Maybe. (I like to have a comment with a link. 😀) But we can also keep it as is if you think it does not make sense. I leave the decision to you.

radoering (approved):

PR #800, opened 2024-12-23
Improve error message for packages in src folder

martinvotruba-r7:
# Problem description

Currently when a `poetry install` command is run on a project, which has source stored under `src/project_name`, but a folder named `project_name` exists in the root, the command will fail with `project_name is not a package.` It can be rather hard to determine the root cause, as one might easily overlook the empty folder.

## Steps to reproduce
1. create a new poetry project `poetry new poetry-demo`
2. Move `poetry_demo/__init__.py` to `src/poetry_demo/__init__.py`
3. run `poetry install`

The command will fail with the following
```
poetry install
Installing dependencies from lock file

Installing the current project: poetry-demo (0.1.0)
poetry_demo is not a package.
```

This is caused by the presence of poetry_demo folder in the root.

```
tree -I ".git" --noreport --charset utf-8 .
.
├── README.md
├── poetry.lock
├── poetry_demo
├── pyproject.toml
├── src
│   └── poetry_demo
│       └── __init__.py
└── tests
    └── __init__.py
```

When I remove the  poetry_demo folder, the `poetry install` command succeeds. 

# Resolution

To provide more information about the actual root cause of the issue, I have added another validation method to the Factory class. This will output a warning when there is both
- an empty `project_name` folder in the root
- non-empty `src/project_name` folder
The provided warning message is the following

```
Found empty directory 'project_name' in project root while the actual package is in 'src/project_name'. This may cause issues with package installation. Consider removing the empty directory.
```

If there is a better place / method how to alert user to the empty `project_name` folder, I am happy to hear it and adjust my implementation.

### DIFF
--- a/src/poetry/core/factory.py
+++ b/src/poetry/core/factory.py
@@ -591,6 +591,9 @@ class Factory:
     def validate_project_in_src(
         cls, toml_data: dict[str, Any], base_path: Path | None
     ) -> dict[str, list[str]]:
+        """
+        Checks the validity of projects located under src/ subdirectory.
+        """
         result: dict[str, list[str]] = {"errors": [], "warnings": []}
 
         project_name = toml_data.get("project", {}).get("name") or toml_data.get(
@@ -602,7 +605,6 @@ class Factory:
 
         project_name = module_name(project_name)
 
-        # Check for empty directory with same name as project
         project_dir = base_path / project_name
         src_project_dir = base_path / "src" / project_name
 
--- a/src/poetry/core/factory.py
+++ b/src/poetry/core/factory.py
@@ -620,7 +620,7 @@ class Factory:
         if project_dir_empty and src_dir_not_empty:
             result["warnings"].append(
                 f"Found empty directory '{project_name}' in project root while the actual package is in "
-                f"src/{project_name}. This may cause issues with package installation. "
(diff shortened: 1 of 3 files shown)


radoering:
> If there is a better place / method how to alert user to the empty `project_name` folder, I am happy to hear it and adjust my implementation.

First thought: This should be part of `validate()` but not a separate method. However, I see that this is not possible without changing the interface of `validate()` because `validate()` only knows about the data in the pyproject.toml and nothing about the project structure. So maybe not such a good idea after all.

Second thought: Somewhere we do check if referenced readme files do exist. That is something similar. Maybe, we should put this check at the same place. We do the readme check in the `check` command: https://github.com/python-poetry/poetry/blob/a7079acc7631eda4f7ba0f9a34d8d1468d50cba3/src/poetry/console/commands/check.py#L84-L93

Doing it in the check command means you will only get the warning when running `poetry check`. The advantage is that the code is not executed every time. The disadvantage is that you have to run `poetry check` to get the warning.

Considering the similarity to the readme check, I think I prefer to move it into the `check` command.

martinvotruba-r7:
@radoering I have moved the implementation into the suggested `check` comand.

https://github.com/python-poetry/poetry/pull/10236

PR #803, opened 2025-01-06
Fix urls entries in pyproject.toml with variable cases

gantoine:
Came across an issue trying to `poetry install` today with `streaming-form-data` in my `pyproject.toml`. The [library's pyproject.toml](https://github.com/siddhantgoel/streaming-form-data/blob/main/pyproject.toml) uses lowercase keys for entries under `[tool.poetry.urls]`, which breaks when `Metadata.from_package` tries to build `project_urls`. Added a couple tests for upper, lower and mixed-case entries.

### DIFF
--- a/src/poetry/core/masonry/metadata.py
+++ b/src/poetry/core/masonry/metadata.py
@@ -107,11 +107,12 @@ class Metadata:
 
         if package.urls:
             for name, url in package.urls.items():
-                if name.lower() == "homepage" and meta.home_page == url:
+                name_lower = name.lower()
+                if name_lower == "homepage" and meta.home_page == url:
                     continue
-                if name == "repository" and url == package.urls["Repository"]:
+                if name_lower == "repository" and url == package.urls.get(name):
                     continue
-                if name == "documentation" and url == package.urls["Documentation"]:
+                if name_lower == "documentation" and url == package.urls.get(name):
                     continue
 
                 meta.project_urls += (f"{name}, {url}",)
--- a/src/poetry/core/packages/package.py
+++ b/src/poetry/core/packages/package.py
@@ -117,6 +117,9 @@ class Package(PackageSpecification):
 
         self._yanked = yanked
 
+        # Currently only used for testing purposes
+        self._urls: dict[str, str] = {}
+
     @property
     def name(self) -> NormalizedName:
         return self._name
@@ -354,6 +357,9 @@ class Package(PackageSpecification):
     @property
     def urls(self) -> dict[str, str]:
         urls = {}
+
+        if self._urls:
+            return self._urls
 
         if self.homepage:
             urls["Homepage"] = self.homepage
--- a/tests/masonry/test_metadata.py
+++ b/tests/masonry/test_metadata.py
@@ -60,3 +60,53 @@ def test_from_package_readme_issues(
         Metadata.from_package(package)
 
     assert str(e.value) == message
+
+def test_from_package_urls_case_sensitive() -> None:
+    package = ProjectPackage("foo", "1.0")
+    package.homepage = "https://example.com"
+    package._urls = {
+        "Homepage": "https://example.com",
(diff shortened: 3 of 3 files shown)


src/poetry/core/masonry/metadata.py
```diff
@@ -107,11 +107,12 @@ def from_package(cls, package: ProjectPackage) -> Metadata:
 
         if package.urls:
             for name, url in package.urls.items():
-                if name.lower() == "homepage" and meta.home_page == url:
+                name_lower = name.lower()
+                if name_lower == "homepage" and meta.home_page == url:
                     continue
-                if name == "repository" and url == package.urls["Repository"]:
+                if name_lower == "repository" and url == package.urls.get(name):
```
dimbleby, line 113: is this not always definitely true?  because `name` and `url` are a pair from `package.urls.items()`.

I am not sure what the intent of this code was but this variation seems pointless?

src/poetry/core/packages/package.py
```diff
@@ -117,6 +117,9 @@ def __init__(
 
         self._yanked = yanked
 
+        # Currently only used for testing purposes
```
dimbleby, line 120: this is not true, per the change at line 360.

src/poetry/core/masonry/metadata.py
dimbleby: perhaps the original code was just a roundabout way of reading `package.repository_url` - and doing that directly is a better fix?

src/poetry/core/packages/package.py
gantoine: Sorry I should have clarified, this is happening in CI for a real project of mine, not just in testing. The addition of `self._urls` is for the tests i wrote in `test_metadata.py`, since `url` is a `@property` and can't be set directly.

https://github.com/rommapp/romm/actions/runs/12623342037/job/35174039307#step:10:3085

src/poetry/core/masonry/metadata.py
gantoine: yeah i think the original code wasn't well written, i'll simplify it 👍🏼

src/poetry/core/packages/package.py
radoering: I would like to avoid adding production code that is only required for testing. Can you use a `PropertyMock` instead? See https://github.com/python-poetry/poetry/blob/5d8f8800b2cbb53da8057965663844bd1e04e455/tests/console/commands/test_check.py#L82-L86 for example.

radoering (changes requested):

src/poetry/core/packages/package.py
dimbleby: I think this is entirely redundant in the current version, the tests can manipulate the homepage and repository-url fields directly

src/poetry/core/masonry/metadata.py
```diff
@@ -107,11 +107,10 @@ def from_package(cls, package: ProjectPackage) -> Metadata:
...
         if package.urls:
             for name, url in package.urls.items():
-                if name.lower() == "homepage" and meta.home_page == url:
+                name_lower = name.lower()
+                if name_lower == "homepage" and meta.home_page == url:
                     continue
-                if name == "repository" and url == package.urls["Repository"]:
-                    continue
-                if name == "documentation" and url == package.urls["Documentation"]:
+                if name_lower == "repository" or name_lower == "documentation":
```
dimbleby, line 113: I guess the tests have started failing because you have changed the logic

probably want to check "if name is repository and url is repository-url", etc

gantoine:
Right updated the PR based on feedback, however it seems the downstream poetry tests are still failing with these changes.

dimbleby:
poetry-core tests are also currently broken

In passing I spotted PEP753 which says that poetry should change its handling of these anyway, so I have submitted #807

radoering:
Superseded by #807

PR #804, opened 2025-01-06
Fix CI, update pyproject.toml and version

Secrus:
## Summary by Sourcery

CI:
- Use poetry check --lock instead of poetry lock --check.


pyproject.toml
```diff
@@ -1,29 +1,42 @@
-[tool.poetry]
+[project]
 name = "poetry-core"
-version = "2.0.0"
+version = "2.0.0.dev0"
```
abn, line 3: This should be either a post release or minor/patch dev

tests/test_core_version.py
```diff
@@ -8,4 +8,4 @@
 
 def test_version_is_synced() -> None:
     pyproject = PyProjectTOML(Path(__file__).parent.parent.joinpath("pyproject.toml"))
-    assert __version__ == pyproject.poetry_config.get("version")
+    assert __version__ == pyproject.data["project"]["version"]
```
abn, line 11: Out of scope for this change but we should expose `pyproject.version` etc.

abn (approved):

PR #805, opened 2025-01-06
fix: Fix handling of "format" in package includes initialization

finswimmer:
Ensure that "format" is always processed as a list when initializing package includes. If not explicit set default to sdist and wheel.

Resolves: https://github.com/python-poetry/poetry/issues/9961

## Summary by Sourcery

Bug Fixes:
- Fix handling of "format" in package includes initialization.

## Summary by Sourcery

Bug Fixes:
- Resolve an issue where package includes were not correctly handled if the "format" field was not a list.

### DIFF
--- a/src/poetry/core/factory.py
+++ b/src/poetry/core/factory.py
@@ -376,7 +376,15 @@ class Factory:
             package.exclude = exclude
 
         if packages := tool_poetry.get("packages"):
-            package.packages = packages
+            packages_ = []
+            for p in packages:
+                formats = p.get("format", ["sdist", "wheel"])
+                if not isinstance(formats, list):
+                    formats = [formats]
+
+                packages_.append({**p, "format": formats})
+
+            package.packages = packages_
 
     @classmethod
     def create_dependency(
--- a/src/poetry/core/factory.py
+++ b/src/poetry/core/factory.py
@@ -7,6 +7,7 @@ from collections.abc import Mapping
 from pathlib import Path
 from typing import TYPE_CHECKING
 from typing import Any
+from typing import Literal
 from typing import Union
 
 from packaging.utils import canonicalize_name
@@ -358,6 +359,22 @@ class Factory:
             package.extras = package_extras
 
     @classmethod
+    def _prepare_formats(
+        cls,
+        items: list[dict[str, Any]],
+        default_formats: list[Literal["sdist", "wheel"]],
+    ) -> list[dict[str, Any]]:
+        result = []
+        for item in items:
+            formats = item.get("format", default_formats)
+            if not isinstance(formats, list):
+                formats = [formats]
+
+            result.append({**item, "format": formats})
+
+        return result
+
+    @classmethod
     def _configure_package_poetry_specifics(
         cls, package: ProjectPackage, tool_poetry: dict[str, Any]
     ) -> None:
@@ -367,24 +384,20 @@ class Factory:
             package.build_config = build or {}
 
         if includes := tool_poetry.get("include"):
-            package.include = [
+            includes = [
                 include if isinstance(include, dict) else {"path": include}
                 for include in includes
             ]
+
+            package.include = cls._prepare_formats(includes, default_formats=["sdist"])
 
         if exclude := tool_poetry.get("exclude"):
             package.exclude = exclude
 
         if packages := tool_poetry.get("packages"):
-            packages_ = []
-            for p in packages:
-                formats = p.get("format", ["sdist", "wheel"])
-                if not isinstance(formats, list):
-                    formats = [formats]
-
-                packages_.append({**p, "format": formats})
-
-            package.packages = packages_
+            package.packages = cls._prepare_formats(
+                packages, default_formats=["sdist", "wheel"]
+            )
 
     @classmethod
     def create_dependency(
--- a/src/poetry/core/masonry/builders/builder.py
+++ b/src/poetry/core/masonry/builders/builder.py
@@ -46,23 +46,16 @@ class Builder:
     def _module(self) -> Module:
         from poetry.core.masonry.utils.module import Module
 
-        packages: list[dict[str, str | dict[str, str]]] = []
-        includes: list[dict[str, str | dict[str, str]]] = []
-        for source_list, target_list, default in [
-            (self._package.packages, packages, ["sdist", "wheel"]),
-            (self._package.include, includes, ["sdist"]),
-        ]:
-            for item in source_list:
-                # Default to including in both sdist & wheel
-                # if the `format` key is not provided.
-                formats = item.get("format", default)
-                if not isinstance(formats, list):
-                    formats = [formats]
-
-                if self.format and self.format not in formats:
-                    continue
-
-                target_list.append({**item, "format": formats})
+        packages = [
+            item
+            for item in self._package.packages
+            if not self.format or self.format in item["format"]
+        ]
+        includes = [
+            item
+            for item in self._package.include
+            if not self.format or self.format in item["format"]
+        ]
 
         return Module(
             self._package.name,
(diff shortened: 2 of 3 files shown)


Secrus:
A test would be in order

src/poetry/core/masonry/utils/module.py
```diff
@@ -72,11 +72,15 @@ def __init__(
             packages = [default_package]
 
         for package in packages:
+            formats = package.get("format", ["sdist", "wheel"])
```
dimbleby, line 75: wasn't the point of #773 to default this only to "sdist"?

anuragnatoo, line 75: Hi @finswimmer 

According to the documentation [here](https://python-poetry.org/docs/pyproject/) (attaching a screenshot) we should default it to only sdist, so would it be better if we set it as `formats = package.get("format", ["sdist"] )` or `formats = package.get("format", "sdist" )`
image

finswimmer: In my understanding this was about the behavior of format for the include keyword and not for packages. 🤷I guess @radoering can say more about it.

radoering (changes requested):
Maybe, the fix should be downstream. Give me some time to think about it. Currently, all the default handling is in https://github.com/python-poetry/poetry-core/blob/61081dd96eb5d5096ef97cb12d15c461a41121cf/src/poetry/core/masonry/builders/builder.py#L49-L65 That is why it does not work for `poetry run`. We have different defaults for `package` includes and `include` includes. See python-poetry/poetry#9691 for details.

src/poetry/core/masonry/utils/module.py
dimbleby: can confirm that I find this confusing

so I see, #773 patched up the package and include objects to add default values if not present - but there is nothing corresponding to this in poetry's `RunCommand._module()`.

Perhaps a better way would be to kill both birds with one stone and do this defaulting at the time the package definition is read, I think in poetry-core's `Factory._configure_package_poetry_specifics`.  Then it is always available for everyone and no further special casing is needed.

finswimmer: Ah, I guess you found the places I was looking for initially. I'm on mobile so I have to take a look at latet/tomorrow.

radoering:
I am wondering if the downstream tests show some unwanted behavior. I remember these tests were the reason I did not do this in the factory but I do not remember if it is a real issue or just the tests... It might be that we are producing unwanted changes in the `pyproject.toml` now.

finswimmer:
> I am wondering if the downstream tests show some unwanted behavior. I remember these tests were the reason I did not do this in the factory but I do not remember if it is a real issue or just the tests... It might be that we are producing unwanted changes in the `pyproject.toml` now.

My previous implementation was problematic because it changes the data read from the toml directly. I changed it, so that the toml data stays untouched and only the data that goes into the `package.packages` attribute is changed. 

The only downstream test that is failing, is the same, that we need to adopt in `poetry-core`. I'm not sure how useful it is, to have this test in `poetry` itself.

radoering:
Looks good now.

> I'm not sure how useful it is, to have this test in `poetry` itself.

Also not sure. It might make sense because downstream we inherit from `Factory` and may want to check that we did not break something.

Can you prepare a downstream PR? (I think we can merge this with the failing test but we should update downstream to use poetry-core's master branch shortly afterwards so that we keep the time of failing downstream tests short.)

radoering:
One more thing: Can we move the default handling completely into `Factory`? Add default handling for` package.include` and remove https://github.com/python-poetry/poetry-core/blob/61081dd96eb5d5096ef97cb12d15c461a41121cf/src/poetry/core/masonry/builders/builder.py#L49-L65

finswimmer:
> Can you prepare a downstream PR? 

I prepared one here: https://github.com/python-poetry/poetry/pull/9992

> One more thing: Can we move the default handling completely into Factory?

Done. (At least I think so :grinning: )

tests/test_factory.py
```diff
@@ -354,20 +354,29 @@ def test_create_poetry_with_packages_and_includes() -> None:
     package = poetry.package
 
     assert package.packages == [
-        {"include": "extra_dir/**/*.py"},
-        {"include": "extra_dir/**/*.py"},
-        {"include": "my_module.py"},
-        {"include": "package_with_include"},
-        {"include": "tests", "format": "sdist"},
+        {"include": "extra_dir/**/*.py", "format": ["sdist", "wheel"]},
```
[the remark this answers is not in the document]
abn: Good suggestion, but might be more relevant for dedicated new PR.

src/poetry/core/factory.py
```diff
@@ -367,16 +367,32 @@ def _configure_package_poetry_specifics(
...
+            includes_ = []
+            for include in includes:
+                if not isinstance(include, dict):
+                    include = {"path": include}
+
+                formats = include.get("format", ["sdist"])
+                if not isinstance(formats, list):
+                    formats = [formats]
+
+                includes_.append({**include, "format": formats})
```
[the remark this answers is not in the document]
abn: @finswimmer we could reduce duplication here as suggested.

src/poetry/core/masonry/builders/builder.py
```diff
@@ -48,21 +48,15 @@ def _module(self) -> Module:
...
         includes: list[dict[str, str | dict[str, str]]] = []
-        for source_list, target_list, default in [
-            (self._package.packages, packages, ["sdist", "wheel"]),
-            (self._package.include, includes, ["sdist"]),
+        for source_list, target_list in [
+            (self._package.packages, packages),
+            (self._package.include, includes),
         ]:
             for item in source_list:
-                # Default to including in both sdist & wheel
```
abn, line 56: We should drop this as well. Seems to be a left over from [1bdbd900c4fc1e2d4ed1ef15dfbcafdcb66dd892](https://github.com/python-poetry/poetry-core/commit/1bdbd900c4fc1e2d4ed1ef15dfbcafdcb66dd892). 

This is no longer accurate. We are defaulting to `sdist` for `includes` and both for `packages`.

abn (changes requested):

src/poetry/core/factory.py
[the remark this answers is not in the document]
radoering: I was thinking that we could do it the same way as it had been done in `builder.py` to reduce duplication. A helper function should be similar and maybe even cleaner.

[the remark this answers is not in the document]
finswimmer: Added a helper function.

abn:
@sourcery-ai review

src/poetry/core/masonry/builders/builder.py
```diff
@@ -46,23 +46,16 @@ def __init__(self, poetry: Poetry, executable: Path | None = None) -> None:
...
+        packages: list[dict[str, str | dict[str, str]]] = [
+            {**item}
+            for item in self._package.packages
+            if not self.format or self.format in item["format"]
+        ]
+        includes: list[dict[str, str | dict[str, str]]] = [
+            {**item}
+            for item in self._package.include
+            if not self.format or self.format in item["format"]
+        ]
```
abn, line 58:
```suggestion
        packages: list[dict[str, str | dict[str, str]]] = [
            item
            for item in self._package.packages
            if not self.format or self.format in item["format"]
        ]
        includes: list[dict[str, str | dict[str, str]]] = [
            item
            for item in self._package.include
            if not self.format or self.format in item["format"]
        ]
```
We should be able to use the item as is safely i reckon.

finswimmer: I wasn't sure about it, that why I haven't touch. So let's :crossed_fingers:

abn (approved):

PR #810, opened 2025-01-08
fix: ensure all poetry-dependencies are added to lock file (#9959)

finswimmer:
Resolves: python-poetry/poetry#9959

## Summary by Sourcery

Bug Fixes:
- Resolve an issue where not all poetry dependencies were added to the lock file.

### DIFF
--- a/src/poetry/core/packages/dependency_group.py
+++ b/src/poetry/core/packages/dependency_group.py
@@ -39,6 +39,7 @@ class DependencyGroup:
         for dep in self._poetry_dependencies:
             poetry_dependencies_by_name[dep.name].append(dep)
 
+        enriched_poetry_dependencies = []
         dependencies = []
         for dep in self._dependencies:
             if dep.name in poetry_dependencies_by_name:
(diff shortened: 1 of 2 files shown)


radoering (changes requested):
Originally, I thought you should not be able to add additional dependencies via `tool.poetry` if you are using `project.dependencies`; you should only be able to enrich `project.dependencies`. However, silently ignoring them is not a good option either. I think my original thought should still hold true for building but maybe we can relax it for locking.

However, I think the underlying issue in https://github.com/python-poetry/poetry/issues/9959 is deeper and not fixed by the PR: If you mix `project.optional-dependencies` and dynamic `project.dependencies` via `tool.poetry.dependencies`, the dependencies will also be missing when building a wheel. The issue is that optional dependencies go into the same dependency group as normal dependencies. The logic inside dependency groups works only well if there is one section in `project` and one section in `tool.poetry`, not if there are two sections each and you you are only using one in `project` and want to have the other dynamic.

I am not sure if fixing the underlying issue will make this PR redundant. Probably not. I will try to find a solution for the underlying issue tomorrow if nobody beats me to it and would like to keep this one open until it is clear if we need it or not.

PR #813, opened 2025-01-13
fix(pyproject): relax validation for non-packages

abn:
Relates-to: python-poetry/poetry#10031 python-poetry/poetry#10032

Proposing this fix to see what the team makes of this. This ensures that non-package projects do not need to adhere to strict project section requirements.

## Summary by Sourcery

Relax project metadata validation for non-package projects.

Bug Fixes:
- Fixed an issue where non-package projects were subject to the same strict project section requirements as package projects.

Tests:
- Added a test case to verify that project creation succeeds for non-package projects even without all required project metadata.

### DIFF
--- a/tests/test_factory.py
+++ b/tests/test_factory.py
@@ -798,6 +798,19 @@ The Poetry configuration is invalid:
     assert str(e.value) == expected
 
 
+def test_create_poetry_succeeds_for_non_package_without_required_project_metadata(
+    common_fixtures_directory: Path,
+) -> None:
+    assert (
+        Factory().create_poetry(
+            common_fixtures_directory
+            / "non_package_mode_with_invalid_project"
+            / "pyproject.toml"
+        )
+        is not None
+    )
+
+
 def test_create_poetry_fails_on_invalid_mode() -> None:
     with pytest.raises(RuntimeError) as e:
         Factory().create_poetry(
--- a/src/poetry/core/factory.py
+++ b/src/poetry/core/factory.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import logging
+import re
 
 from collections import defaultdict
 from collections.abc import Mapping
@@ -569,12 +570,19 @@ class Factory:
 
         result: dict[str, list[str]] = {"errors": [], "warnings": []}
 
+        # we can ignore project.<name|version> as they need to be validated only
+        # when package-mode is True, this is done so later on in the code
+        ignore_required_properties = re.compile(
+            r"data must contain \[((, )?'(name|version)')+] properties"
+        )
+
         # Validate against schemas
         project = toml_data.get("project")
         if project is not None:
             project_validation_errors = [
                 e.replace("data", "project")
                 for e in validate_object(project, "project-schema")
+                if strict or not re.match(ignore_required_properties, e)
             ]
             result["errors"] += project_validation_errors
         # With PEP 621 [tool.poetry] is not mandatory anymore. We still create and
--- a/tests/fixtures/non_package_mode_with_invalid_project/pyproject.toml
+++ b/tests/fixtures/non_package_mode_with_invalid_project/pyproject.toml
@@ -1,0 +1,8 @@
+[project]
+requires-python = ">=3.10,<3.13"
+
+[tool.poetry]
+package-mode = false
+
+[tool.poetry.dependencies]
+cleo = "^0.6"


tests/test_factory.py
```diff
@@ -798,6 +798,20 @@ def test_create_poetry_fails_on_invalid_configuration() -> None:
     assert str(e.value) == expected
 
 
+def test_create_poetry_succeeds_for_non_package_without_required_project_metadata() -> (
+    None
+):
+    assert (
+        Factory().create_poetry(
+            Path(__file__).parent
+            / "fixtures"
```
Secrus, line 807:
```suggestion
def test_create_poetry_succeeds_for_non_package_without_required_project_metadata(common_fixtures_directory: Path) -> (
    None
):
    assert (
        Factory().create_poetry(
            common_fixtures_directory
```

Secrus (commented):
I am all for that. Consider it approved from my side.

radoering:
Thus, we will allow to violate the standard in non-package mode?

abn:
Yeah, thinking is in a non-package mode those are not required and seems to be a foot gun mostly. But that is just my thinking atm, I am not strongly for or against this approach. Seems better as user's seem to not expect non-packages to need these properties - I tend to agree, at least for version.

dimbleby:
> violate the standard in non-package mode?

that is of course exactly the objection to this approach.  (All these years of folk beating up on poetry for not following PEP621 - and now here we go...!)

IMO the compliant and consistent thing is to tell people that if they don't want a package then they shouldn't be using the `project` table: because the whole point of that table is packaging, and it is defined by the python packaging authority, and poetry has no business changing the rules for how it is used.  

(In that case, it may be that that there are some `tool.poetry.xxx` bits that should be considered not-deprecated in non-packaging mode: because the corresponding `project.xxx` is not a good replacement in non-packaging mode.)

While I think this all is _mostly_ not very important, I bet that it will sooner or later cause bug reports and incompatibilities.  Eg there will be folk using some tool `foo` which also uses `pyproject.toml` - and when `foo` does not work properly with non-compliant files, they're going to complain here.

finswimmer:
I agree with @dimbleby as already mentioned on discord. :)

dimbleby:
I guess if you wanted to talk yourself into believing that what is proposed here is compliant, you could argue that 
- PEP621 does not apply _at all_ in non-packaging mode
- the `project` table that poetry uses in non-packaging mode is not the same as the `project` table in the PyPA specifications
  - but as a convenience for users, it just happens to look almost exactly like it.

I don't love this line of reasoning!  But it has its merits.

radoering:
We could also take a middle path: We do not create non-compliant `project` sections (`poetry init`), but we are able to work with them. However, the question would still be what to do in `poetry check`.

abn:
> (In that case, it may be that that there are some tool.poetry.xxx bits that should be considered not-deprecated in non-packaging mode: because the corresponding project.xxx is not a good replacement in non-packaging mode.)

This is what motivated me proposing this. If we say that properties available in `project` are not deprecated in `tool.poetry` when using `package-mode = false`, we can have a more rigid stance.

> We could also take a middle path: We do not create non-compliant project sections (poetry init), but we are able to work with them. However, the question would still be what to do in poetry check.

Does this mean poetry wont use `project` section, when let's say `poetry add <dep>` is used when `package-mode = false` ? In that case, with check, we can enforce that a `project` section cannot exist with `package-mode = false`.

abn:
To add to this, we could also improve cli usability to add features that let users do `poetry <new|init> --non-package` and decide on what is generated. But there are several places we will have to support this branching.

radoering:
> Does this mean poetry wont use `project` section, when let's say `poetry add <dep>` is used when `package-mode = false` ? In that case, with check, we can enforce that a `project` section cannot exist with `package-mode = false`.

My thinking was that Poetry will use the `project` section if it exists. It will not care if the existing `project` section is standard-compliant or not, it will just keep the state of standard-compliance. However, if the `project` section does not exist it will add the dependency to `tool.poetry.dependencies`, not creating a non-standard-compliant `project` section.

Your proposal solves the `check` issue, but I think there will be users who want to use the `project` section in non-package mode. And I would strongly argue that we should at least allow to use a standard-compliant `project` section in non-package mode.

abn:
I am okay with that approach. In this case I think we can reject this and make the needed changes in the frontend.

johnthagen:
> I think there will be users who want to use the project section in non-package mode.

Yes, many of us see PEP 621 (for all of its ... unfortunate design choices for `dependencies` formatting) as _the_ way now, so there is a big onus to adopting it in all cases. It's valuable to align with the rest of the community whenever possible.

> PEP621 does not apply at all in non-packaging mode

This was my original thinking for https://github.com/python-poetry/poetry/issues/10031 as well. I see the requirements imposed on `[project]` to make sense for a _package_ that might be installed from sdist and into a wheel.

But I can also see the counter argument against this.

As example what other tools do in this space, `uv` requires `name` and `version`, but Rye does not

- https://github.com/astral-sh/uv/issues/10204#issuecomment-2563976787 
- https://github.com/astral-sh/uv/issues/10204#issuecomment-2563996532

PR #823, opened 2025-01-21
wip: poc pep 735

finswimmer:
Resolves: python-poetry#


finswimmer:
@sourcery-ai review

finswimmer:
@sourcery-ai review

finswimmer:
@sourcery-ai review

tests/test_factory.py
```diff
@@ -869,6 +869,19 @@ def test_create_poetry_with_invalid_dev_dependencies(caplog: LogCaptureFixture)
     assert any("dev" in r.groups for r in poetry.package.all_requires)
 
 
+def test_create_poetry_with_invalid_dependency_groups() -> None:
+    with pytest.raises(RuntimeError) as e:
+        _ = Factory().create_poetry(
+            fixtures_dir / "project_with_invalid_dependency_groups", with_groups=False
```
radoering, line 875: Since you explicitly passed `with_groups=False`, I am wondering if this should raise an error if `with_groups` is `False` or only if `with_groups` is `True`?

`with_groups=False` is used for building, i.e. when dependency groups are irrelevant. On the other side, violating the scheme is so essential it may be fine to prevent building.

Anyway, I would make the test parametrized with `with_groups` `False` and `True`.

finswimmer: > On the other side, violating the scheme is so essential it may be fine to prevent building.

I agree. I parametrized the test.

radoering (changes requested):
Support for https://peps.python.org/pep-0735/#dependency-object-specifiers is missing, which might be confusing for users.

fosskers:
I'm sure there are many who are eagerly awaiting this, thank you.

src/poetry/core/factory.py
```diff
@@ -377,6 +407,65 @@ def _configure_package_dependencies(
 
             package.extras = package_extras
 
+    @classmethod
+    def _normalize_depenency_group_names(
```
citelao, line 411: Nit: typo (missing the second `d`; should be `_normalize_dependency_group_names`)

citelao:
Has development migrated to #837? Or is this PR blocked on that one? I'm happy to do some work to help this come in.

radoering:
@sourcery-ai review

src/poetry/core/factory.py
```diff
@@ -333,25 +360,62 @@
...
+                    group=group,
+                    dependencies=dependencies,
+                )
+                pep739_include_groups[group_name] = included_groups
+            # create groups from the tool.poetry.group section
+            # with no corresponding entry in dependency-groups
+            # and add dependency information for existing groups
+            poetry_include_groups = {}
+            for group_name, group_config in tool_poetry_groups.items():
+                poetry_include_groups[group_name] = group_config.get(
```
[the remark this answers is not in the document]
radoering: Normalization is handled implicitly when fetching groups via `package.dependency_group(name)`. We want to output the original names in messages if possible.

src/poetry/core/factory.py
```diff
@@ -87,7 +88,27 @@ def get_package(cls, name: str, version: str) -> ProjectPackage:
...
+        for constraint in dependencies:
+            if isinstance(constraint, str):
+                dep = Dependency.create_from_pep_508(
+                    constraint,
+                    relative_to=package.root_dir,
+                    groups=[group.pretty_name],
+                )
+                group.add_dependency(dep)
+            else:
+                group_includes.append(constraint["include-group"])
```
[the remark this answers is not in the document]
radoering: This cannot happen because the schema does not allow it, but to be safe in case the schema is extended some day it is probably a good idea anyway.

radoering:
I refactored the logic for group includes a bit so that we use the same logic for PEP 735 groups and `tool.poetry` groups.

The PR should be ready for merging now but I think I will wait until I reviewed python-poetry/poetry#10130, in case I spot any issues while reviewing this one.

PR #830, opened 2025-02-01
fix(metadata): ignore optional deps not part of an extra

abn:
Resolves: python-poetry/poetry#2357

## Summary by Sourcery

Ignore optional dependencies that are not part of an extra when building a wheel.

Bug Fixes:
- Fixed a bug where optional dependencies that are not part of an extra were included in the wheel metadata.

Tests:
- Added a test to verify that optional dependencies without extras are excluded from the metadata.

### DIFF
--- a/tests/masonry/builders/fixtures/with_optional_without_extras/pyproject.toml
+++ b/tests/masonry/builders/fixtures/with_optional_without_extras/pyproject.toml
@@ -1,0 +1,19 @@
+[project]
+name = "my-packager"
+description = "Something"
+version = "0.1"
+classifiers = [
+    "Topic :: Software Development :: Build Tools",
+    "Topic :: Software Development :: Libraries :: Python Modules"
+]
+dynamic = ["dependencies"]
+
+[tool.poetry.dependencies]
+requests = { version = "^0.28.1", optional = true }
+httpx = { version = "^0.28.1", optional = true }
+grpcio = { version = "^0.2.0" }
+pycowsay = { version = "^0.1.0" }
+
+[tool.poetry.extras]
+http = ["httpx"]
+grpc = ["grpcio"]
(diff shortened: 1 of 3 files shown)


radoering (approved):

tests/masonry/builders/fixtures/with_optional_without_extras/pyproject.toml
```diff
@@ -0,0 +1,19 @@
...
+classifiers = [
+    "Topic :: Software Development :: Build Tools",
+    "Topic :: Software Development :: Libraries :: Python Modules"
+]
+dynamic = ["dependencies"]
+
+[tool.poetry.dependencies]
+requests = { version = "^0.28.1", optional = true }
+httpx = { version = "^0.28.1", optional = true }
+grpcio = { version = "^0.2.0" }
```
dimbleby, line 14: This test case seems to show that it is ok not to bother marking grpc - or any other extra - as optional.  Is that the intention?

I would expect that a mismatch (in either direction) between declared extras and "optional" dependencies should result in an error, not in quietly taking a guess as to what was meant?

radoering: Yes, I have noticed that, too. And I thought that at least `poetry check` should give a warning. On the other side, we deprecated `poetry.extras` anyway.

dimbleby: Yes, this is place where the PEP621 syntax improves on poetry's - by making such mistakes impossible.

abn: I just added that case for coverage of status quo to be honest and to verify that the change did not accidentally regress the case.

But yes, PEP 621makes it impossible for this.

PR #833, opened 2025-02-04
add `has_upper_bound` method to `VersionConstraint`

radoering:
Required for python-poetry/poetry#10146

## Summary by Sourcery

Tests:
- Add tests for the `has_upper_bound` method.

### DIFF
--- a/src/poetry/core/constraints/version/empty_constraint.py
+++ b/src/poetry/core/constraints/version/empty_constraint.py
@@ -22,6 +22,14 @@ class EmptyConstraint(VersionConstraint):
     def is_simple(self) -> bool:
         return True
 
+    def has_upper_bound(self) -> bool:
+        # Rationale:
+        # 1. If no version can satisfy the constraint,
+        #    this is like an upper bound of 0 (not included).
+        # 2. The opposite of an empty constraint, which is *, has no upper bound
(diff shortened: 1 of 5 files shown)


radoering:
@sourcery-ai review

src/poetry/core/constraints/version/empty_constraint.py
```diff
@@ -22,6 +22,9 @@ def is_any(self) -> bool:
     def is_simple(self) -> bool:
         return True
 
+    def has_upper_bound(self) -> bool:
+        return True
```
abn, line 26: Sanity question, does an empty constraint have a max?

radoering: Good question. My thinking was that the opposite of an empty constraint, which is `*`, has no upper bound and the two extremes often behave the other way around. Another explanation might be that if there is no version that can satisfy the constraint, this is like an upper bound of `0` (not included).

Further, it should practically irrelevant - at least for now.

abn: That sort of makes sense to me. 

Nitpick, maybe add a comment in there so someone doesn't change it later?

abn (approved):

PR #837, opened 2025-02-10
feat: support nesting dependency groups

finswimmer:
Include dependency groups into other dependency groups was requested earlier, e.g. https://github.com/python-poetry/poetry/issues/7494. Also PEP 735 describes this mechanism.

- [x] Updated **documentation** for changed code -> https://github.com/python-poetry/poetry/pull/10166

## Summary by Sourcery

New Features:
- Allow including dependencies from other groups within a dependency group.

## Summary by Sourcery

Add support for nested dependency groups.

New Features:
- Allow dependency groups to include other groups.

Tests:
- Add tests for nested dependency groups.

## Summary by Sourcery

Add support for nested dependency groups.

New Features:
- Allow dependency groups to include other groups.

Tests:
- Add tests for nested dependency groups.

### DIFF
--- a/src/poetry/core/factory.py
+++ b/src/poetry/core/factory.py
@@ -350,6 +350,20 @@ class Factory:
                     dependencies=group_config["dependencies"],
                 )
 
+            for group_name, group_config in tool_poetry["group"].items():
+                if include_groups := group_config.get("include-groups", []):
+                    current_group = package.dependency_group(group_name)
+                    for name in include_groups:
+                        try:
+                            group_to_include = package.dependency_group(name)
+                        except ValueError as e:
+                            raise ValueError(
+                                f"Group '{group_name}' includes group '{name}'"
+                                " which is not defined."
+                            ) from e
+
+                        current_group.include_dependency_group(group_to_include)
+
         if with_groups and "dev-dependencies" in tool_poetry:
             cls._add_package_group_dependencies(
                 package=package,
@@ -614,6 +628,8 @@ class Factory:
                 ' Use "poetry.group.dev.dependencies" instead.'
             )
 
+        cls._validate_dependency_groups_includes(toml_data, result)
+
         if strict:
             # Validate relation between [project] and [tool.poetry]
             cls._validate_legacy_vs_project(toml_data, result)
@@ -621,6 +637,35 @@ class Factory:
             cls._validate_strict(config, result)
 
         return result
+
+    @classmethod
+    def _validate_dependency_groups_includes(
+        cls, toml_data: dict[str, Any], result: dict[str, list[str]]
+    ) -> None:
+        """Ensure that dependency groups do not include themselves."""
+        config = toml_data.setdefault("tool", {}).setdefault("poetry", {})
+
+        group_includes = {}
+        for group_name, group_config in config.get("group", {}).items():
+            if include_groups := group_config.get("include-groups", []):
+                group_includes[canonicalize_name(group_name)] = [
+                    canonicalize_name(name) for name in include_groups
+                ]
+
+        for group_name in group_includes:
+            visited = set()
+            stack = [group_name]
+            while stack:
+                group = stack.pop()
+                visited.add(group)
+                for include in group_includes.get(group, []):
+                    if include in visited:
+                        result["errors"].append(
+                            f"Cyclic dependency group include in {group_name}:"
+                            f" {group} -> {include}"
+                        )
+                        continue
+                    stack.append(include)
 
     @classmethod
     def _validate_legacy_vs_project(
--- a/src/poetry/core/factory.py
+++ b/src/poetry/core/factory.py
@@ -347,7 +347,7 @@ class Factory:
                 cls._add_package_group_dependencies(
                     package=package,
                     group=group,
-                    dependencies=group_config["dependencies"],
+                    dependencies=group_config.get("dependencies", {}),
                 )
 
             for group_name, group_config in tool_poetry["group"].items():
--- a/src/poetry/core/factory.py
+++ b/src/poetry/core/factory.py
@@ -355,8 +355,8 @@ class Factory:
                     current_group = package.dependency_group(group_name)
                     for name in include_groups:
                         try:
-                            # `name` isn't normalized, but `.dependency_group()`
-                            # handles that.
+                            # `name` isn't normalized,
+                            # but `.dependency_group()` handles that.
                             group_to_include = package.dependency_group(name)
                         except ValueError as e:
                             raise ValueError(
--- a/src/poetry/core/factory.py
+++ b/src/poetry/core/factory.py
@@ -355,6 +355,8 @@ class Factory:
                     current_group = package.dependency_group(group_name)
                     for name in include_groups:
                         try:
+                            # `name` isn't normalized, but `.dependency_group()`
+                            # handles that.
                             group_to_include = package.dependency_group(name)
                         except ValueError as e:
                             raise ValueError(
@@ -645,7 +647,7 @@ class Factory:
         """Ensure that dependency groups do not include themselves."""
         config = toml_data.setdefault("tool", {}).setdefault("poetry", {})
 
-        group_includes = {}
+        group_includes: dict[NormalizedName, list[NormalizedName]] = {}
         for group_name, group_config in config.get("group", {}).items():
             if include_groups := group_config.get("include-groups", []):
                 group_includes[canonicalize_name(group_name)] = [
@@ -653,18 +655,32 @@ class Factory:
                 ]
 
         for group_name in group_includes:
-            visited = set()
+            ancestors: defaultdict[NormalizedName, set[NormalizedName]] = defaultdict(
+                set
+            )
             stack = [group_name]
             while stack:
                 group = stack.pop()
-                visited.add(group)
+                current_ancestors = ancestors.get(group, set())
+                child_ancestors = current_ancestors.union({group})
+
                 for include in group_includes.get(group, []):
-                    if include in visited:
+                    if include in child_ancestors:
                         result["errors"].append(
                             f"Cyclic dependency group include in {group_name}:"
                             f" {group} -> {include}"
                         )
+                        # Avoid infinite loop; we've already found an error.
                         continue
+
+                    # This must be a union with any existing known ancestors.
+                    # Otherwise, we might accidentally miss a cycle (since we'd
+                    # be tossing out known dependencies). That cycle will be
+                    # caught when we use the share dependency as `group_name`,
+                    # but best to be safe.
+                    ancestors[include] = ancestors.get(include, set()).union(
+                        child_ancestors
+                    )
                     stack.append(include)
 
     @classmethod
--- a/src/poetry/core/factory.py
+++ b/src/poetry/core/factory.py
@@ -654,34 +654,22 @@ class Factory:
                     canonicalize_name(name) for name in include_groups
                 ]
 
-        for group_name in group_includes:
-            ancestors: defaultdict[NormalizedName, set[NormalizedName]] = defaultdict(
-                set
-            )
-            stack = [group_name]
+        for root in group_includes:
+            # group, path to group, ancestors
+            stack: list[
+                tuple[NormalizedName, list[NormalizedName], set[NormalizedName]]
+            ] = [(root, [], {root})]
             while stack:
-                group = stack.pop()
-                current_ancestors = ancestors.get(group, set())
-                child_ancestors = current_ancestors.union({group})
-
+                group, path, ancestors = stack.pop()
                 for include in group_includes.get(group, []):
-                    if include in child_ancestors:
+                    new_path = [*path, include]
+                    if include in ancestors:
                         result["errors"].append(
-                            f"Cyclic dependency group include in {group_name}:"
-                            f" {group} -> {include}"
+                            f"Cyclic dependency group include in {root}:"
+                            f" {' -> '.join(new_path)}"
                         )
-                        # Avoid infinite loop; we've already found an error.
-                        continue
-
-                    # This must be a union with any existing known ancestors.
-                    # Otherwise, we might accidentally miss a cycle (since we'd
-                    # be tossing out known dependencies). That cycle will be
-                    # caught when we use the share dependency as `group_name`,
-                    # but best to be safe.
-                    ancestors[include] = ancestors.get(include, set()).union(
-                        child_ancestors
-                    )
-                    stack.append(include)
+                    else:
+                        stack.append((include, new_path, ancestors | {include}))
 
     @classmethod
     def _validate_legacy_vs_project(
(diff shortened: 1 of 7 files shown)


finswimmer:
@sourcery-ai review

finswimmer:
@sourcery-ai review

tests/test_factory.py
```diff
@@ -1060,3 +1060,130 @@ def test_poetry_build_system_dependencies(
...
+    ] == [
+        ("pytest", "testing"),
+        ("pytest-cov", "testing"),
+        ("black", "dev"),
+        ("pytest", "dev"),
+        ("pytest-cov", "dev"),
+    ]
+
+
+def test_create_poetry_with_invalid_nested_dependency_groups(
```
radoering, line 1098: We should test three variants of this:

- self-reference
- direct cycle (that is what is tested now)
- indirect cycle (with three groups)

(and as mentioned in the other comment improve the error message)

radoering (changes requested):

finswimmer:
@sourcery-ai review

tests/test_factory.py
finswimmer: Test case added.

src/poetry/core/factory.py
```diff
@@ -622,6 +638,33 @@ def validate(
...
+                ]
+
+        for group_name in group_includes:
+            visited = set()
+            stack = [group_name]
+            while stack:
+                group = stack.pop()
+                if group in visited:
+                    result["errors"].append(
+                        f"Dependency group '{group}' includes itself."
```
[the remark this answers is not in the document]
finswimmer: I improved the error message.

src/poetry/core/packages/dependency_group.py
```diff
@@ -22,35 +25,42 @@ def __init__(
...
                     *self._dependencies,
                     *(d for d in self._poetry_dependencies if d.is_optional()),
+                    *included_group_dependencies,
                 ]
-        return self._dependencies
+        return self._dependencies + included_group_dependencies
 
     @property
     def dependencies_for_locking(self) -> list[Dependency]:
+        included_group_dependencies = self._resolve_included_dependency_groups()
```
radoering, line 58: It might make no difference at the moment, but I think this is wrong because `_resolve_included_dependency_groups` uses `dependencies` instead of `dependencies_for_locking`. I think we need two versions of `_resolve_included_dependency_groups` (or pass a parameter that tells which dependencies to use).

src/poetry/core/packages/dependency_group.py
```diff
@@ -22,35 +25,42 @@ def __init__(
...
         return self._name
 
     @property
     def dependencies(self) -> list[Dependency]:
+        included_group_dependencies = self._resolve_included_dependency_groups()
+
         if not self._dependencies:
             # legacy mode
-            return self._poetry_dependencies
+            return self._poetry_dependencies + included_group_dependencies
```
radoering, line 40: Maybe, we should switch from multiple returns to one return in the end so that we just have to add `included_group_dependencies` in one place instead of four?

src/poetry/core/packages/dependency_group.py
```diff
@@ -3,6 +3,9 @@
 from collections import defaultdict
 from typing import TYPE_CHECKING
 
+from poetry.core._vendor.packaging.utils import NormalizedName
+from poetry.core._vendor.packaging.utils import canonicalize_name
```
radoering, line 7: Wrong imports.

tests/test_factory.py
```diff
@@ -1060,3 +1060,247 @@ def test_poetry_build_system_dependencies(
...
+    ] == [
+        ("pytest", "testing"),
+        ("pytest-cov", "testing"),
+        ("black", "dev"),
+        ("pytest", "dev"),
+        ("pytest-cov", "dev"),
+    ]
+
+
+def test_create_poetry_with_self_referenced_dependency_groups(
```
radoering, line 1098: `parametrize` with different names (test normalization).

tests/test_factory.py
```diff
@@ -1060,3 +1060,247 @@ def test_poetry_build_system_dependencies(
...
+        _ = Factory().create_poetry(temporary_directory)
+
+    expected = """\
+The Poetry configuration is invalid:
+  - Cyclic dependency group include downstream the definition of testing: testing -> testing
+"""
+    assert str(error.value) == expected
+
+
+def test_create_poetry_with_direct_cycled_dependency_groups(
```
radoering, line 1127:
```suggestion
def test_create_poetry_with_direct_cyclic_dependency_groups(
```

tests/test_factory.py
```diff
@@ -1060,3 +1060,247 @@ def test_poetry_build_system_dependencies(
...
+
+    expected = """\
+The Poetry configuration is invalid:
+  - Cyclic dependency group include downstream the definition of testing: dev -> testing
+  - Cyclic dependency group include downstream the definition of dev: testing -> dev
+"""
+    assert str(error.value) == expected
+
+
+def test_create_poetry_with_indirect_full_cycled_dependency_groups(
```
radoering, line 1164:
```suggestion
def test_create_poetry_with_indirect_full_cyclic_dependency_groups(
```

tests/test_factory.py
```diff
@@ -1060,3 +1060,247 @@ def test_poetry_build_system_dependencies(
...
+    expected = """\
+The Poetry configuration is invalid:
+  - Cyclic dependency group include downstream the definition of group-1: group-2 -> group-1
+  - Cyclic dependency group include downstream the definition of group-2: group-3 -> group-2
+  - Cyclic dependency group include downstream the definition of group-3: group-1 -> group-3
+"""
+    assert str(error.value) == expected
+
+
+def test_create_poetry_with_indirect_partial_cycled_dependency_groups(
```
radoering, line 1208:
```suggestion
def test_create_poetry_with_indirect_partial_cyclic_dependency_groups(
```

tests/test_factory.py
```diff
@@ -1060,3 +1060,247 @@ def test_poetry_build_system_dependencies(
     poetry = Factory().create_poetry(temporary_directory)
 
     assert set(poetry.build_system_dependencies) == expected
+
+
+def test_create_poetry_with_nested_dependency_groups(temporary_directory: Path) -> None:
```
radoering, line 1065: `parametrize` with different names (test normalization).

radoering (changes requested):

src/poetry/core/packages/dependency.py
```diff
@@ -329,6 +329,10 @@ def with_constraint(self: T, constraint: str | VersionConstraint) -> T:
         dependency.constraint = constraint  # type: ignore[assignment]
         return dependency
 
+    def move_to_groups(self, groups: Iterable[str]) -> None:
```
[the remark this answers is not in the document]
finswimmer: :+1: I wonder if there is any risk if we make the `_groups` Attribut public :thinking:

src/poetry/core/packages/dependency_group.py
finswimmer: Actually I guess I need more information about what's the difference between `dependencies` and `dependencies_for_locking` :sweat_smile:

tests/test_factory.py
finswimmer: I find it hard to parametrize those test, without hiding what is actually tested to much. I've introduce a helper method to avoid some code duplication. Now I find one could follow the purpose of the tests, there input and the expected outcome quite good.

src/poetry/core/packages/dependency_group.py
finswimmer: I tried something ...

tests/packages/test_dependency_group.py
```diff
@@ -491,3 +491,52 @@ def test_dependencies_for_locking_failure(
...
+        assert dep.groups == {"group"}
+
+    assert [dep.name for dep in group_2.dependencies] == ["bar"]
+    assert [dep.name for dep in group_2.dependencies_for_locking] == ["bar"]
+    for dep in group_2.dependencies:
+        assert dep.groups == {"group2"}
+
+
+@pytest.mark.parametrize("group_name", ["group_2", "Group-2", "group-2"])
+def test_include_dependency_group_raise_if_already_included(group_name: str) -> None:
```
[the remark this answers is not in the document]
finswimmer: Testcase is added.

src/poetry/core/packages/dependency_group.py
radoering: I guess you do not need to know the difference to consider it wrong: It feels clear to me that nesting groups should not violate the distinction between the two.

Nevertheless: `dependencies` are used for metadata (when building), `dependencies_for_locking` are used for locking. This distinction is probably not that relevant for groups because we only need `dependencies_for_locking` and do not use `dependencies` but that does not mean that we should not care about implementing it correctly.

src/poetry/core/packages/dependency_group.py
radoering: Looks good. The same for `dependencies_for_locking`, please. 😃

tests/test_factory.py
radoering: As long as at least one test fails if I remove (exactly) one arbitrary `canonicalize_name`, I am fine.

src/poetry/core/packages/dependency.py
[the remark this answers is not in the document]
radoering: No more risk than introducing a method, which sets it. I think, in accordance with the existing code, we would:

- not introduce a public method setting it
- just set `_groups` in the `with_groups` method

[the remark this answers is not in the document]
finswimmer: I changed `_groups`  to a public attribute. I doesn't feel right to use a private attribute from outside.

tests/test_factory.py
finswimmer: Ah, now I guess I get what you want to parametrize. :sweat_smile: I added some.

tests/packages/test_dependency_group.py
```diff
@@ -491,3 +491,64 @@ def test_dependencies_for_locking_failure(
...
+    for dep in group.dependencies:
+        assert dep.groups == {"group"}
+
+    assert [dep.name for dep in group_2.dependencies] == ["bar"]
+    assert [dep.name for dep in group_2.dependencies_for_locking] == ["bar"]
+    for dep in group_2.dependencies:
+        assert dep.groups == {"group2"}
+
+
+def test_include_dependency_group_raise_if_including_itself() -> None:
```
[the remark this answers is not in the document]
finswimmer: Done

src/poetry/core/factory.py
```diff
@@ -622,6 +638,35 @@ def validate(
...
+
+        for group_name in group_includes:
+            visited = set()
+            stack = [group_name]
+            while stack:
+                group = stack.pop()
+                visited.add(group)
+                for include in group_includes.get(group, []):
+                    if include in visited:
+                        result["errors"].append(
```
citelao, line 663: I think this algorithm will provide a false error if presented a specific diamond pattern:

* a
  * b1
     * shared 
  * b2
    * c
       * shared

It'll visit `a`, `b1`, `b2`, `shared`, and then it will fail on `c`.

---

In testing, this also fails with:


* a
  * b1
     * shared 
  * b2
     * shared

---

I think the simplest algorithm is to keep a list of ancestors for every group you visit, verifying there are no duplicates.

citelao:
I attempted to address some of the open feedback: https://github.com/finswimmer/core/pull/2. Hope it helps; I'm excited to see this change.

src/poetry/core/factory.py
```diff
@@ -347,9 +347,23 @@ def _configure_package_dependencies(
                 cls._add_package_group_dependencies(
                     package=package,
                     group=group,
-                    dependencies=group_config["dependencies"],
+                    dependencies=group_config.get("dependencies", {}),
                 )
 
+            for group_name, group_config in tool_poetry["group"].items():
```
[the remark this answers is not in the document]
citelao: Added a parameterization that tests that version: https://github.com/finswimmer/core/pull/2.

tests/test_factory.py
citelao: Done: https://github.com/finswimmer/core/pull/2

src/poetry/core/factory.py
citelao: Fixed: https://github.com/finswimmer/core/pull/2

src/poetry/core/packages/dependency_group.py
citelao: I've attempted to address this issue in https://github.com/finswimmer/core/pull/2 & added a test for this case.

citelao:
Hi @radoering, @finswimmer---is there any way I can help get this work in? I've addressed most of the feedback. Happy to get my PR integrated into this branch or start my own PR; whatever is easiest for the team.

citelao:
I've rebased my code!

radoering:
> I've rebased my code!

Thanks, I started reviewing it and will continue next week.

radoering:
I assume the CI failures on Windows may be unrelated to this PR but we have to sort it out anyway.

radoering:
> I assume the CI failures on Windows may be unrelated to this PR but we have to sort it out anyway.

Looks like the `actions/cache` was corrupt. Removing the cache did the trick.

tests/test_factory.py
```diff
@@ -1274,7 +1351,266 @@ def test_create_poetry_with_indirect_partial_cyclic_dependency_groups(
...
+    ) == [
+        ("bar", "child-1"),
+        ("bar", "root"),
+        ("baz", "child-2"),
+        ("baz", "root"),
+        ("foo", "root"),
+        ("quux", "child-1"),
+        ("quux", "child-2"),
+        # Duplicates because dependency is included via several groups.
+        # This is ok because they are merged during depenendency resolution.
```
citelao, line 1420: Nit: `dependency`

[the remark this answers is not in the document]
radoering: > I assume this is already tested? I'm new to this codebase. As long as there's a test for this, I'm happy.

This is only relevant for locking not for building. That is why it is tested downstream. There should be some tests in https://github.com/python-poetry/poetry/blob/main/tests/puzzle/test_solver.py.

> As a longer answer, I _don't_ think the PEP really considered this diamond-dependency pattern (I think it focused more on isolated groups of dependencies), so I think it would be _reasonable_ to remove this literally-identical dependencies (also, `SHOULD NOT` grants us that flexibility according to [RFC2119](https://www.rfc-editor.org/rfc/rfc2119)).

Agreed.

> However, I tried this, and the equality implementation for `Dependency` is not complete; it only compares a few of the properties (some of the constraints were not compared; can't remember which). So I left it as-is.

Yes, that is also why I did not bother with trying to remove duplicates. I think this may open a can of worms. There are surely spots in the code where we rely on the implementation of `__eq__` not considering all attributes.

citelao:
Woohoo! Thanks for this :)

PR #843, opened 2025-02-23
make complex marker unions and intersections deterministic

radoering:
Requires: #842

## Summary by Sourcery

This pull request makes complex marker unions and intersections deterministic by preventing duplicate constraints and improving performance.

Enhancements:
- Ensures that marker unions and intersections are deterministic by avoiding duplicate constraints when intersecting union constraints.
- Improves the performance of marker operations by using generators instead of sets to determine unique markers.

Tests:
- Adds a test to ensure that complex marker unions are deterministic, preventing sporadic test failures.

### DIFF
--- a/tests/version/test_markers.py
+++ b/tests/version/test_markers.py
@@ -2154,6 +2154,35 @@ def test_complex_intersection() -> None:
     )
 
 
+def test_complex_union_is_deterministic() -> None:
+    """
+    This test might fail sporadically if marker operations are not deterministic!
+    """
+    m1 = parse_marker(
+        'sys_platform != "darwin" and python_version >= "3.12"'
+        ' and platform_system != "Emscripten" and (python_version < "4.0"'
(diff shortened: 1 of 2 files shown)


tests/version/test_markers.py
```diff
@@ -2154,6 +2154,35 @@ def test_complex_intersection() -> None:
     )
 
 
+def test_complex_union_is_deterministic() -> None:
+    """
+    This test might fail sporadically if marker operations are not deterministic!
```
Secrus, line 2159: Does that mean it can fail because of anything, or will it detect changes to the marker operations code that would make them undeterministic?

radoering: If the marker operations code becomes undeterministic there will be a good chance that the test fails sporadically. As long as the marker operations code is determistic the test will not fail.

Secrus (approved):

PR #849, opened 2025-03-08
markers: recognize empty `python_version` markers

radoering:
Without this fix, the intersection of `python_version > "3.8"` and `python_version < "3.9"` is not detected as empty marker.

Related to: python-poetry/poetry#10249

## Summary by Sourcery

Tests:
- Add tests to verify that the intersection of mutually exclusive markers is correctly identified as empty.

### DIFF
--- a/tests/version/test_markers.py
+++ b/tests/version/test_markers.py
@@ -172,6 +172,22 @@ def test_single_marker_intersect() -> None:
 
 
 @pytest.mark.parametrize(
+    ("marker1", "marker2"),
+    [
+        ('sys_platform == "win32"', 'sys_platform == "linux"'),
+        ('sys_platform != "win32"', 'sys_platform == "win32"'),
+        ('python_version >= "3.6"', 'python_version < "3.6"'),
+        ('python_version == "3.6"', 'python_version == "3.7"'),
+        ('python_version > "3.6"', 'python_version <= "3.6"'),
+        ('python_version >= "3.6"', 'python_version < "3.6"'),
+        ('python_version > "3.6"', 'python_version < "3.7"'),
+    ],
+)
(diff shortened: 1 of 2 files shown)


tests/version/test_markers.py
```diff
@@ -171,6 +171,22 @@ def test_single_marker_intersect() -> None:
...
+@pytest.mark.parametrize(
+    ("marker1", "marker2"),
+    [
+        ('sys_platform == "win32"', 'sys_platform == "linux"'),
+        ('sys_platform != "win32"', 'sys_platform == "win32"'),
+        ('python_version >= "3.6"', 'python_version < "3.6"'),
+        ('python_version == "3.6"', 'python_version == "3.7"'),
+        ('python_version > "3.6"', 'python_version <= "3.6"'),
+        ('python_version >= "3.6"', 'python_version < "3.6"'),
+        ('python_version > "3.6"', 'python_version < "3.7"'),
```
radoering, line 183: The last test is the relevant one but since I did not find such a basic test function that tests for empty single marker intersections, I just added some additional tests.

tests/version/test_markers.py
```diff
@@ -2260,16 +2271,16 @@ def test_intersection_no_endless_recursion() -> None:
         ' and extra != "postgis"'
     )
     m2 = parse_marker(
-        'python_version > "3.12" and python_version < "3.13" or extra != "databricks"'
+        'python_version >= "3.12" and python_version < "3.13" or extra != "databricks"'
```
radoering, line 2274: This marker is actually empty and does not trigger a potential endless recursion anymore. I checked that the changed marker triggers the recursion error without the fix from #832.

Secrus (approved):

PR #850, opened 2025-03-10
fix: :bug: do not exclude files in exlplicitly included directorys

Katze719:
Resolves: python-poetry#10259


Katze719:
NVM using poetry_test_issue_10259/myLibs/**/* works 🙃🙃🙃

do what ever you want with this code

Katze719:
maybe, could we add to the documentation that globing is supported for including/excluding files within the pyproject.toml?

PR #868, opened 2025-05-10
canonicalize dependency group names

radoering:
Closes:  #852

Alternative to #852 with some small changes:

- do not introduce `CanonicalizedDict` because _in my opinion_ it is a bit overkill and I am not sure if would help to avoid errors or would be a source of new errors
- normalize groups of a dependency
- add test for `dependency_group_names`

## Summary by Sourcery

Canonicalize dependency group names to provide case-insensitive and normalized handling of dependency groups across the package management system.

New Features:
- Add support for case-insensitive dependency group name matching

Enhancements:
- Normalize dependency group names using canonicalization
- Introduce case-insensitive lookup for dependency groups

Tests:
- Add tests to verify case-insensitive dependency group handling
- Add tests for dependency group name normalization

### DIFF
--- a/src/poetry/core/packages/dependency_group.py
+++ b/src/poetry/core/packages/dependency_group.py
@@ -3,6 +3,9 @@ from __future__ import annotations
 from collections import defaultdict
 from typing import TYPE_CHECKING
 
+from packaging.utils import NormalizedName
+from packaging.utils import canonicalize_name
+
 from poetry.core.version.markers import parse_marker
 
 
@@ -10,22 +13,27 @@ if TYPE_CHECKING:
     from poetry.core.packages.dependency import Dependency
     from poetry.core.version.markers import BaseMarker
 
-MAIN_GROUP = "main"
+MAIN_GROUP: NormalizedName = canonicalize_name("main")
 
 
 class DependencyGroup:
     def __init__(
         self, name: str, *, optional: bool = False, mixed_dynamic: bool = False
     ) -> None:
-        self._name: str = name
+        self._name: NormalizedName = canonicalize_name(name)
+        self._pretty_name: str = name
         self._optional: bool = optional
         self._mixed_dynamic = mixed_dynamic
         self._dependencies: list[Dependency] = []
         self._poetry_dependencies: list[Dependency] = []
 
     @property
-    def name(self) -> str:
+    def name(self) -> NormalizedName:
         return self._name
+
+    @property
+    def pretty_name(self) -> str:
+        return self._pretty_name
 
     @property
     def dependencies(self) -> list[Dependency]:
@@ -126,7 +134,11 @@ class DependencyGroup:
 
     def __repr__(self) -> str:
         cls = self.__class__.__name__
-        return f"{cls}({self._name}, optional={self._optional})"
+        return (
+            f"{cls}({self._pretty_name!r},"
+            f" optional={self._optional},"
+            f" mixed_dynamic={self._mixed_dynamic})"
+        )
 
 
 def _enrich_dependency(
--- a/tests/packages/test_package.py
+++ b/tests/packages/test_package.py
@@ -8,6 +8,8 @@ from typing import cast
 
 import pytest
 
+from packaging.utils import canonicalize_name
+
 from poetry.core.constraints.version import Version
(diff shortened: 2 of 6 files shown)


tests/packages/test_package.py
```diff
@@ -667,3 +692,54 @@ def test_package_empty_python_versions() -> None:
...
+    [("optional", True), ("Optional", True), ("OpTiOnAl", True), ("other", False)],
+)
+def test_package_has_dependency_group(
+    package_with_groups: Package, group_name: str, expected: bool
+) -> None:
+    assert package_with_groups.has_dependency_group(group_name) is expected
+
+
+@pytest.mark.parametrize("group_name", ["optional", "Optional", "OpTiOnAl"])
+def test_package_get_dependency_group(
```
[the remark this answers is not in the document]
radoering: out of scope of this PR

src/poetry/core/packages/dependency_group.py
```diff
@@ -3,30 +3,38 @@
...
         self._dependencies: list[Dependency] = []
         self._poetry_dependencies: list[Dependency] = []
 
     @property
-    def name(self) -> str:
+    def name(self) -> NormalizedName:
         return self._name
 
+    @property
+    def pretty_name(self) -> str:
```
citelao, line 35: Is `pretty_name` the clearest name for this property? I'd expect something like `original_name`.

I can undestand this causing confusion for folks (what property should I use?), but I think you've mitigated that by preferring `NormalizedName` in API signatures.

citelao: (I see that's aligned with `Package`)

src/poetry/core/packages/dependency_group.py
```diff
@@ -3,30 +3,38 @@
...
 class DependencyGroup:
     def __init__(
         self, name: str, *, optional: bool = False, mixed_dynamic: bool = False
     ) -> None:
-        self._name: str = name
+        self._name: NormalizedName = canonicalize_name(name)
+        self._pretty_name: str = name
         self._optional: bool = optional
         self._mixed_dynamic = mixed_dynamic
         self._dependencies: list[Dependency] = []
```
citelao, line 27: Are `Dependency` names already normalized? I see fns like `remove_dependency` assume they are.

src/poetry/core/packages/dependency_group.py
```diff
@@ -3,30 +3,38 @@
...
+MAIN_GROUP: NormalizedName = canonicalize_name("main")
 
 
 class DependencyGroup:
     def __init__(
         self, name: str, *, optional: bool = False, mixed_dynamic: bool = False
     ) -> None:
-        self._name: str = name
+        self._name: NormalizedName = canonicalize_name(name)
+        self._pretty_name: str = name
```
citelao, line 24: You may want to update `__repr__` for all of these classes (should they use `pretty_name`?)

citelao (approved):
I'm not sure I have enough context to give you a high-quality review, but I took a look and I think this change looks good!

src/poetry/core/packages/dependency_group.py
radoering: Yes, I chose the name for consistency, which is more important to me than the actual name. If we decided that `pretty_name` is not a good name, we should deprecate it in `Package` and introduce the new name there, too. (In my opinion, `pretty_name` is not that bad that it is worth the trouble.)

src/poetry/core/packages/dependency_group.py
radoering: `Dependency` inherits from `PackageSpecification`, which has a normalized `name` and an unnormalized `pretty_name`.

src/poetry/core/packages/dependency_group.py
radoering: Good question. I have to check later what impact that would have.

src/poetry/core/packages/dependency_group.py
citelao: Seems good to me :)

src/poetry/core/packages/dependency_group.py
citelao: Any answer makes sense to me; I called it out simply as something I noticed.

finswimmer (approved):

PR #879, opened 2025-08-23
fix python constraint derived from markers to better support pre-release versions of Python

radoering:
https://github.com/python-poetry/poetry/actions/runs/17171517004/job/48721759731?pr=10514 (failure in `tests/utils/test_isolated_build.py::test_isolated_env_install_discards_requirements_not_needed_by_env` with Python 3.14rc2) shows that our installer does not handle environments of Python pre-release versions well. A dependency with the marker `python_version == "3.14"` is not installed because it is translated into a Python constraint `>=3.14,<3.15`, which does not allow Python 3.14rc2. With this PR the Python constraint is `>=3.14.dev0,<3.15`.

Downstream tests are fixed in python-poetry/poetry#10515

## Summary by Sourcery

Improve marker-to-constraint translation to properly support Python pre-release versions by introducing `.dev0` bounds and wildcard expansions for partial-precision markers

New Features:
- Support pre-release Python versions in marker-derived constraints by using `.dev0` for lower bounds

Enhancements:
- Normalize Python version markers to expand `==x` and `!=x` to `x.*` and append `.dev0` to `>=x` bounds when precision is less than two segments
- Treat marker-based `>=` constraints as stable versions in the version parser for correct handling of dev releases

Tests:
- Add tests for parsing and formatting of `.dev0`-based version constraints and update existing expectations to match new wildcard and dev0 conventions

### DIFF
--- a/src/poetry/core/constraints/version/parser.py
+++ b/src/poetry/core/constraints/version/parser.py
@@ -180,6 +180,8 @@ def parse_single_constraint(
         if op == ">":
             return VersionRange(min=version)
         if op == ">=":
+            if is_marker_constraint:
+                version = version.stable
             return VersionRange(min=version, include_min=True)
 
         if m.group("wildcard") is not None:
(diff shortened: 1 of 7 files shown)


dimbleby:
Discussion [here](https://discuss.python.org/t/requires-python-and-pre-release-python-versions/62959/3)

Tldr the spec is unhelpful, insofar as there is a consensus it probably is that poetry's current behaviour is correct - but the change you are making is desirable and the spec ought to be changed

However I see no such change has happened.

PR #904, opened 2025-12-28
improve type hint for `Package.files`

radoering:
It is a matter of weighing things up but I believe checking the keys is worth more than checking immutability.

See python-poetry/poetry#10673 for issues that the improved typing uncovered.

## Summary by Sourcery

Improve typing of package file metadata on packages.

Enhancements:
- Introduce a PackageFile TypedDict to describe package file metadata.
- Tighten the Package.files attribute type to use the new PackageFile TypedDict instead of a generic mapping.

### DIFF
--- a/src/poetry/core/packages/package.py
+++ b/src/poetry/core/packages/package.py
@@ -4,6 +4,7 @@ import warnings
 
 from typing import TYPE_CHECKING
 from typing import ClassVar
+from typing import TypedDict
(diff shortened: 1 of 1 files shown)


dimbleby:
> checking the keys is worth more than checking immutability

`typing.ReadOnly` (backported to `typing_extensions`) should allow you to do both

PR #950, opened 2026-06-06
Fix union(range, public_version) to fill local-variant punctures

dimbleby:
The constraint algebra treated '==X' as the literal point [X, X] when computing 'range.union(public_X)', returning the range unchanged whenever the range already allowed X. But per PEP 440 release-equality '==X' also matches every local-tagged variant 'X+local', so a range whose bound is such a variant (e.g. '>=X,<X+local' or '>X+local,<Y') must be broadened to include those variants in the union.

Apply the broadening symmetrically in VersionRange.union(Version), handling both bounds in a single pass, and delegate Version.union(non Version) to the other operand so the rules apply regardless of operand order.

Resolves: python-poetry#

### DIFF
--- a/src/poetry/core/constraints/version/version_range.py
+++ b/src/poetry/core/constraints/version/version_range.py
@@ -215,6 +215,9 @@ class VersionRange(VersionRangeConstraint):
                 and not other.is_local()
                 and other.allows(self.min)
             ):
+                # Strictly speaking, next_patch() is not quite correct
+                # because you cannot specify the upper bound. It only
+                # works for versions with a precision of three or less.
                 upper = other.stable.next_patch()
                 return _range_or_empty(
                     min=self.min,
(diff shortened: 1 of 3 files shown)


src/poetry/core/constraints/version/version_range.py
```diff
@@ -260,6 +260,39 @@ def union(self, other: VersionConstraint) -> VersionConstraint:
...
+                ):
+                    new_min = other
+                    new_include_min = True
+                    broadened = True
+                if (
+                    self._max is not None
+                    and self._max.is_local()
+                    and other.allows(self._max)
+                ):
+                    new_max = other.stable.next_patch()
```
radoering, line 303: This is wrong for versions with a higher precision than 3. We have already had a discussion about that in https://github.com/python-poetry/poetry-core/pull/579#discussion_r1187581400 and decided that we accept it. (I will probably add a comment so that I do not have to dig again, the next time I am wondering that this is wrong.)

dimbleby: oh I had forgotten that wrinkle.  python versions really are the worst!

radoering (approved):

PR #955, opened 2026-07-16
fix: correct allows_all containment check for "!=" vs "not in"

### LINKED ISSUE #10977: poetry-core drops the `!=` term from `x != "a" and "b" not in x`
### Description

I'm working on marker logic and doing a comparison of different libraries, while doing a scan of poetry, I found a couple of bugs.

`Constraint.allows_all` decides that `x != V` is implied by `W not in x` using the wrong containment test. `parse_marker` then drops the `!=` term and `validate` returns the wrong result.

```python
>>> from poetry.core.version.markers import parse_marker
>>> m = parse_marker('os_name != "a" and "b" not in os_name')
>>> print(m)
"b" not in os_name
>>> print(m.validate({"os_name": "a"}))
True
```

### Workarounds

Don't use weird marker combinations, this is obscure, but that's because I'm doing a correctness analysis of all marker handling across the Python ecosystem.

### Poetry Installation Method

other

### Operating System

ubuntu 24.04 / Windows 11 / Anything

### Poetry Version

poetry-core 2.4.1

### Poetry Configuration

```bash session
I'm running all my tests via uvx so I think this isn't relevant but here you go:


Installed 41 packages in 362ms
cache-dir = "C:\\Users\\damia\\AppData\\Local\\pypoetry\\Cache"
data-dir = "C:\\Users\\damia\\AppData\\Roaming\\pypoetry"
installer.max-workers = null
installer.no-binary = null
installer.only-binary = null
installer.parallel = true
installer.re-resolve = false
keyring.enabled = true
python.installation-dir = "{data-dir}\\python"  # C:\Users\damia\AppData\Roaming\pypoetry\python
requests.max-retries = 0
solver.lazy-wheel = true
solver.min-release-age = 0
solver.min-release-age-exclude = null
solver.min-release-age-exclude-source = null
system-git-client = false
virtualenvs.create = true
virtualenvs.in-project = null
virtualenvs.options.always-copy = false
virtualenvs.options.no-pip = false
virtualenvs.options.system-site-packages = false
virtualenvs.path = "{cache-dir}\\virtualenvs"  # C:\Users\damia\AppData\Local\pypoetry\Cache\virtualenvs
virtualenvs.prompt = "{project_name}-py{python_version}"
virtualenvs.use-poetry-python = false
```

### Python Sysconfig

<details>
  <summary>sysconfig.log</summary>
  

  ```
  Paste the output of 'python -m sysconfig', over this line.
  ```
</details>

### Example pyproject.toml

```TOML

```

### Poetry Runtime Logs

<details>
  <summary>poetry-runtime.log</summary>
  

  ```
  Paste the output of 'poetry -vvv <command>', over this line.
  ```
</details>

hexonal:
Fixes python-poetry/poetry#10977

`Constraint.allows_all()` had a containment check backwards for the case where `self` is a `!=` constraint and `other` is a `not in` constraint. It tested `self.value not in other.value`, but the correct check has the same shape as the existing `not in` / `not in` branch just above it: `other.value in self.value`.

Concretely:

```python
>>> from poetry.core.version.markers import parse_marker
>>> m = parse_marker('os_name != "a" and "b" not in os_name')
>>> str(m)  # before: the `!=` clause silently disappears
'"b" not in os_name'
>>> m.validate({"os_name": "a"})  # before: True (wrong)
>>> m.validate({"os_name": "a"})  # after: False (correct)
```

This happened because `MultiMarker`/constraint intersection logic uses `allows_all` to drop a clause it believes is redundant. The old formula wrongly considered `os_name != "a"` redundant given `"b" not in os_name`, even though `os_name == "a"` satisfies the latter but not the former.

I merged the two branches for `self._operator == "not in"` and `self._operator == "!="` under `other.operator == "not in"` since they resolve to the identical correct formula.

Added regression tests in `tests/constraints/generic/test_constraint.py` (direct `allows_all` cases) and `tests/version/test_markers.py` (the exact marker/validate scenario from the issue, plus two adjacent cases).

### DIFF
--- a/src/poetry/core/constraints/generic/constraint.py
+++ b/src/poetry/core/constraints/generic/constraint.py
@@ -88,9 +88,8 @@ class Constraint(BaseConstraint):
             if other.operator == "in" and self._operator == "in":
                 return self.value in other.value
 
-            if other.operator == "not in":
-                if self._operator in ("not in", "!="):
-                    return other.value in self.value
+            if other.operator == "not in" and self._operator in ("not in", "!="):
+                return other.value in self.value
 
             return self == other
 
--- a/src/poetry/core/constraints/generic/constraint.py
+++ b/src/poetry/core/constraints/generic/constraint.py
@@ -89,10 +89,8 @@ class Constraint(BaseConstraint):
                 return self.value in other.value
 
             if other.operator == "not in":
-                if self._operator == "not in":
+                if self._operator in ("not in", "!="):
                     return other.value in self.value
-                if self._operator == "!=":
-                    return self.value not in other.value
 
             return self == other
 
--- a/tests/constraints/generic/test_constraint.py
+++ b/tests/constraints/generic/test_constraint.py
@@ -253,6 +253,21 @@ def test_allows(
             True,
             False,
         ),
+        # "!=" combined with "not in" where the excluded value is not a
+        # substring of the excluded-substring's value: allows_all must be
+        # False, since e.g. "a" itself satisfies "'b' not in x" but not "!= a"
+        (
+            Constraint("a", "!="),
+            Constraint("b", "not in"),
+            True,
+            False,
+        ),
+        (
+            Constraint("ab", "!="),
+            Constraint("a", "not in"),
+            True,
+            True,
+        ),
         (
             Constraint("1.2.3-tegra", "=="),
             Constraint("tegra", "in"),
--- a/tests/version/test_markers.py
+++ b/tests/version/test_markers.py
@@ -1466,6 +1466,10 @@ def test_multi_marker_removes_duplicates() -> None:
             {"platform_release": "4.9.253"},
             True,
         ),
+        # both terms must be kept, not just the "not in" one
+        ("os_name != 'a' and 'b' not in os_name", {"os_name": "a"}, False),
+        ("os_name != 'a' and 'b' not in os_name", {"os_name": "ab"}, False),
+        ("os_name != 'a' and 'b' not in os_name", {"os_name": "c"}, True),
         (
             "platform_release >= '6.6.0+rpt-rpi-v8'",
             {"platform_release": "6.6.20+rpt-rpi-v8"},


skdas20 (commented):
Checked this out locally — it fixes more than the reported crash, which I think strengthens the case for it.

Beyond #10977, the same containment logic silently produces a **wrong marker** (not an exception) whenever a `!=` and a `not in` on the same non-version-like name are ANDed. On `main` the `!=` term is dropped entirely:

```python
>>> from poetry.core.version.markers import parse_marker
>>> parse_marker('platform_version != "1" and "a" not in platform_version')
"a" not in platform_version                       # != term silently gone
>>> _.validate({'platform_version': '1'})
True                                               # should be False
```

Same on `sys_platform` (`!= "win32" and "lin" not in ...` wrongly matches `win32`) and `platform_machine`. That's the more dangerous failure mode of the two, since a too-permissive marker means a dependency gets installed on an environment it was meant to exclude, with nothing raised to signal it.

With this patch applied, all three evaluate correctly.

I also applied this together with #957 (the `platform_release` `AttributeError` fix). They don't conflict and address different paths — #957 covers version-like names where the two forms parse to different constraint types, this one covers generic constraints where they merge but merge wrongly. Full `tests/version` + `tests/constraints` pass with both: 1559 passed.

Worth adding the `validate` case above as a regression test here if you haven't already — the current tests would still pass if the term-dropping came back in a form that didn't crash.

hexonal:
Thanks for actually running it.

One scope note in case a triager reads this thread first: this PR is tied to python-poetry/poetry#10977, which is the silent dropped-term bug rather than a crash. The `AttributeError: 'Constraint' object has no attribute 'flatten'` is python-poetry/poetry#10978, and it still reproduces unchanged with this patch applied — #956/#957 are on that path. Your note over on #957 draws exactly that line, so this is really just for anyone landing here first.

On the regression test: the `validate` cases are already in this PR — `tests/version/test_markers.py::test_validate` gained `os_name != "a" and "b" not in os_name` and `os_name != "a" and "a" not in os_name`. I checked they actually bite rather than passing incidentally: with `git checkout main -- src/` and the PR's tests left in place, `pytest tests/version/test_markers.py tests/constraints/generic/test_constraint.py` gives 2 failed / 782 passed, both failures being those cases. So the term-dropping coming back in a non-crashing form is covered.

I'm not adding `platform_version` / `sys_platform` / `platform_machine` variants of it, for a code-path reason rather than a taste one: every marker name outside `_VERSION_LIKE_MARKER_NAME` (`markers.py:385-389`) reaches the identical `Constraint.allows_all` branch at `constraints/generic/constraint.py:91`, so they'd all exercise the same line. `os_name` is representative.

radoering (approved):

PR #957, opened 2026-07-20
fix: don't crash when merging incompatible platform_release marker forms

Cyrus580529:
Resolves: python-poetry/poetry-core#10978

`platform_release != "1"` parses to a version-based constraint, but the swapped form `"a" not in platform_release` parses to a generic string `Constraint` (this is intentional — the swapped form is a substring check, not a version comparison). `_merge_single_markers` assumed same-name markers always share a constraint type and crashed with `AttributeError: 'Constraint' object has no attribute 'flatten'` when trying to intersect the two.

Added a type check before merging: if the two constraints aren't both version-like, fall back to keeping them as separate ANDed/ORed markers (the existing fallback path for unmergeable markers), instead of crashing. Added regression tests.

- [x] Added tests for changed code.
- [ ] Updated documentation for changed code.

### DIFF
--- a/src/poetry/core/version/markers.py
+++ b/src/poetry/core/version/markers.py
@@ -1286,6 +1286,16 @@ def _merge_single_markers(
     if marker1.name != marker2.name:
         return None
 
+    if isinstance(marker1.constraint, VersionConstraint) != isinstance(
+        marker2.constraint, VersionConstraint
+    ):
+        # Same marker name, but the constraints come from incompatible parses:
+        # e.g. `platform_release != "1"` produces a VersionConstraint, while
+        # the swapped form `"a" not in platform_release` produces a generic
+        # Constraint. They can't be intersected/unioned together, so leave
+        # them as separate ANDed/ORed markers instead of crashing below.
+        return None
+
     if merge_class == MultiMarker:
         merge_method = marker1.constraint.intersect
     else:
--- a/tests/version/test_markers.py
+++ b/tests/version/test_markers.py
@@ -1476,6 +1476,25 @@ def test_multi_marker_removes_duplicates() -> None:
             {"platform_release": "4.9.254-tegra"},
             True,
         ),
+        # Mixing the plain (`!=`) and swapped (`not in`) forms of a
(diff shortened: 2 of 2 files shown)


skdas20 (commented):
I checked this out and ran it locally — the diagnosis and the fix both look right to me.

**Verified:** `tests/version/test_markers.py` is green (555 passed) with the patch applied.

I also poked at some neighbouring shapes that aren't in the added tests, and they all behave correctly with this change:

```
"a" not in platform_release and platform_release != "1"     # swapped form first
"a" not in platform_release or platform_release != "1"
"a" in platform_release and platform_release >= "2"          # 'in' + comparison
platform_release != "1" and "a" not in platform_release and platform_release != "2"   # 3-way
```

And `python_version != "3.8" and "x" not in python_version` still raises `InvalidMarkerError` rather than reaching the merge path at all — correct, since `python_version` is in `PYTHON_VERSION_MARKERS` and the swapped form was never valid there.

**Re: overlap with #955** — worth noting for whoever triages these, since they look adjacent but are not the same bug:

- This PR fixes the *crash* (`AttributeError: 'Constraint' object has no attribute 'flatten'`), which only reaches a **version-like** name (`platform_release`), where the two forms parse to different constraint types.
- #955 fixes a *silent wrong result* on **non-version-like** names, where both forms parse to a generic `Constraint` and merging drops a term:

  ```python
  >>> parse_marker('platform_version != "1" and "a" not in platform_version')
  "a" not in platform_version          # the != term is gone
  >>> _.validate({'platform_version': '1'})
  True                                  # should be False
  ```

  (same on `sys_platform`, `platform_machine`.)

So neither one subsumes the other. I applied **both** patches together: they don't conflict, and the full `tests/version` + `tests/constraints` suites pass (1559 passed). Both bug classes resolve.

Only nit, take it or leave it: the comment says the two forms "can't be intersected/unioned together", which is true for the mismatched pair but reads slightly like generic constraints can't be merged at all. Something like "a version constraint and a generic constraint have no common merge implementation, so leave them un-merged" might be marginally clearer. Not blocking.

radoering:
already resolved by #956

PR #965, opened 2026-08-29
Cache compiled JSON schema validators

dimbleby:
A bit artificial - I do not expect this to make any real difference to users, a typical execution of poetry will not run this path often.

But: running the complete set of poetry unit test - where we load the same schema many many times - this saves something like 20% of wall time.  Which is enough of an improvement to the developer experience that it seems worth having.

### DIFF
--- a/src/poetry/core/json/__init__.py
+++ b/src/poetry/core/json/__init__.py
@@ -2,7 +2,9 @@ from __future__ import annotations
 
 import json
 
+from functools import cache
 from importlib.resources import files
+from typing import TYPE_CHECKING
 from typing import Any
 
 import fastjsonschema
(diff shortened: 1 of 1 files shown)


dimbleby:
less than 20% but more than 10% comparing the [previous CI run](https://github.com/python-poetry/poetry-core/actions/runs/33071228731/job/98513708366) with the [one for this pull request](https://github.com/python-poetry/poetry-core/actions/runs/33273112200/job/99155054145)

before: 85s, after 76s

there is quite a lot of noise eg because of randomisation of test order, I was seeing more locally, it is what it is

radoering (approved):

